Issue #1, opened 2015-04-01
 32bit support? 


GuillaumeGomez:
[original](https://github.com/jeremyletang/rgtk/issues/250).

StephanvanSchaik:
This can now be closed.

PR #22, opened 2015-04-09
GtkRecentInfo: use time_t for arguments instead, and cast results to i64.

StephanvanSchaik:
Fixes issue #1 by applying the following patches:

 - [0001-GtkRecentInfo-use-c_int-to-interface-with-GTK-and-ca.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0001-GtkRecentInfo-use-c_int-to-interface-with-GTK-and-ca.patch)
 - [0002-GtkRecentInfo-use-time_t-instead.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0002-GtkRecentInfo-use-time_t-instead.patch)
 - [0003-GtkRecentInfo-types-can-be-inferred.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0003-GtkRecentInfo-types-can-be-inferred.patch)
 - [0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch)
 - [0005-GtkRecentInfo-use-match-instead-of-if.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0005-GtkRecentInfo-use-match-instead-of-if.patch)
 - [0006-GtkRecentInfo-code-clean-up.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0006-GtkRecentInfo-code-clean-up.patch)
 - [0007-GtkRecentInfo-added-indents-to-get-the-match-body-to.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0007-GtkRecentInfo-added-indents-to-get-the-match-body-to.patch)

Uses `time_t` for the arguments of type `time_t` rather than `i64` to make them compatible with both 32-bit and 64-bit architectures. Furthermore, any results of type `time_t` are cast to `i64`.

### DIFF
--- a/src/gtk/widgets/recent_info.rs
+++ b/src/gtk/widgets/recent_info.rs
@@ -100,8 +100,8 @@ impl RecentInfo {
             let mut time_ = 0;
 
             match to_bool(ffi::gtk_recent_info_get_application_info(
-                GTK_RECENT_INFO(self.unwrap_widget()), app_name.borrow_to_glib().0,
-                &mut app_exec, &mut count, &mut time_)) {
+                    GTK_RECENT_INFO(self.unwrap_widget()), app_name.borrow_to_glib().0,
+                    &mut app_exec, &mut count, &mut time_)) {
                 true => Some((FromGlibPtrNotNull::borrow(app_exec), count, time_ as u64)),
                 _ => None
             }
--- a/src/gtk/widgets/recent_info.rs
+++ b/src/gtk/widgets/recent_info.rs
@@ -27,10 +27,9 @@ impl RecentInfo {
     pub fn _ref(&self) -> Option<RecentInfo> {
         let tmp_pointer = unsafe { ffi::gtk_recent_info_ref(GTK_RECENT_INFO(self.unwrap_widget())) };
 
-        if tmp_pointer.is_null() {
-            None
-        } else {
-            Some(gtk::FFIWidget::wrap_widget(tmp_pointer as *mut ffi::C_GtkWidget))
+        match tmp_pointer.is_null() {
+            false => Some(gtk::FFIWidget::wrap_widget(tmp_pointer as *mut ffi::C_GtkWidget)),
+            _ => None
         }
     }
 
@@ -91,26 +90,17 @@ impl RecentInfo {
         unsafe { to_bool(ffi::gtk_recent_info_get_private_hint(GTK_RECENT_INFO(self.unwrap_widget()))) }
     }
 
-    pub fn get_application_info(&self, app_name: &str) -> Option<(String, u32, i64)> {
+    pub fn get_application_info(&self, app_name: &str) -> Option<(String, u32, u64)> {
         unsafe {
             let mut app_exec = ptr::null();
             let mut count = 0;
             let mut time_ = 0;
 
-            let ret = to_bool(
-                ffi::gtk_recent_info_get_application_info(
-                    GTK_RECENT_INFO(self.unwrap_widget()),
-                    app_name.borrow_to_glib().0,
-                    &mut app_exec,
-                    &mut count,
-                    &mut time_));
-
-            if ret {
-                let app_exec = FromGlibPtrNotNull::borrow(app_exec);
-                Some((app_exec, count, time_ as i64))
-            }
-            else {
-                None
+            match to_bool(ffi::gtk_recent_info_get_application_info(
+                GTK_RECENT_INFO(self.unwrap_widget()), app_name.borrow_to_glib().0,
+                &mut app_exec, &mut count, &mut time_)) {
+                true => Some((FromGlibPtrNotNull::borrow(app_exec), count, time_ as u64)),
+                _ => None
             }
         }
     }
--- a/src/gtk/widgets/recent_info.rs
+++ b/src/gtk/widgets/recent_info.rs
@@ -66,16 +66,25 @@ impl RecentInfo {
         }
     }
 
-    pub fn get_added(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
+    pub fn get_added(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) } {
+            x if x >= 0 => Some(x as u64),
+            _ => None
+        }
     }
 
-    pub fn get_modified(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
+    pub fn get_modified(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) } {
+            x if x >= 0 => Some(x as u64),
+            _ => None
+        }
     }
 
-    pub fn get_visited(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_visited(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
+    pub fn get_visited(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_visited(GTK_RECENT_INFO(self.unwrap_widget())) } {
+            x if x >= 0 => Some(x as u64),
+            _ => None
+        }
     }
 
     pub fn get_private_hint(&self) -> bool {
--- a/src/gtk/widgets/recent_info.rs
+++ b/src/gtk/widgets/recent_info.rs
@@ -19,7 +19,7 @@ use gtk::FFIWidget;
 use gtk::cast::GTK_RECENT_INFO;
 use std::ptr;
 use glib::translate::{FromGlibPtr, FromGlibPtrNotNull, FromGlibPtrContainer, ToGlibPtr};
-use libc::c_char;
+use libc::{c_char, time_t};
 
 struct_Widget!(RecentInfo);
 
@@ -67,15 +67,15 @@ impl RecentInfo {
     }
 
     pub fn get_added(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) }
+        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
     }
 
     pub fn get_modified(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) }
+        unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
     }
 
     pub fn get_visited(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_visited(GTK_RECENT_INFO(self.unwrap_widget())) }
+        unsafe { ffi::gtk_recent_info_get_visited(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
     }
 
     pub fn get_private_hint(&self) -> bool {
@@ -86,7 +86,7 @@ impl RecentInfo {
         unsafe {
             let mut app_exec = ptr::null();
             let mut count = 0u32;
-            let mut time_ = 0i64;
+            let mut time_: time_t = 0;
 
             let ret = to_bool(
                 ffi::gtk_recent_info_get_application_info(
@@ -98,7 +98,7 @@ impl RecentInfo {
 
             if ret {
                 let app_exec = FromGlibPtrNotNull::borrow(app_exec);
-                Some((app_exec, count, time_))
+                Some((app_exec, count, time_ as i64))
             }
             else {
                 None
--- a/src/gtk/widgets/recent_info.rs
+++ b/src/gtk/widgets/recent_info.rs
@@ -19,7 +19,7 @@ use gtk::FFIWidget;
 use gtk::cast::GTK_RECENT_INFO;
 use std::ptr;
 use glib::translate::{FromGlibPtr, FromGlibPtrNotNull, FromGlibPtrContainer, ToGlibPtr};
-use libc::{c_char, time_t};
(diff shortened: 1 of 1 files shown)


src/gtk/widgets/recent_info.rs
```diff
@@ -86,7 +86,7 @@ impl RecentInfo {
         unsafe {
             let mut app_exec = ptr::null();
             let mut count = 0u32;
-            let mut time_ = 0i64;
+            let mut time_: time_t = 0;
```
gkoz: FYI the type can be inferred here.

src/gtk/widgets/recent_info.rs
```diff
@@ -67,15 +67,15 @@ impl RecentInfo {
     }
 
     pub fn get_added(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) }
+        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) as i64 }
```
gkoz: Is there a reason the type is signed?

StephanvanSchaik:
[`gtk_recent_info_get_added`](https://people.gnome.org/~shaunm/girdoc/C/Gtk.RecentInfo.get_added.html) returns a value of type `time_t`, Unix- and POSIX-compliant systems implement this as a signed integer.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: [`gtk_recent_info_get_added`](https://people.gnome.org/~shaunm/girdoc/C/Gtk.RecentInfo.get_added.html) returns a value of type `time_t`, Unix- and POSIX-compliant systems implement this as a signed integer.

gkoz: They can fail? Uh. Yeah, an `Option<u64>` seems better.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: The following patch will fix this:

 - [0003-GtkRecentInfo-types-can-be-inferred.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0003-GtkRecentInfo-types-can-be-inferred.patch)

src/gtk/widgets/recent_info.rs
StephanvanSchaik: The following patch will fix this:

 - [0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch)

StephanvanSchaik:
Complete patch set:

 - [0001-GtkRecentInfo-use-c_int-to-interface-with-GTK-and-ca.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0001-GtkRecentInfo-use-c_int-to-interface-with-GTK-and-ca.patch)
 - [0002-GtkRecentInfo-use-time_t-instead.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0002-GtkRecentInfo-use-time_t-instead.patch)
 - [0003-GtkRecentInfo-types-can-be-inferred.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0003-GtkRecentInfo-types-can-be-inferred.patch)
 - [0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0004-GtkRecentInfo-use-Option-u64-instead-of-i64.patch)
 - [0005-GtkRecentInfo-use-match-instead-of-if.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0005-GtkRecentInfo-use-match-instead-of-if.patch)
 - [0006-GtkRecentInfo-code-clean-up.patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0006-GtkRecentInfo-code-clean-up.patch)

src/gtk/widgets/recent_info.rs
```diff
@@ -141,23 +142,22 @@ impl RecentInfo {
 
     pub fn has_group(&self, group_name: &str) -> bool {
         unsafe {
-            to_bool(
-                ffi::gtk_recent_info_has_group(GTK_RECENT_INFO(self.unwrap_widget()),
-                                               group_name.borrow_to_glib().0))
+            to_bool(ffi::gtk_recent_info_has_group(GTK_RECENT_INFO(self.unwrap_widget()),
+                group_name.borrow_to_glib().0))
         }
```
gkoz: Is this more readable? Which function takes `group_name...` here?
Why is this change in the `time_t` PR?
Also, if we touch a line with `to_bool` we should replace it with `to_glib()` anyway.

src/gtk/widgets/recent_info.rs
```diff
@@ -40,68 +41,69 @@ impl RecentInfo {
...
     }
 
-    pub fn get_added(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) }
+    pub fn get_added(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) } {
+            x if x >= 0 => Some(x as u64),
+            _ => None
+        }
```
gkoz: The let boilerplatey way would probably be to implement a conversion in `glib::translate`.

src/gtk/widgets/recent_info.rs
```diff
@@ -25,12 +25,13 @@ struct_Widget!(RecentInfo);
...
-        let tmp_pointer = unsafe { ffi::gtk_recent_info_ref(GTK_RECENT_INFO(self.unwrap_widget())) };
+        let tmp_pointer = unsafe {
+            ffi::gtk_recent_info_ref(GTK_RECENT_INFO(self.unwrap_widget()))
+        };
 
-        if tmp_pointer.is_null() {
-            None
-        } else {
-            Some(gtk::FFIWidget::wrap_widget(tmp_pointer as *mut ffi::C_GtkWidget))
+        match tmp_pointer.is_null() {
```
gkoz: Why?

src/gtk/widgets/recent_info.rs
StephanvanSchaik: It makes sense to integrate a code clean up within this file, because this pull request already modifies this particular file. However, such changes can always be reverted. The `group_name` argument is passed to `ffi::gtk_recentinfo_has_group`, which is clearly shown by any editor that shows matching parentheses. Besides, the problem with alignment of arguments is that there are plenty of cases where you would have to trade a lot of horizontal space for vertical space, for no particular reason other than decorative purposes.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: Do you want me to do that within this pull request or another one? My goal within this pull request is to add support for 32-bit architectures such as x86 and ARMv7a, which is currently being neglected, as I actually want to use this Rust binding for one of my projects.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: Because Rust is more a functional than an imperative programming language by nature, it is more logical to use `match` rather than `if`. At least, in my opinion.

src/gtk/widgets/recent_info.rs
gkoz: Matching parentheses? What if I want to see what goes where without moving the cursor?

These changes introduce pointless code churn that has little to no effect on the number of lines. Talk about decorative purposes.

src/gtk/widgets/recent_info.rs
gkoz: I count 7 occurences of 'false =>` in Rust source code. Doesn't seem very idiomatic. The effort could be spent better on eliminating the `if` altogether.

src/gtk/widgets/recent_info.rs
gkoz: I guess it's okay like this but adding a conversion would be nice.

src/gtk/widgets/recent_info.rs
```diff
@@ -40,68 +41,69 @@ impl RecentInfo {
...
-            else {
-                None
+            let mut count = 0;
+            let mut time_ = 0;
+
+            match to_bool(ffi::gtk_recent_info_get_application_info(
+                GTK_RECENT_INFO(self.unwrap_widget()), app_name.borrow_to_glib().0,
+                &mut app_exec, &mut count, &mut time_)) {
+                true => Some((FromGlibPtrNotNull::borrow(app_exec), count, time_ as u64)),
+                _ => None
```
gkoz: Here you gain some vertical space from making a mixture of function calls and control structure that's very hard to decipher. And we weren't short on vertical space because this is the whole function, it fits on the screen just fine.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: I am not even sure when you ever want to know what function that argument will be passed to, unless you are editing those functions or doing some kind of code review, in which case the difference can clearly be neglected as it is hard to measure the difference in time. In both cases you will be moving the cursor often, and for most editors it should require minimal effort.

What I do care more about is the amount of vertical space that is used, especially when it is the result of minimising the amount of horizontal space that can be used due to alignment constraints imposed by these decorative purposes I am talking about. However, I do agree that these are hard to notice in these particular cases, but in such cases you should consider functions that accept more than one argument, where you usually end up allocating a single line per argument, as well as the overall effect. An example, which you have spotted already, can be found at `ffi::gtk_recent_info_get_application_info` (line 92).

src/gtk/widgets/recent_info.rs
StephanvanSchaik: It kind of depends when you want to use something like: `_ => ...`. Especially when adding new entries to enum types, it would be wise to be explicit about it. However, for booleans, I do agree that cases of `false => ...` are pointless.

Coming both from a functional and an imperative background, I think that `match` is much stronger in what you can do with it, as opposed to `if`, but I don't think you can eliminate `if`-statements from the Rust language entirely. The reason why is because you often need an additional `if`-clausule for some `match`-cases.

src/gtk/widgets/recent_info.rs
StephanvanSchaik: It doesn't really matter whether function calls do fit or do not fit on a screen, mostly because specifying a large amount of arguments is being frowned upon, and rightfully so. What does matter is whether function definitions, i.e. functions including their code body, themselves do fit on the screen, or alternatively, how much you have to scroll to get to certain places in the code you do not know the exact line number of. Then again, that mostly depends on the editor and your way of working with it.

I prefer minimised vertical spacing, because of this, and because I like functions to be simple, i.e. do one thing and do it right. The latter is something I tend to measure in terms of compactness, where early return conditions can be somewhat neglected, if those are needed.

Of course, if you prefer not to apply these specific changes, I can revert them (partially).

src/gtk/widgets/recent_info.rs
gkoz: The biggest problem for me here is touching stuff that wasn't broken, doesn't benefit from changing and on top of that becomes harder to read at least to me.

StephanvanSchaik: Well, that it is not broken is not entirely true, since some of the lines of code were not conform to the hundred column rule.

src/gtk/widgets/recent_info.rs
gkoz: The whole function body had fit on the screen, it didn't need abbreviating. But whatever. It's the same level of indentation of the function call and the following match branches that is highly suboptimal.

StephanvanSchaik: In what sense is it suboptimal?

src/gtk/widgets/recent_info.rs
gkoz: Broken as in buggy.

StephanvanSchaik: Sure, but that still does not change the fact that the code is not conform to other aspects of code style, in this case, the hundred column rule.

src/gtk/widgets/recent_info.rs
gkoz: The whitespace suggests that this is all a big function call and then you encounter the match syntax and have to backtrack and spend effort on separating what could be obviously separate by virtue of being indented differently.
We can turn to the Rust codebase again and see that in cases of multiline function call in the match expression the indentation always makes it obvious where the match body begins.
We'll also find that Rust is okay with putting each argument on a separate line and doesn't seem to strive to save as much vertical space as possible.

StephanvanSchaik: I think that I see what you mean with that. It could be solved by using double indents, rather than trying to align each and every argument with the opening bracket. Unless you have very specific reasoning why it is much better to waste effort on getting everything aligned appropriately, of course.

gkoz: Double indenting will probably work. Don't forget that no formatting effort at all was required in this function, just changing two lines. Refactoring `if`s was uncalled for.

gkoz:
After some more thinking, I feel messing with `time_t` doesn't really help. What do you think of just checking for `-1` and returning `Option<time_t>` so the user can plug that value into whatever time library they're using? It can be revisited when some actual time type is available.

StephanvanSchaik:
From what I have seen, the `rust-time` library seems to use `i64` as well within their `Timespec` struct, since they are using `clock_gettime` and friends on other platforms.

gkoz:
Yeah, you're right, both `time` and `chrono` use `i64` for unix timestamps. That's the obvious choice then, sorry for the run-around. The `Option` still makes sense though.

GuillaumeGomez:
@StephanvanSchaik: can you please update your PR ? I'll take a look after that. Thanks for your work !

StephanvanSchaik:
All right, I'll re-organise my fork as well, then.

GuillaumeGomez:
You could have left your PR open. Well, whatever... I'll look forward to the next one.

StephanvanSchaik:
Well, I could have, but I wanted to redo the branches, as this was currently pulling it from the `master` branch.

PR #25, opened 2015-04-09
Housekeeping

gkoz:
It does feel weird not having travis tests... :)
I wonder if we could use Homu for this project. Catching causes of breakage between five repos might get hairy.


gkoz:
Ah, one more thing should go here -- move the whole gtk up one level.

gkoz:
That's far from easy though, so better merge as is.

gkoz:
Okay, `mod gtk` has been unwrapped. This probably is the last of major breaking changes.

gkoz:
Importing `gtk::*` no longer works because it introduces a conflict with `fn main()`. Importing `gtk::traits::*` should suffice.

GuillaumeGomez:
Oh right, I'll re-add travis for this.

gkoz:
@GuillaumeGomez we may need some way to check for cross-repo breakage, e.g. if a change to `glib` or `gdk` breaks `gtk`.

GuillaumeGomez:
Hum... I don't have a lot of knowledges in that field but yes, we need it. I'll find information about this.

gkoz:
A brute force solution might be to checkout `gtk` master and try to build examples. Cargo allows overriding dependencies locally so if you're in the `gdk` repo for example you can
```
mkdir .cargo
echo 'paths = ["."]' > .cargo/config
git clone https://guthub.com/rust-gnome/gtk _test
cd _test/examples
cargo build
```
Cargo will find the config and use the checked out `gdk` tree it references.

GuillaumeGomez:
I'm afraid this way would take a veeeeeeeeeeeeeeeery long time for each PR.

gkoz:
You might be surprised when we axe the old signals stuff.

GuillaumeGomez:
Oh that's right, I totally forgot that. Then what about you create a PR for that (it's your idea after all :p) ?

gkoz:
All right, now that travis is hooked up again, I'll look into that.

GuillaumeGomez:
Thanks !

PR #29, opened 2015-04-11
GtkRecentInfo: implement error handling, and cast time_t to u64.

StephanvanSchaik:
This is a follow up on my previous pull request #22, which attempts to fix a few functions in GtkRecentInfo by checking for negative values of type `time_t`, and casting values of `time_t` to u64. Additionally, this fix will allow the project to be built on 32-bit platforms such as linux x86 and armv7a (cfr. issue #1).

### DIFF
--- a/src/widgets/recent_info.rs
+++ b/src/widgets/recent_info.rs
@@ -66,42 +66,42 @@ impl RecentInfo {
         }
     }
 
-    pub fn get_added(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) }
+    pub fn get_added(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_added(GTK_RECENT_INFO(self.unwrap_widget())) } {
+            x if x >= 0 => Some(x as u64),
+            _ => None
+        }
     }
 
-    pub fn get_modified(&self) -> i64 {
-        unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) }
+    pub fn get_modified(&self) -> Option<u64> {
+        match unsafe { ffi::gtk_recent_info_get_modified(GTK_RECENT_INFO(self.unwrap_widget())) } {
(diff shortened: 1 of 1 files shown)


StephanvanSchaik:
Additional note: I am aware that `FromGlib` has been implemented to cast `i64` to `Option<u64>`. However, at the moment I think it would be better to be explicit about the error checks, as it is hard to deduce that from the implicit cast, without prior knowledge, as `i64` does not necessarily imply C's common way of error handling.

gkoz:
It may even be worth checking against exactly `-1`, who knows, maybe someone can manage to put a negative value there.

StephanvanSchaik:
It is common on UNIX to check if the return value is less than zero, with a few exceptions, namely:
 - when the documentation tells you that certain negative values have a different meaning.
 - when the documentation tells you that certain negative values represent different errors, and you are interested in knowing the specific error (in which case, you often use something like `perror`).

I don't think GTK+ is different in this regard, and I also think it is a lot better to generalise instead of being specific for this particular case.

gkoz:
Yeah I guess we could treat a negative `time_t` as just an error.
However the convention for returning negative values for errors is what motivated the aforementioned implementations. I see your point about explicitness but there is `Option` in the signature and boilerplate does seem unpleasant.
With those conversions I aim to support generating as much of this code as possible automatically from GIR definitions and possibly small metadata additions, we need take that objective into account when weighing different approaches.

GuillaumeGomez:
A similar issue appeared in rustc, they're making a whole new RFC for handling date time on File system. I don't really have an opinion on this, I'll wait for @jeremyletang's point of view before going any further.

StephanvanSchaik:
Would it be possible to merge this as it is now? Since this would allow compilation on 32-bit platforms, which is something that I and some others are interested in. I think we can always change this file within the scope of future discussions.

gkoz:
@GuillaumeGomez this is a pretty minimal change (from one int type to another in fact) and the time story in Rust won't be figure out for some time...

GuillaumeGomez:
Yes, I guess. We'll have to go back on that.

Issue #26, opened 2015-04-10
Notebooks expect labels rather than widgets.


StephanvanSchaik:
The [documentation for GTK+ 3](https://developer.gnome.org/gtk3/stable/GtkNotebook.html) shows that most functions that currently expect a `Label`, should in fact be able to accept widgets of any type (i.e. use `Widget`).

I am working on a patch that fixes this. Although if there is anything I need to know while working on this, then that kind of information is welcome.

gkoz:
It looks like we'll be taking random types of widgets a lot. It might be worth pondering whether [static](https://github.com/rust-gnome/gtk/blob/master/src/gtk/widgets/note_book.rs#L33) or dynamic dispatch is preferable here.

gkoz:
The position arguments seem to call for a type that would let us specify either a specific number, or 'end'. `Option` could work for that but would be less explicit. Do any other parts of gtk use the same convention?

StephanvanSchaik:
At the moment, I would propose the following [patch](http://www.synkhronix.com/pub/rust/patches/rgtk/0001-notebook-changed-functions-to-accept-widgets-rather-.patch) to fix the issue with the widget types (I'll send a pull request later).

Well, there are similar cases of `insert` functions, such as [this one](https://developer.gnome.org/gtk3/stable/GtkToolbar.html#gtk-toolbar-insert).

gkoz:
The input position parameters will need a separate PR then to decide on the type to use.
I'll make a PR to add a conversion for `Option<u32>` so we can at least give the output values the `time_t` treatment.
You can avoid a little bit of `gtk::` repetitiveness by importing the `WidgetTrait`.

gkoz:
https://github.com/rust-gnome/glib/pull/5 should help reduce boilerplate in cases where negative values represents failure.

GuillaumeGomez:
With the PR merged, I guess this issue can be closed ?

StephanvanSchaik:
Yes, this issue has been resolved :).

PR #30, opened 2015-04-12
GtkNoteBook: allow more generic arguments, and an example.

StephanvanSchaik:
This pull request changes the `GtkNoteBook` methods to accept widgets rather than labels (cfr. issue #26) and to check for failures. In addition, there is also an example of how to use a notebook. However, the use of closures in that example needs some polishing.


GuillaumeGomez:
Nice ! Thanks for that !

Issue #28, opened 2015-04-11
How to draw in-memory images to a window?


cybergeek94:
My application needs to draw in-memory RGB and grayscale images to the screen for display. Piston and Conrod make this easy but I'm looking for a GUI toolkit that is more robust and can produce a native look and feel.

It looks like this might be implemented by creating the `GdkPixBuf` struct and then adding support for creating an `rgtk::gtk::Image` backed by one and getting a mutable reference to it.

StephanvanSchaik:
At the moment, [the implementation of `GtkImage`](https://github.com/rust-gnome/gtk/blob/master/src/widgets/image.rs) is a basic one. To support the creation of a `GtkImage` from a `GdkPixBuf`, [`gtk_image_new_from_pixbuf`](https://developer.gnome.org/gtk3/stable/GtkImage.html#gtk-image-new-from-pixbuf) should be implemented. The `GdkPixBuf` already has been implemented as [`Pixbuf`](https://github.com/rust-gnome/gdk/blob/master/src/widgets/pixbuf.rs).

Once my pull request for 32-bit has been merged. I will pick that up.

In the meanwhile, it should be possible to use a GtkDrawingArea in conjunction with Cairo. See [this example](https://github.com/rust-gnome/gtk/blob/master/examples/src/cairotest.rs). Although, it kind of depends on what you are planning to do exactly.

cybergeek94:
It's for [this project][1]. I want to display images already loaded into memory with [rust-image][2] from Piston, alongside a generated black-and-white image of their hash, as shown in [this article][3].

I'm currently already doing this with Conrod, but I'm having some stuttering issues with loading new images to display, as they have to be copied into OpenGL textures on the main thread because OpenGL contexts cannot be used across threads. I also am having trouble expressing advanced GUI constructs in Conrod's primitive API and I would prefer to have a native look and feel because Conrod's GUI design is pretty homely.

[1]: https://github.com/cybergeek94/img-dup
[2]: https://github.com/PistonDevelopers/image
[3]: http://www.hackerfactor.com/blog/?/archives/432-Looks-Like-It.html

StephanvanSchaik:
[...]

PR #32, opened 2015-04-12
GtkImage: added functions to work with pixbufs.

StephanvanSchaik:
Added the `Pixbuf` methods to `GtkImage` (cfr. issue #28).


GuillaumeGomez:
Thanks !

Issue #3, opened 2015-04-01
Work out the thread safety/affinity story 


GuillaumeGomez:
[original](https://github.com/jeremyletang/rgtk/issues/243).

gkoz:
Dynamic checks to prevent misuse have been added in #189. Having another set of APIs enforcing this statically still is desirable but not a priority.

PR #34, opened 2015-04-14
Introduction of the Gtk struct to guarantee that Gtk is used within a single thread.

StephanvanSchaik:
Initial implementation of the proposal in issue #3:

 - Introduces a `Gtk` struct which should be instantiated to use gtk functions.
 - Introduces `Builder` traits for each of the widgets to introduce constructors, that are implemented for the `Gtk` struct.
 - Makes the old constructors and main gtk functions unavailable.
 - Integrates these changes within the examples.

**Note**: not all of the widgets do implement `Builder` traits yet (about twenty or so).


gkoz:
Woah. I'm afraid, starting this from converting the constructors is premature and likely to cause pointless busywork. That's why I created that issue - so we could discuss strategy.

StephanvanSchaik:
I think what was proposed within that issue was already a good way of limiting the use of GTK+ to a single thread: changing the functions to be methods of a struct that cannot be sent to another thread. The changes within this pull request yield very usable results, at least, in my opinion (see the examples to get an idea). Furthermore, I have to say that this is probably the most straight-forward method, that doesn't involve checking the context at run-time.

To add to that issue, I don't think it is necessary to have method chaining, as it already seems to be simple enough as it is (e.g. `gtk.button_with_label(...)`). 

However, I do get your concern with the amount of work involved with these changes. It would be nice to get the most common widgets in shape first (especially when integrating the new signal system). If that is desired, I can split this change up in smaller steps to work and contemplate at a slower pace.

StephanvanSchaik:
Or would we still be expecting some major structural changes to the widgets in general?

gkoz:
Moving slower would be more comfortable.

Chaining does add a bit of complication but dumping all of the constructors into one namespace doesn't seem ideal either. Consider also [this PR](https://github.com/jeremyletang/rgtk/pull/246) where I explore adding some structure to the modules layout. It may or may not be worthwhile to mimic that structure with these methods so the invocation would be along the lines of `gtk.dialog().about(...)`, but we should move away from the flat organization.

A couple of smaller remarks:
* the constructors shouldn't return `Option`, that convention was a mistake, there pointer is never `NULL`;
* as is, the `Builder` suffix doesn't seem right, the traits make `Gtk` a factory not a builder;
* rehashing the constructors is a good occasion to get rid of the separation between widgets and traits, see the linked PR.

As to major structural changes, I would like to reorganize the inheritance system, some prototypes I explore [here](https://github.com/jeremyletang/rgtk/pull/197) and [here](https://github.com/gkoz/rgtk/commits/convert)

StephanvanSchaik:
The other major category of widgets to apply that to would be buttons.

That convention could best be removed within its own pull request. Similar to the explicit casts that the compiler can infer nowadays.

On the `Builder` name, I agree with you that `Factory` would be a better naming choice.

Thanks for the links, I'll be reading through the material posted there.

GuillaumeGomez:
I really don't like it this way. Having a factory seems so unnatural to me...

gkoz:
@GuillaumeGomez
Yes, it does seem awkward, possibly no other GUI library has made us deal with this. But we need to anchor the widgets to a thread and the choice seems to be between passing a reference to `Gtk` as the first argument to every constructor or making it a factory. And these two alternatives are almost one and the same because the first argument of a method is the receiver.

@StephanvanSchaik 
There's another complication I hadn't recalled yesterday: one may want to create widgets in a signal handler and sneaking a `Gtk` reference in manually seems like a hassle. If the handler were to have access to a `Gtk` reference automatically, it would require every widget to actually carry that reference (or a clone of the struct).

GuillaumeGomez:
I add as "against argument" that we can create a thread_local in gtk::init which set a value when it's created in the main thread. Then, every created widget after that check if it's created in the main thread by looking at the thread_local data. With that, we avoid all that factory (I *really* don't like it).

@jeremyletang: good for you ? (he's also against that factory)

gkoz:
Okay, so you prefer runtime checks then?

GuillaumeGomez:
Totally.

gkoz:
Would you please state your and @jeremyletang's preference in #3 for the record?

jeremyletang:
Hi, 
I think that make a big `Gtk` factory is not rusty at all. A better way to do this will be I think to create a `thread_local` storage during the first call of `gtk::init`(we consider that the user call it in the main thread, the behaviour stay the same between the two methods).

Then at each instantiation of a widget we can check if the `thread_local storage` is already set. If not I think we can `fail()` we an error message like `widget MUST be instantiates inside main thread as this will only be saw be the developer when he make it's tests.

According to the `Gtk` documentation, widgets instantiation cannot fail, so maybe we can not return an option on widget construction.

So we can fail if:
* The c pointer == null
* The user don't instantiate the widget in the main thread.

gkoz:
One might argue that employing dynamic checks may not be very Rusty either, but okay, that can work.

src/rt.rs
```diff
@@ -13,84 +13,90 @@
...
 use glib::translate::{FromGlibPtr};
 use glib::{to_bool, to_gboolean};
 
-pub fn init() {
-    unsafe {
-        ffi::gtk_init(ptr::null(), ptr::null());
+pub struct Gtk;
+
+impl Gtk {
+    pub fn new() -> Gtk {
```
bombless: I guess better to return `Result<Gtk, GtkError>` or something here.
So when user try to call it the second time on a different thread, `unwrap` can print the error to show that it need a main thread to work.

gkoz: Umm, I think if we go down the runtime checks path, we'll just panic when you try to touch GTK from a wrong thread. There could be a function to find out if the thread is the right one.

PR #35, opened 2015-04-14
Adapt to glib::Value changes

gkoz:


gkoz:
Depends on https://github.com/rust-gnome/glib/pull/13

GuillaumeGomez:
Thanks !

PR #36, opened 2015-04-15
Make all methods take widgets by & instead of &mut

gkoz:
All of the widget types are wrappers around recounted shared references
and contain no data we would ever mutate. There's no meaningful difference
between a shared (&) and a mutable (&mut) reference to a widget, the latter
doesn't guarantee the uniqueness of the reference it wraps.

We have a practical reason to change this now: a `Fn` closure can't call any
mutating methods on the variables moved into it so the status quo would
artificially limit the usefulness of signal handlers proposed in #33.


GuillaumeGomez:
Ok, I get your explanation. The plan was to say "we need a &mut because values in the pointer wrapped in the struct will change". However, if you need it that way, fine by me ! Thanks !

PR #40, opened 2015-04-20
Implement more new style signals

gkoz:
I'd appreciate someone taking the time to review the closure and callback signatures against [the GTK docs](https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget.signals). The compiler can't be of much help here and it's easy to make a mistake.
The `GtkWidget` signals are here, other widgets to follow.

### DIFF
--- a/gtk-sys/src/lib.rs
+++ b/gtk-sys/src/lib.rs
@@ -319,6 +319,8 @@ pub struct C_GtkMovementStep;
 pub struct C_GtkSocket;
 #[repr(C)]
 pub struct C_GtkEventBox;
+#[repr(C)]
+pub struct C_GtkTooltip;
 
 //pub type GtkTreeModelForeachFunc = fn(model: *mut C_GtkTreeModel, path: *mut C_GtkTreePath, iter: *mut C_GtkTreeIter, data: gpointer) -> Gboolean;
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -97,6 +97,7 @@ pub use self::widgets::{
     EntryBuffer,
     Entry,
     Switch,
+    Range,
     Scale,
     SpinButton,
     Spinner,
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -7,9 +7,9 @@ use std::mem::transmute;
 
 use glib::signal::connect;
 use glib::translate::*;
-use glib::{ParamSpec};
+use glib::{FFIGObject, ParamSpec};
 
-use ffi::{Gboolean, C_GtkAdjustment};
+use ffi::{Gboolean, C_GtkAdjustment, C_GtkTreeSelection, C_GtkTreeViewColumn};
 use gdk::{
     EventAny, EventButton, EventConfigure, EventCrossing, EventExpose, EventFocus, EventGrabBroken,
     EventKey, EventMotion, EventProperty, EventProximity, EventScroll, EventWindowState,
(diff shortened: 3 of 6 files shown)


GuillaumeGomez:
For this, I think we'll have to write new examples. Maybe adding new ones to the current ? I'll create an example repo to split out examples from gtk.

gkoz:
@GuillaumeGomez can you elaborate on the examples issue? I'm also curious why you put an older version of examples into the new repo.

GuillaumeGomez:
I just spitted out the existing gtk examples.

gkoz:
You gtk repo might've not been up to date then. Compare
https://github.com/rust-gnome/gtk/blob/master/examples/src/gtktest.rs and https://github.com/rust-gnome/examples/blob/master/src/gtktest.rs

GuillaumeGomez:
Weird... Wanna fix it or I do it ?

gkoz:
You do it please =)

GuillaumeGomez:
Ok !

GuillaumeGomez:
It's fixed.

GuillaumeGomez:
Still working on it ?

gkoz:
Yeah, I had a cold so couldn't do much, going to pick it up now. I won't mind someone else doing the rest of the signals though haha

GuillaumeGomez:
I'll have a busy week-end but why not next week ? Mail me to remind me it ! :-)

gkoz:
This should be it.
I've omitted a few signals that aren't recommended to connect in applications. In the unlikely case that anyone needs to add them, there're plenty examples now. Other than that the old `signals.rs` functionality should be covered, so I'm going to prepare a patch that removes the old style signals.
More fresh eyes on this are welcome cause there's been a lot of copy-pasting.

GuillaumeGomez:
Sorry but I'll only take a look at it tomorrow. However, thanks for your work, the little part I saw is very nice.

GuillaumeGomez:
I just checked your code. Nothing to say, it's just excellent. Thanks a lot for your work !

gkoz:
Thanks :)

GuillaumeGomez:
Well, actually it's thanks to you but thanks for thanking me for thanking you haha.

oakes:
I've been out of the loop for a bit but I've been spending some time today trying to rebase onto the latest changes. I can't get the new signals to work because I can't pass mutable references to Fn closures. My IDE works by passing a mutable reference to a "state" struct to all my callbacks. Should I fix this by wrapping it in a `Cell` perhaps, or am I overlooking a better solution?

gkoz:
@oakes I'm pretty sure the closures can't be made `FnMut` because they need to be reentrant, so you'd have to use cells. Right now they're also `'static` i.e. can't capture any non-static references, so, you might also need `Rc`. The latter limitation could be overcome of it turned out to be a big issue.

oakes:
I've managed to overhaul my code to use `Rc<RefCell>`. The only remaining problem is that it is panicing due to conflicting borrows. There's a code-path where I borrow the state from the `RefCell`, pass it to a function that refreshes my `TreeView`, which then triggers the "row_expanded" signal, which in turn borrows the state from the `RefCell`. I didn't consider this beforehand, but it means I can't borrow my state in any code that might trigger a signal that needs that state as well. I'll have to think about this more I suppose.

gkoz:
This sounds like proof that allowing `FnMut`s would've been a disaster...
I wonder if you can partition the state into finer grained elements that could be borrowed in smaller scopes.

oakes:
In this case I was able to avoid implementing a queue system, because I realized I don't actually need those signal handlers to run when I refresh the tree. I only need them to run when the user interacts with the tree. Therefore, I was able to fix it by not running the handlers if the tree is currently being refreshed. There may be side-effects I haven't considered yet, but so far it seems fine.

PR #45, opened 2015-04-29
Axe the legacy signals framework

gkoz:
This is a big [breaking-change].
The new way of using signals is demonstrated in the examples repo.
The new traits for assigning signal handlers are defined in signal.rs.
The crate doesn't take ages to compile any more, yay!


GuillaumeGomez:
And thanks for this PR too !

PR #46, opened 2015-04-30
Add possibility to remove model from a tree view

GuillaumeGomez:


src/widgets/tree_model.rs
```diff
@@ -187,4 +187,4 @@ fn my_fn(model: *mut ffi::C_GtkTreeModel, path: *mut ffi::C_GtkTreePath, iter: *
     glib::to_gboolean(data.0(&mut TreeModel::wrap_pointer(model), &mut TreePath::wrap_pointer(path), &mut TreeIter::wrap_pointer(iter), data.1))
 }
 
-impl_drop!(TreeModel, GTK_TREE_MODEL);
+impl_drop!(TreeModel, GTK_TREE_MODEL);
```
gkoz: Your editor seems to be seriously messed up =)

GuillaumeGomez: It seems so. XD

PR #51, opened 2015-04-30
Switch travis to container mode

gkoz:


gkoz:
Do I need another PR to see if it works?

GuillaumeGomez:
No, I'll just restart #49.

gkoz:
I guess that changeset had the old `.travis.yml` and it would need to be rebased to see the new one.

GuillaumeGomez:
Hum... Hold on !

gkoz:
It seems I broke something there, it doesn't seem schedule builds any more :(

GuillaumeGomez:
Yup, that's weird. For now I'll revert your PR, we'll take anoter look at it tomorrow ! ^^

gkoz:
I've found the problem with my edit.
The bigger issue is the gtk package they're giving me is 3.4.2 not even 3.6. They're almost equally old at this point so it might not make a big difference...

GuillaumeGomez:
Hum... Well, for now let's keep the "old" way.

PR #62, opened 2015-05-07
Old signals remnants cleanup

gkoz:


GuillaumeGomez:
I was sure they've already been removed from the code... A part of the old rgtk is gone :'(

gkoz:
Macros are evil >:-0

GuillaumeGomez:
Without macros, a big part of rust wouldn't exist ! So they're heaven !

gkoz:
Yeah, that too. They're both then.

GuillaumeGomez:
There's good and evil in everything (the wise man has spoke !).

PR #64, opened 2015-05-09
Update dialog parent parameters to borrow window

lucaswerkmeister:
Neither do we require ownership of the parent window, nor are dialog creators usually in a position to give it up.


GuillaumeGomez:
All good ! Thanks for your work !

lucaswerkmeister:
You’re welcome, thanks for the quick response!

PR #59, opened 2015-05-05
[RFC] GObject wrappers refactoring

gkoz:
_Do not merge!_

Here's my latest and likely final (modulo small improvements) version of the `GObject` wrapper framework first proposed at https://github.com/jeremyletang/rgtk/pull/197 that's supposed to replace the current approach of using macros and a C glue module.

Objectives:
* reduce code duplication by encapsulating reference counting and containers management;
* reduce boilerplate by factoring out most common conversions and checks;
* reduce reliance on macros in favor of generic programming that allows better type checking;
* improve support for implementing multiple interfaces (e.g. `GtkTreeModel` is an interface implemented by `GtkTreeStore`, it shouldn't be represented as a separate widget);
* support implementing some widgets in separate crates (e.g. `libvte`);
* remove the need for `gtk_glue.c` and build complications it brings;
* move toward automatic code generation from GIR by unifying the translations to/from the FFI types.

The `glib::translate` module shows all its power now: you can have a string, a widget, a nullable widget or a `GList` of widgets, doesn't matter, it just works, all applicable asserts included:
```Rust
    pub fn new() -> Button {
        unsafe { Widget::borrow_from_glib(ffi::gtk_button_new()).downcast_unchecked() }
    }

    fn get_name(&self) -> Option<String> {
        unsafe {
            from_glib_none(ffi::gtk_widget_get_name(self.upcast().to_glib_none().0))
        }
    }

    fn set_name(&self, name: &str) {
        unsafe {
            ffi::gtk_widget_set_name(self.upcast().to_glib_none().0, name.to_glib_none().0)
        }
    }

    fn get_ancestor(&self, widget_type: Type) -> Option<Widget> {
        unsafe {
            from_glib_none(
                 ffi::gtk_widget_get_ancestor(self.upcast().to_glib_none().0, widget_type.to_glib()))
        }
    }

    fn get_children(&self) -> Vec<Widget> {
        unsafe {
            Vec::from_glib_container(
                ffi::gtk_container_get_children(self.upcast().to_glib_none().0))
        }
    }
```

The `basic` example (not present in this PR) demonstrates the use of the only two widgets supported in this proof of concept - `Window` and `Button` each implementing its own trait as well as `WidgetTrait` and `ContainerTrait`, and downcasting from `Widget` to `Button`.

[...]

PR #65, opened 2015-05-09
Signals: make all the closure type bounds include the 'static bound

gkoz:
This was intended from the start because in general we can't tell
anything about the closure's lifetime, it may potentially outlive
everything else.
https://github.com/rust-gnome/gtk/issues/16#issuecomment-92441810

So although a [breaking-change] this is a bug fix.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -40,56 +40,56 @@ impl ToGlib for Inhibit {
 unsafe fn into_raw<T>(b: Box<T>) -> *mut T { transmute(b) }
 
 pub trait WidgetSignals {
-    fn connect_notify<F: Fn(Widget, &ParamSpec)>(&self, f: F) -> u64;
-    fn connect_accel_closures_changed<F: Fn(Widget)>(&self, f: F) -> u64;
-    fn connect_button_press_event<F: Fn(Widget, &EventButton) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_button_release_event<F: Fn(Widget, &EventButton) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_can_activate_accel<F: Fn(Widget, u64) -> bool>(&self, f: F) -> u64;
-    fn connect_child_notify<F: Fn(Widget, &ParamSpec)>(&self, f: F) -> u64;
-    fn connect_composited_changed<F: Fn(Widget)>(&self, f: F) -> u64;
-    fn connect_configure_event<F: Fn(Widget, &EventConfigure) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_damage_event<F: Fn(Widget, &EventExpose) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_delete_event<F: Fn(Widget, &EventAny) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_destroy<F: Fn(Widget)>(&self, f: F) -> u64;
-    fn connect_destroy_event<F: Fn(Widget, &EventAny) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_direction_changed<F: Fn(Widget, TextDirection)>(&self, f: F) -> u64;
-    fn connect_draw<F: Fn(Widget, Context) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_enter_notify_event<F: Fn(Widget, &EventCrossing) -> Inhibit>(&self, f: F) -> u64;
-    fn connect_event<F: Fn(Widget, &EventAny) -> Inhibit>(&self, f: F) -> u64;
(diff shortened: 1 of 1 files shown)


gkoz:
@oakes Sorry, if this breaks your projects again but we discussed this requirement, I just messed up the execution...

gkoz:
This will fail travis until https://github.com/rust-gnome/examples/pull/5 is merged.

GuillaumeGomez:
Ok, I'll restart travis after example PR is merged.

GuillaumeGomez:
And it's done. I merge it. Thanks !

oakes:
I can't currently test this because I'm getting an error about unresolved imports in `glib`. Do you guys get this too?

gkoz:
@oakes The tests are green. Some of you repos may be out of date.

oakes:
Yep, I fetched but forgot to rebase. My bad.

oakes:
So, the error is that the closures may outlive the data I'm passing to them. It suggests using `move`, but that won't work since I have multiple closures using this data. Any suggestions about how to deal with this?

gkoz:
The `Rc<RefCell<T>>>` trick for one.
I feel like everyone is going to hate me for this. 8-/

oakes:
The errors are occurring with `Rc<RefCell<T>>` objects though. For example, the `ui` object here:

```
src/main.rs:230:35: 233:6 error: closure may outlive the current function, but it borrows `ui`, which is owned by the current function [E0373]
src/main.rs:230     rename_button.connect_clicked(|_| {
src/main.rs:231         ::projects::rename_file(ui.deref().borrow().deref(), prefs.deref().borrow_mut().deref_mut(), write_fd);
src/main.rs:232         should_update_tree.deref().set(true);
src/main.rs:233     });
src/main.rs:231:33: 231:35 note: `ui` is borrowed here
src/main.rs:231         ::projects::rename_file(ui.deref().borrow().deref(), prefs.deref().borrow_mut().deref_mut(), write_fd);
                                                ^~
src/main.rs:230:35: 233:6 help: to force the closure to take ownership of `ui` (and any other referenced variables), use the `move` keyword, as shown:
src/main.rs:        rename_button.connect_clicked(move |_| {
src/main.rs:            ::projects::rename_file(ui.deref().borrow().deref(), prefs.deref().borrow_mut().deref_mut(), write_fd);
src/main.rs:            should_update_tree.deref().set(true);
src/main.rs:        });
```

gkoz:
You clone an `Rc` or a widget and move the clone into the closure.

oakes:
I finally found some time today to work on this. I managed to do all the `Rc` cloning for each closure and add `move` to them, and consequently eliminated the "closure may outlive" errors. However, it is still showing errors when I initially try to create the `ui` object, a struct that stores all my widgets. It says the widgets don't live long enough. For example:

```
src/main.rs:195:18: 195:24 error: `window` does not live long enough
src/main.rs:195         window: &window,
                                 ^~~~~~
note: reference must be valid for the static lifetime...
src/main.rs:129:75: 442:2 note: ...but borrowed value is only valid for the block suffix following statement 74 at 129:74
src/main.rs:129     let window = widgets::Window::new(gtk::WindowType::TopLevel).unwrap();
```

I think I understand conceptually why the errors are happening -- I'm making the widgets at runtime and they are clearly not static. However, I don't see how else to do it.

gkoz:
Are you putting references to widgets in that struct?

oakes:
Yeah, I didn't even think about that. I'll see if I can refactor it to not take references.

oakes:
That worked, thanks. I can't seem to build the most commits of gtk because it can't find `RectangleInt` in cairo, though. My version of rust is probably a week old, if that matters.

gkoz:
You need to either use all the crates from crates.io or from git master, mixing them will often break.

gkoz:
Here's my attempt at a slight improvement for the cloning-and-moving suckiness (compare with the [notebook example](https://github.com/rust-gnome/examples/blob/master/src/notebook.rs#L40-45)):
```Rust
        let env = (self.notebook.clone(), widget.clone());
        button.connect_clicked(move |_| {
            let (ref notebook, ref widget) = env;
            let index = notebook.page_num(widget).unwrap();
            notebook.remove_page(index as i32);
        });
```
You put everything you need in the closure into a tuple which is moved there, and then use pattern matching with `ref` to get references out of it.

oakes:
Ah, I suppose that is a complication in maintaining a fork =) I'll just delay rebasing for now. And yeah it looks like I could cut down some lines by using pattern matching, I'll look into that.

gkoz:
After I complete the refactoring from #59 you should be able to implement the `vte` widgets in a separate crate so the fork shouldn't be necessary.

oakes:
Great, I'll keep an eye on that.

PR #71, opened 2015-05-11
Add GtkFlowBox

csoriano89:
Hi,
So first time contributing in github and with Rust. So probably I made a mess, but let see =)
I attached a patch with the addition of the GtkFlowBox and another patch only intended for testing for review which should not be committed to master.

Also, it would be awesome to have a way to reach you on IRC or something, I have some questions about the general design.

Cheers


GuillaumeGomez:
Hey ! Thanks for your contribution ! You actually can talk on [gitter](https://gitter.im/rust-gnome/gtk). I'm taking a look at your code.

GuillaumeGomez:
All examples are in another repository (rust-gnome/examples to be precise). Except that, your code is good, I have nothing to add. What do you think of it @gkoz ?

csoriano89:
I didn't add all the signals yet, I wanted to have an early preview given my inexperience with Rust and github.

It is necessary to be merged tho right? Or we can add them later. Depending how you normally work.

GuillaumeGomez:
No, I can let this PR open until you say it's ready to be merged. However, I won't merge it until you move the example part into the good repository.

gkoz:
Seems like `gtk::cast` and `gtk_glue.c` bits are missing.
This should be gated on `cfg(feature = "gtk_3_12")` if I'm not mistaken.

GuillaumeGomez:
@gkoz: "GtkFlowBox was added in GTK+ 3.12."

src/traits/mod.rs
```diff
@@ -59,6 +62,7 @@ pub mod misc;
 pub mod button;
 pub mod label;
 pub mod _box;
+pub mod flow_box;
```
gkoz: This should be preceded by `#[cfg(feature = "gtk_3_12")]`.
And I don't see a similar import in `widgets/mod.rs` at all.

GuillaumeGomez:
[FlowBox](https://github.com/gtk-rs/gtk/blob/master/src/auto/flow_box.rs) type is now generated. I close this PR.

PR #73, opened 2015-05-11
Change return type to Option<u32>

mitaa:
These two functions would return -1 on failure. 
Also renamed `get_n_page` to `get_n_pages` (I assume a typo).

### DIFF
--- a/src/widgets/note_book.rs
+++ b/src/widgets/note_book.rs
@@ -7,7 +7,7 @@
 use ffi;
 use cast::GTK_NOTEBOOK;
 use FFIWidget;
-use glib::translate::{from_glib_none, ToGlibPtr};
+use glib::translate::{from_glib, from_glib_none, ToGlibPtr};
 use glib::{to_bool, to_gboolean};
 
 /// GtkNotebook — A tabbed notebook container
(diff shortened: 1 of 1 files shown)


src/widgets/note_book.rs
```diff
@@ -96,9 +96,12 @@ impl NoteBook {
         }
     }
 
-    pub fn get_current_page(&self) -> i32 {
-        unsafe {
+    pub fn get_current_page(&self) -> Option<u32> {
+        match unsafe {
             ffi::gtk_notebook_get_current_page(GTK_NOTEBOOK(self.pointer))
```
gkoz: What do you think of using [this conversion implementation](https://github.com/rust-gnome/glib/blob/master/src/translate.rs#L271-281) to express this as
`from_glib(ffi::gtk_notebook_get_current_page(GTK_NOTEBOOK(self.pointer)))`

mitaa: Oh.. I'd like that.
I followed the same style/pattern as the other trait methods - should I change those too?

GuillaumeGomez:
Thanks for your PR ! Please fix what pointed out @gkoz and then I'll merge it !

mitaa:
Done.

GuillaumeGomez:
Thanks ! Let's now wait for travis and appveyor.

GuillaumeGomez:
Perfect ! Thanks for your work !

Issue #44, opened 2015-04-28
Unable to compile rust-gnome\examples on Win: ld cannot find -limm32


andrewd18:
**OS:** Win 7 64-bit in a brand new VM

**Installed:**
 * rustc 1.1.0-nightly <97d4e76c2 2015-04-27> <built 2015-04-28>
 * GTK 3.6.4 SDK (unzipped, added bin to path)
 * win-builds with the following packages
    - binutils
    - gcc
    - gettext
    - gmp
    - libmpc
    - mingw-w64
    - mpfr
    - win-iconv
    - winpthreads
    - zlib

**PATH order:**
 * win-builds bin folder, GTK bin folder, Rust bin folder

**Steps to Reproduce**
 1. Clone rust-gnome\examples
 2. `cargo build`
 3. `ld: Cannot find -limm32`

Judging by the very long error string, it looks like rust-gnome\gtk is failing because C:\win_builds\x86_64-w64-mingw32\lib32 folder (which has libimm32.a) is not included in the -L search locations.

andrewd18:
```
... "cargo build" plus prereqs snipped thanks to cmd.exe's max buffer ...

   Compiling cairo-sys v0.1.0 (https://github.com/rust-gnome/cairo#93cb13e5)
   Compiling gdk-sys v0.1.0 (https://github.com/rust-gnome/gdk#2b1fc225)
   Compiling glib-sys v0.1.0 (https://github.com/rust-gnome/glib#e8fc0580)
   Compiling pango-sys v0.1.0 (https://github.com/rust-gnome/pango#3ce36f5e)
   Compiling gtk-sys v0.1.0 (https://github.com/rust-gnome/gtk#8d9b0434)
   Compiling glib v0.0.1 (https://github.com/rust-gnome/glib#e8fc0580)
   Compiling cairo v0.0.1 (https://github.com/rust-gnome/cairo#93cb13e5)
   Compiling pango v0.0.1 (https://github.com/rust-gnome/pango#3ce36f5e)
   Compiling gdk v0.0.1 (https://github.com/rust-gnome/gdk#2b1fc225)
   Compiling gtk v0.0.1 (https://github.com/rust-gnome/gtk#8d9b0434)
error: linking with `gcc` failed: exit code: 1
note: "gcc" "-Wl,--enable-long-section-names" "-fno-use-linker-plugin" "-Wl,--nx
[...]

PR #74, opened 2015-05-11
Readme: update Windows building instructions

gkoz:
This better reflects the findings from #44

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,9 @@ Make sure both mingw's and the sdk's `bin` directories are in your `PATH` e.g. (
 ```
 C:\> set PATH="C:\mingw-w64\bin;C:\gtk\bin;%PATH%"
 ```
-It's crucial that GCC from mingw is used by Rust so either make sure that mingw is earlier in the `PATH` or delete `gcc.exe` and `ld.exe` from the Rust installation.
+If your Rust installation has `gcc.exe` and `ld.exe` in its `bin` directory, you may
+get a linking error `ld: cannot find -limm32`. In that case remove those executables,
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ Make sure both mingw's and the sdk's `bin` directories are in your `PATH` e.g. (
 ```
 C:\> set PATH="C:\mingw-w64\bin;C:\gtk\bin;%PATH%"
 ```
-It's crucial that GCC from mingw is used by Rust so either make sure that mingw is earlier in the `PATH` or delete `gcc.exe` and `ld.exe` from the Rust installation.
+You will get a linking error `ld: cannot find -limm32` if `gcc.exe` and `ld.exe` that come with Rust are present in its `bin` directory so make sure they're not there.
```
GuillaumeGomez: "that come__s__", no ?

gkoz: There're two files so plural. I feel though that the sentence may not be clear enough so if you have better ideas, go for it =)

GuillaumeGomez: Hum... I didn't read it like that. I'd replace "that" by "which" in this case. But I'm not an english native (and maybe you are !), so ?

gkoz: I'm not, but I've looked into "that"/"which" question recently. They're completely equivalent, what matters is the presence or absence of a comma before "that".

gkoz:
I'll try to improve the wording later today

gkoz:
There.

GuillaumeGomez:
Yes sir !

GuillaumeGomez:
Way better like this, thanks !

PR #75, opened 2015-05-12
FFI types update

gkoz:


GuillaumeGomez:
That's a lot better like that ! Thanks ! =D

PR #80, opened 2015-05-16
Add stable build

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,7 @@ language: rust
 rust:
 - nightly
 - beta
+- stable
 sudo: true
 env:
   global:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: rust
 rust:
 - nightly
-- beta
 - stable
 sudo: true
 env:


gkoz:
I guess we should discuss our strategy with regard to various existing Rust branches.
BTW, one option I hadn't considered before could be tracking only stable and beta.

Keeping with the status quo would be changing beta to stable so we would be targeting 1.0-release and 1.2-nightly (there should've been 1.1-beta released yesterday and the nightly should've become 1.2 but that part may have been postponed for Monday or whatever).
Surely we won't abandon 1.0 today, as for 1.1 I'm not sure of its significance yet but whatever builds with 1.0 should be compatible with that as well.

GuillaumeGomez:
Just keeping beta and stable ? Hum... I'm afraid that would take a look at travis would open issue to say "Is it working on the stable version ? I didn't see it building with the last stable version of rust on travis" or stuff like that. I was thinking about that when I created this PR. If you really don't see it as useful, then let's just close this issue.

gkoz:
Beta and stable is just an idea.
Beta isn't too useful now because it was a surrogate for missing stable.
I'd leave just stable and nightly for now.

GuillaumeGomez:
Okay, let's do that for now then. I merge. I'll do that for all other libraries as well.

gkoz:
You'll probably want to do the same to appveyor.yml

GuillaumeGomez:
Can you take at it then ? It'll be faster that way.

gkoz:
Okay

GuillaumeGomez:
Thanks !

Issue #83, opened 2015-05-19
Can't build SolidOak because of some error with Gtk


vi:
```
Build failed, waiting for other jobs to finish...

failed to run custom build command for `gtk-sys v0.1.0 (https://github.com/oakes/gtk#b9b3d2bc)`
Process didn't exit successfully: `/home/vi/home/rust/SolidOak/target/debug/build/gtk-sys-4fc3492795cbacc8/build-script-build` (exit code: 101)
--- stdout
cargo:rustc-link-search=native=/usr/lib/i386-linux-gnu
cargo:rustc-link-lib=gtk-3
cargo:rustc-link-lib=gdk-3
cargo:rustc-link-lib=atk-1.0
cargo:rustc-link-lib=gio-2.0
cargo:rustc-link-lib=pangocairo-1.0
cargo:rustc-link-lib=gdk_pixbuf-2.0
cargo:rustc-link-lib=cairo-gobject
cargo:rustc-link-lib=pango-1.0
cargo:rustc-link-lib=cairo
cargo:rustc-link-lib=gobject-2.0
cargo:rustc-link-lib=glib-2.0
cargo:rustc-link-search=native=/usr/lib/i386-linux-gnu
cargo:rustc-link-search=native=/usr/lib
cargo:rustc-link-lib=vte2_90
cargo:rustc-link-lib=gtk-3
cargo:rustc-link-lib=gdk-3
cargo:rustc-link-lib=atk-1.0
cargo:rustc-link-lib=pangocairo-1.0
cargo:rustc-link-lib=gdk_pixbuf-2.0
cargo:rustc-link-lib=cairo-gobject
cargo:rustc-link-lib=pango-1.0
cargo:rustc-link-lib=gio-2.0
cargo:rustc-link-lib=gobject-2.0
cargo:rustc-link-lib=glib-2.0
cargo:rustc-link-lib=cairo
cargo:rustc-link-lib=X11
TARGET = Some("i686-unknown-linux-gnu")
CARGO_MANIFEST_DIR = Some("/home/vi/home/rust/.cargo/git/checkouts/gtk-38cf3f4d9926c12a/master/gtk-sys")
OUT_DIR = Some("/home/vi/home/rust/SolidOak/target/debug/build/gtk-sys-4fc3492795cbacc8/out")
TARGET = Some("i686-unknown-linux-gnu")
OPT_LEVEL = Some("0")
PROFILE = Some("debug")
debug 0
[...]

PR #86, opened 2015-05-19
Gate LevelBar cast on gtk_3_6

gkoz:
cc #83


GuillaumeGomez:
Thanks ! Before I merge, is there any other missing ?

gkoz:
According to https://github.com/rust-gnome/gtk/issues/83#issuecomment-103460902 this was the only problem as long as you make sure `gtk_3_6` isn't enabled.

GuillaumeGomez:
I was speaking about other missing defined. ^^

gkoz:
The recursive cargo features should be taking care of that, same as in the Rust sources. Were they not I wouldn't expect the builds to succeed.

GuillaumeGomez:
Welll, we'll see if someone else finds something.

PR #88, opened 2015-05-19
Remove doc export

GuillaumeGomez:


.travis.yml
```diff
@@ -20,5 +17,3 @@ script:
   - cargo build
   - cd ..
   - cargo doc
```
gkoz: I'm thinking we might want to remove this line as well, otherwise that dance with checking out everything from git would have to be applied to the current crate as well

GuillaumeGomez: Yes, now I think about it it's quite useless now.

gkoz:
Oh, and if it's not too much work, while you're modifying the travis files, would you please change the `--depth` argument to git from `1` to `50`? :) Putting `1` there was silly of me...

GuillaumeGomez:
No problem.

GuillaumeGomez:
Do you want me to do the same with the other travis files as well ?

gkoz:
Yeah, it's good.

GuillaumeGomez:
Is that a "yes change others too" or "no just this one, that's good" ? ^^'

gkoz:
:) The others too please.

GuillaumeGomez:
Ok haha.

PR #94, opened 2015-05-31
Added gtk::init_check()

cjgreenaway:
A small change to make gtk_init_check() available through gtk::init_check() - returning a bool to report whether GTK has been initialised successfully.

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -8,6 +8,9 @@ use ffi;
 use glib::translate::{from_glib_none};
 use glib::{to_bool, to_gboolean};
 
+/// Call this function before using any other GTK+ functions.
+/// Note that this function calls gtk_init_check() rather than gtk_init(),
+/// so will not cause the program to terminate if GTK could not be initialized.
 pub fn init() -> Result<(), ()> {
     unsafe {
 	match to_bool(ffi::gtk_init_check(ptr::null(), ptr::null())) {
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -14,6 +14,10 @@ pub fn init() {
     }
 }
 
+pub fn init_check() -> bool {
+    unsafe { to_bool(ffi::gtk_init_check(ptr::null(), ptr::null())) }
+}
+
 pub fn main() {
     unsafe {
         ffi::gtk_main();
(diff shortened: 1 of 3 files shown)


gkoz:
Thanks for bringing this up.

What if we took this one step further though and instead of adding `init_check` changed `init` so it would call `gtk_init_check` internally and return `Result<(), ()>`?

cjgreenaway:
I like the idea of returning Result<(), ()>. I was trying to think of a way to do this (something more Rusty than true or false), but I'm a relative beginner in Rust and I didn't know you could use that double bracket syntax to say the return value would be Ok / Err without values.

I'm more wary of changing the behaviour of gtk::init(). gtk_init() causes the application to exit if initialisation fails. This will no longer happen if we are using gtk_init_check to return a Result. This is probably a good thing, but it depends on how much backwards compatibility we want and also how close we want the rust-gnome API to be to the GTK API.

Anyway, I've made both of those changes and pushed them to the branch.

GuillaumeGomez:
@cjgreenaway: Thanks for your PR ! The code is also good for me.
@gkoz: Is that good for you too or do you see something else ?

gkoz:
> I'm more wary of changing the behaviour of gtk::init(). gtk_init() causes the application to exit if initialisation fails. This will no longer happen if we are using gtk_init_check to return a Result.

They'll get a clue -- `warning: unused result which must be used`.

But you're right, when making a departure from the GTK API we have to document the new behavior. So would you please add some documentation to the function?

Other than that looks good to me.

cjgreenaway:
Added a comment as requested.

src/rt.rs
```diff
@@ -8,9 +8,15 @@ use ffi;
 use glib::translate::{from_glib_none};
 use glib::{to_bool, to_gboolean};
 
-pub fn init() {
+/// Call this function before using any other GTK+ functions.
+/// Note that this function calls gtk_init_check() rather than gtk_init(),
```
gkoz: Can you check the rendered docs? I seem to recall there needs to be an empty comment line after the summary...

Also being more clear about what the `Err` variant means might be nice.

GuillaumeGomez: To make a backline, it does need to have an empty comment line. Example:

```Rust
/// line1
/// line1
/// 
/// line2
/// line2
```

cjgreenaway:
Changed as requested - although there's very little extra information we can infer from an Err other than failure.

gkoz:
Great! :+1:

GuillaumeGomez:
Perfect, thanks for your work ! I merge.

PR #98, opened 2015-06-04
Added move_, set_type_hint, and get_window methods

gsingh93:
Note that `move` was renamed to `move_`, as `move` is a keyword. Adding an underscore has been the trend I've seen in other projects when this name collision happens.

### DIFF
--- a/src/traits/widget.rs
+++ b/src/traits/widget.rs
@@ -360,6 +360,12 @@ pub trait WidgetTrait: ::FFIWidget + ::GObjectTrait {
         unsafe { to_bool(ffi::gtk_widget_get_double_buffered(self.unwrap_widget())) }
     }
 
+    fn get_window(&self) -> Option<gdk::Window> {
+        unsafe {
+            from_glib_none(ffi::gtk_widget_get_window(self.unwrap_widget()))
+        }
+    }
+
     fn get_has_window(&self) -> bool {
         unsafe { to_bool(ffi::gtk_widget_get_has_window(self.unwrap_widget())) }
     }
--- a/src/traits/window.rs
+++ b/src/traits/window.rs
@@ -6,8 +6,21 @@ use glib::translate::{from_glib_none, ToGlibPtr};
 use ffi;
 use glib::to_gboolean;
 use cast::GTK_WINDOW;
+use gdk;
 
 pub trait WindowTrait : ::WidgetTrait {
+    fn move_(&self, x: i32, y: i32) {
+        unsafe {
+            ffi::gtk_window_move(GTK_WINDOW(self.unwrap_widget()), x, y);
+        }
+    }
+
+    fn set_type_hint(&self, hint: gdk::WindowTypeHint) {
+        unsafe {
+            ffi::gtk_window_set_type_hint(GTK_WINDOW(self.unwrap_widget()), hint);
+        }
+    }
+
     fn set_title(&self, title: &str) -> () {
         unsafe {
             ffi::gtk_window_set_title(GTK_WINDOW(self.unwrap_widget()), title.to_glib_none().0);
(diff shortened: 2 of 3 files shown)


src/traits/window.rs
```diff
@@ -6,8 +6,21 @@ use glib::translate::{from_glib_none, ToGlibPtr};
 use ffi;
 use glib::to_gboolean;
 use cast::GTK_WINDOW;
+use gdk;
 
 pub trait WindowTrait : ::WidgetTrait {
+    fn move_(&self, x: i32, y: i32) {
```
GuillaumeGomez: I usually directly replace c_int with isize since isize is architecture dependent too.

gsingh93: But `c_int` is an `i32`: http://doc.rust-lang.org/libc/libc/types/os/arch/c95/type.c_int.html

GuillaumeGomez: That's super weird ! It shouldn't !

gsingh93: `int`s in C are never architecture dependent. Are you thinking of `long`s?

GuillaumeGomez: Not at all. Otherwise, why uint32_t and similar types would have been created ? Some links :
 * http://stackoverflow.com/questions/2533319/c-variable-type-sizes-are-machine-dependent-is-it-really-true-signed-unsig
 * http://www.nongnu.org/c-prog-book/online/x401.html
 * http://www.linuxquestions.org/questions/programming-9/int-size-is-compiler-dependent-or-os-dependent-915923/

gkoz: Generally `int`s might be architecture dependent and that must be why there exists `c_int` (so `isize` would be a mistake).

gsingh93: I guess it's possible for the `int` type to change, but all of the compilers/systems I've used equate `int` to `i32` and `long` to `isize`.

GuillaumeGomez: @gsingh93: Don't worry, I just thought about it when I read your code. =D
@gkoz: Where did you find the `isize` size ? I thought it was supposed to be the "replacement" for C `int` types. I'm quite surprised.

gkoz: [isize](http://doc.rust-lang.org/std/primitive.isize.html). That page is somewhat out of date and confusing because it talks about Rust's `int` type which is different from C `int`.

`usize`/`isize` (former `uint`/`int`) are guaranteed to fit a pointer while C's `int` is on all current platforms equivalent to `i32`.

gsingh93:
Just pushed a change to use `from_glib_none` instead of doing the null checks and conversion myself.

src/traits/widget.rs
```diff
@@ -360,6 +360,13 @@ pub trait WidgetTrait: ::FFIWidget + ::GObjectTrait {
         unsafe { to_bool(ffi::gtk_widget_get_double_buffered(self.unwrap_widget())) }
     }
 
+    fn get_window(&self) -> Option<gdk::Window> {
+        unsafe {
+            let tmp = ffi::gtk_widget_get_window(self.unwrap_widget());
+            from_glib_none(tmp as *mut _)
```
gkoz: Hmm, this shouldn't require any casts. Does the build fail without the cast?

gsingh93: Oops, the old code needed it but this doesn't. I've removed it.

src/traits/widget.rs
```diff
@@ -360,6 +360,13 @@ pub trait WidgetTrait: ::FFIWidget + ::GObjectTrait {
         unsafe { to_bool(ffi::gtk_widget_get_double_buffered(self.unwrap_widget())) }
     }
 
+    fn get_window(&self) -> Option<gdk::Window> {
+        unsafe {
+            let tmp = ffi::gtk_widget_get_window(self.unwrap_widget());
+            from_glib_none(tmp)
```
gkoz: The temporary variable isn't of much use now so let's remove it here and in the other PR to keep the style consistent with other crates.

gsingh93:
Fixed.

gkoz:
Looks good :+1:

GuillaumeGomez:
Everything's good, I merge. Thanks for your work !

PR #100, opened 2015-06-07
Updated README.md to respect project/caret name

umurgdk:
Caret's name is "gtk" not "rgtk". Cargo can't find caret named "rgtk" when I tried.


GuillaumeGomez:
Totally, thanks for your PR !

PR #103, opened 2015-06-14
Add screenshots

GuillaumeGomez:


Anachron:
Wouldn't it be better to put the screenshots inside the example repository?

GuillaumeGomez:
I was wondering if it was better but then I thought that people would get on this repo first. Showing them screenshots in here made more sense in my mind. However, we can add all the screenshots in examples.

@gkoz: Do you have an opinion on this ?

Anachron:
@GuillaumeGomez maybe it would be enough to make a link to the examples directory, saying there wuold be some screenshots for those wondering?

GuillaumeGomez:
The fewer clicks the better for me. But once again, I'm totally open to proposition. If other people prefer your solution, then we'll like you say. ;-)

Anachron:
Fair enough!

gkoz:
Couldn't we put the screenshots into the examples repo and link to them from the readmes, so @GuillaumeGomez wouldn't have to host them?

GuillaumeGomez:
@gkoz: I have unlimited space so it's not an issue here. ;-)

gkoz:
It's just separation of concerns. This way you don't need to remember that some readmes are dependent on your site to work so there's less harm if it goes down for two weeks or in a year's time you decide to change the directory structure.

GuillaumeGomez:
Actually, github create a copy on its server of the file. The link in the readme is just for github's generation, so even if I removed the file from the server, it'd still be "visible" on github. But if you find another solution, that's totally fine by me ! =D

gkoz:
It might be acting as a proxy for protection. I bet if your server returns 404, the images will disappear. :p

GuillaumeGomez:
I tried to change the image and it didn't work in github, so I think they store it.

PR #104, opened 2015-06-16
Add CSS styling support

gsingh93:
This PR adds CSS styling support by implementing `CssProvider`, `StyleProvider`, and `StyleContext`. An approach similar to that used with `WidgetTrait` was taken. `get_screen` is also implemented for `WidgetTrait` so this screen can be used with `StyleContext::add_provider_for_screen`.

There are some TODOs in the code for adding the proper `#![feature]` attributes. The Gtk documentation doesn't give this number for every method/class, and I think this is due to a lack of documentation, not the fact that the method/class was introduced before gtk3. How should I find these feature numbers?

### DIFF
--- a/src/traits/style_provider.rs
+++ b/src/traits/style_provider.rs
@@ -2,12 +2,27 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+use ffi;
 use glib::FFIGObject;
 
-pub const PRIORITY_FALLBACK: u32 = 1;
-pub const PRIORITY_THEME: u32 = 200;
-pub const PRIORITY_SETTINGS: u32 = 400;
-pub const PRIORITY_APPLICATION: u32 = 600;
-pub const PRIORITY_USER: u32 = 800;
+/// The priority used for default style information that is used in the absence of themes.
+/// Note that this is not very useful for providing default styling for custom style classes,
+/// themes are likely to override styling provided at this priority with catch-all * {...} rules.
+pub const STYLE_PROVIDER_PRIORITY_FALLBACK: u32 = ffi::GTK_STYLE_PROVIDER_PRIORITY_FALLBACK;
+
+/// The priority used for style information provided by themes.
+pub const STYLE_PROVIDER_PRIORITY_THEME: u32 = ffi::GTK_STYLE_PROVIDER_PRIORITY_THEME;
+
+/// The priority used for style information provided via GtkSettings.
+/// This priority is higher than STYLE_PROVIDER_PRIORITY_THEME to let settings override themes.
+pub const STYLE_PROVIDER_PRIORITY_SETTINGS: u32 = ffi::GTK_STYLE_PROVIDER_PRIORITY_SETTINGS;
+
+/// A priority that can be used when adding a GtkStyleProvider for application-specific style
+/// information.
+pub const STYLE_PROVIDER_PRIORITY_APPLICATION: u32 = ffi::GTK_STYLE_PROVIDER_PRIORITY_APPLICATION;
+
+/// A priority that can be used when adding a GtkStyleProvider for application-specific style
+/// information.
+pub const STYLE_PROVIDER_PRIORITY_USER: u32 = ffi::GTK_STYLE_PROVIDER_PRIORITY_USER;
 
 pub trait StyleProviderTrait : FFIGObject {}
--- a/src/traits/style_provider.rs
+++ b/src/traits/style_provider.rs
@@ -1,0 +1,13 @@
+// Copyright 2013-2015, The Rust-GNOME Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib::FFIGObject;
+
+pub const PRIORITY_FALLBACK: u32 = 1;
+pub const PRIORITY_THEME: u32 = 200;
+pub const PRIORITY_SETTINGS: u32 = 400;
+pub const PRIORITY_APPLICATION: u32 = 600;
+pub const PRIORITY_USER: u32 = 800;
+
+pub trait StyleProviderTrait : FFIGObject {}
--- a/src/widgets/css_provider.rs
+++ b/src/widgets/css_provider.rs
@@ -1,0 +1,56 @@
+// Copyright 2013-2015, The Rust-GNOME Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use std::fmt::{self, Display, Formatter};
+use ffi::{self, GtkCssProvider};
+use glib::translate::{ToGlibPtr, from_glib_full};
+use glib::{self, GlibContainer};
+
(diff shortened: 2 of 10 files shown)


GuillaumeGomez:
Very nice work ! =D
@gkoz: Code looks fine for me, I'm waiting your review. ;-)

src/widgets/css_provider.rs
```diff
@@ -0,0 +1,64 @@
...
+        unsafe {
+            CssProvider { pointer: ffi::gtk_css_provider_get_named(name.to_glib_none().0,
+                                                                   variant.to_glib_none().0) }
+        }
+    }
+
+    pub fn load_from_path(path: &str) -> Result<CssProvider, glib::Error> {
+        unsafe {
+            let pointer = ffi::gtk_css_provider_new();
+            let mut error: *mut glib::ffi::GError = ::std::mem::zeroed();
```
gkoz: This can simply be `let mut error = ptr::null_mut();`

src/widgets/style_context.rs
```diff
@@ -0,0 +1,43 @@
...
+    // FIXME: This doesn't need to be public after object reform lands
+    pub pointer: *mut GtkStyleContext
+}
+
+impl StyleContext {
+    pub fn new() -> Self {
+        unsafe { StyleContext { pointer: ffi::gtk_style_context_new() } }
+    }
+
+    pub fn add_provider<T: ::StyleProviderTrait>(&self, provider: T,
```
gkoz: You probably want `provider: &T` so it won't be consumed.

src/widgets/css_provider.rs
GuillaumeGomez: It __should__ be `let mut error = ptr::null_mut();`.

src/widgets/style_context.rs
```diff
@@ -0,0 +1,43 @@
...
+                                                 priority: StyleProviderPriority) {
+        unsafe {
+            ffi::gtk_style_context_add_provider(self.pointer, provider.unwrap_provider(),
+                                                priority)
+        };
+    }
+
+    pub fn add_provider_for_screen<T: ::StyleProviderTrait>(screen: &gdk::Screen, provider: T,
```
gkoz: `provider: &T`

gkoz:
Since you use `impl_GObjectFunctions!` I don't think there's a need for `FFIStyleProvider`.

The reference counting isn't handled well but I guess this will be fixed by the objects reform.

If the minimum GTK version for a method or class isn't specified it's safe to assume that it's present in GTK 3.4, which is the lowest version we support (also 'since 3.0' and 'since 3.2' can be ignored).

src/widgets/css_provider.rs
gsingh93: Ah, I was looking for how to do that. I saw `ptr::null()`, but missed `ptr::null_mut()`.

gsingh93:
Ok, fixed.

One more question: The `gtk_style_context_add_provider` method takes a `guint` method but I pass in an enum. Could the size mismatch cause a problem?

gkoz:
> The gtk_style_context_add_provider method takes a guint method but I pass in an enum. Could the size mismatch cause a problem?

I believe a lot of the definitions currently assume that `sizeof(c_enum) == sizeof(c_int)` but here we can avoid that assumption by making the ffi type `c_uint` and casting the enum to `c_uint` in `add_provider`. What if someone wanted to pass a value in between the constants (e.g. `117`) though?

gsingh93:
> What if someone wanted to pass a value in between the constants (e.g. 117) though?

Should I add another variant called `CustomPriority(u32)`?

gkoz:
> Should I add another variant called CustomPriority(u32)?
It would not be a C-like enum any more so you'd lose any conveniences and some derived traits. And anyone could put `100` or `200` there etc. Maybe simply using constants is the way. Or we could leave the type as `u32` and tell the users to do the casting themselves if they want predefined values.

gsingh93:
Ok, I think I'll change the type to a `u32` and convert the enum into constants.

gsingh93:
Ok, I added the constants. You can access them with `gtk::traits::style_provider::PRIORITY_USER`, as I decided not to export them.

gsingh93:
I also just added a commit that documents the priorities. I copied the documentation directly from the Gtk docs.

src/traits/style_provider.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2013-2015, The Rust-GNOME Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use glib::FFIGObject;
+
+/// The priority used for default style information that is used in the absence of themes.
+/// Note that this is not very useful for providing default styling for custom style classes,
+/// themes are likely to override styling provided at this priority with catch-all * {...} rules.
+pub const PRIORITY_FALLBACK: u32 = 1;
```
gkoz: The idiomatic approach would be to define `GTK_STYLE_PROVIDER_PRIORITY_*` constants in the `sys` crate (those definitions will become autogenerated eventually) and use those on the right hand side here.

gkoz:
> You can access them with gtk::traits::style_provider::PRIORITY_USER, as I decided not to export them.

It looks like we're going to reexport all symbols in the top namespace and the names would have to be longer in that case, e.g. `STYLE_PROVIDER_PRIORITY_USER`.

We may want to investigate using associated constants -- they seem to be available in the nightlies. Example: https://github.com/rust-lang/rust/blob/master/src/test/run-pass/associated-const-inherent-impl.rs

gsingh93:
> It looks like we're going to reexport all symbols in the top namespace and the names would have to be longer in that case, e.g. STYLE_PROVIDER_PRIORITY_USER

Why not just use proper namespaces?

> We may want to investigate using associated constants -- they seem to be available in the nightlies. Example: https://github.com/rust-lang/rust/blob/master/src/test/run-pass/associated-const-inherent-impl.rs

StyleProvider is a trait here, so it wouldn't work. Maybe after object reform?

gsingh93:
> We may want to investigate using associated constants -- they seem to be available in the nightlies. Example: https://github.com/rust-lang/rust/blob/master/src/test/run-pass/associated-const-inherent-impl.rs
> > StyleProvider is a trait here, so it wouldn't work. Maybe after object reform?

Err, actually if you can provide default values for constants (which wasn't in the documentation), than we could do this.

gkoz:
> Why not just use proper namespaces?

Some reasons here: https://github.com/rust-gnome/glib/pull/47#issuecomment-112803730

It may just be a fool's errand to try to add structure to all of GTK because it's too diverse. While converting every module to the new framework I find it very difficult to organize them coherently. An arbitrary mediocre structure may be worse that the flat convention.

gkoz:
I guess, we could also just have a flat series of modules (`gtk::widget`, `gtk::button`, `gtk::style_provider`, etc) and have constants live there. I'm not enthusiastic about this approach.

gsingh93:
Ok, I just left them as normal constants.

gkoz:
I guess we can reexport these consts in the root module now.

Otherwise seems okay to merge. Small style nits I'd just tweak in the process of refactoring.

gsingh93:
Oops, forgot to reexport. Fixed.

gkoz:
:+1: 
@GuillaumeGomez

GuillaumeGomez:
@gkoz: Thanks for taking care of this one and helping him !
@gsingh93: Thanks for your work !

PR #105, opened 2015-06-18
Added EntrySignals

gsingh93:


src/signal.rs
```diff
@@ -18,9 +18,9 @@ use gdk::{
 use cairo::{Context, RectangleInt};
 
 use {
-    Adjustment, Button, Dialog, DirectionType, Range, ScrollType, SpinButton, StateFlags,
-    TextDirection, ToolButton, Tooltip, TreeIter, TreePath, TreeSelection, TreeView,
-    TreeViewColumn, Widget, WidgetHelpType,
+    Adjustment, Button, DeleteType, Dialog, DirectionType, Entry, MovementStep,
```
gsingh93: Should I put each of these on their own lines to avoid this much churn in the future?

GuillaumeGomez: Yes please.

gsingh93: Ok, done.

GuillaumeGomez:
Perfect, thanks !
@gkoz: Is it good for you too ?

src/signal.rs
```diff
@@ -667,6 +696,135 @@ mod widget {
...
+pub trait EntrySignals {
+    fn connect_activate<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_backspace<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_copy_clipboard<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_cut_clipboard<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_paste_clipboard<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_toggle_overwrite<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
+    fn connect_delete_from_cursor<F: Fn(Entry, DeleteType, i32) + 'static>(&self, f: F) -> u64;
+    fn connect_move_cursor<F: Fn(Entry, MovementStep, i32, bool) + 'static>(&self, f: F) -> u64;
+    fn connect_insert_at_cursor<F: Fn(Entry, &str) + 'static>(&self, f: F) -> u64;
```
gkoz: The idea to use `&str` instead of `String`s is nice but I'm not sure what the lifetime of this `&str` is. Need to investigate.

gkoz: Turns out `Fn(&str)` is sugar for `for<'r> Fn(&'r str)` which means that the closure is generic over the lifetime of its argument, and (as I see it) since it has to accept arbitrarily short `'r`s it's safe to pass in our short-lived reference.

GuillaumeGomez: @gkoz: Good to know.

gkoz:
It's good to merge but next we should add a `FromGlibPtr` implementation for `&str` to avoid boilerplate.
@GuillaumeGomez

GuillaumeGomez:
@gkoz: Okay !
@gsingh93: Thanks for your work !

PR #108, opened 2015-06-27
Add ComboBox signals

mitaa:


GuillaumeGomez:
@mitaa: Thanks for your PR !
@gkoz: We should really write a macro to generate such code, it's almost always the same...

gkoz:
:+1: 

@GuillaumeGomez We might be able to automate this anyway.

GuillaumeGomez:
Okay, if I have time, I'll do it in the week. For now I merge this one.

PR #120, opened 2015-07-19
Replace features with cfg and automatic version detection

gkoz:
The user is no longer required to set the target library version, the builds script detects it from `pkg-config` output and sets the appropriate `cfg` directives.


gkoz:
cc @EPashkin

GuillaumeGomez:
Thanks !

PR #123, opened 2015-07-29
Use the glib::translate API with cairo::Context

gkoz:


GuillaumeGomez:
Thanks a lot. I wait for appveyor status and then I merge !

PR #129, opened 2015-08-13
Use the new sys crates

gkoz:


gkoz:
@GuillaumeGomez I'm afraid our current CI can't check these properly so we'd have to merge and then see if nothing's broke. But take a look at Cargo.tomls at least, maybe you'll see some problems i didn't.

GuillaumeGomez:
And here's the last one ! Thanks a lot ! I'll update the crates in the week-end (or maybe you want me to wait before ?).

gkoz:
I'm considering putting up a page or small blog that would communicate the breaking and simply important changes. If we do that, it should precede the release of the new crate versions.

GuillaumeGomez:
You want me to host it (something like gnome.guillaume-gomez.fr) or you want to create a specific url ? Or maybe we could host it directly on a repo in rust-gnome (using github.io pages). What do you prefer ?

gkoz:
Github pages seems a good candidate. We'll probably use them for the docs as well.

GuillaumeGomez:
Perfect ! I create a "blog" repo on rust-gnome ?

gkoz:
It's probably a good idea to play in a personal repo first, see what works. I've never done this kind of publishing before.

GuillaumeGomez:
Oh ? I let you experiment then ! You might want to take a look at the c_vec published pages as well maybe.

gkoz:
I'll take a look, thanks.

PR #130, opened 2015-08-13
Remove features from Cargo.toml

gkoz:
The build scripts already detect the library version automatically.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,14 +17,6 @@ keywords = ["gtk", "gnome", "GUI"]
 [lib]
 name = "gtk"
 
-[features]
-gtk_3_4 = []
-gtk_3_6 = []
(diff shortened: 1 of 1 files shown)


gkoz:
Interesting...

GuillaumeGomez:
What's interesting ?

gkoz:
There's a lot of breakage and I didn't notice some of it because of checking out the wrong examples branch, but still the errors like this are weird
```
unresolved import `self::button::Button`. There is no `Button` in `widgets::button`
```

EPashkin:
I got similar error on circular modules dependencies.

gkoz:
@EPashkin I've repaired almost everything now (there's still glib::Date breakage on windows). Don't forget to cargo update.

PR #131, opened 2015-08-13
Remove stray #[cfg]

gkoz:
This should make travis build sucessfully. There also was an appveyor error in glib...


GuillaumeGomez:
I'll rerun tests after this merge then.

PR #133, opened 2015-08-14
Fix some integer casts incompatible with the new gtk-sys

gkoz:


gkoz:
The only error seems to be in `examples` now.

PR #135, opened 2015-08-17
Fix build error (missing cfg 3_12)

GuillaumeGomez:


src/widgets/places_sidebar.rs
```diff
@@ -42,19 +42,23 @@ impl PlacesSidebar {
...
     pub fn set_local_only(&self, local_only: bool) {
         unsafe { ffi::gtk_places_sidebar_set_local_only(GTK_PLACES_SIDEBAR(self.unwrap_widget()), to_gboolean(local_only)) }
     }
 
+    #[cfg(gtk_3_12)]
     pub fn get_local_only(&self) -> bool {
         unsafe { to_bool(ffi::gtk_places_sidebar_get_local_only(GTK_PLACES_SIDEBAR(self.unwrap_widget()))) }
     }
 
+    #[cfg(gtk_3_12)]
```
gkoz: This and the following one probably had to be 3_14.

GuillaumeGomez: Arf, my bad. I thought it was all the same pack. I update.

PR #138, opened 2015-08-21
Add TreeSortable trait

GuillaumeGomez:
@gkoz: Ping ! (It is the last one for today ^^)

### DIFF
--- a/src/widgets/text_buffer.rs
+++ b/src/widgets/text_buffer.rs
@@ -6,7 +6,10 @@ use std::ptr;
 
 use ffi;
 use FFIWidget;
+use TextIter;
 
+use cast::GTK_TEXT_BUFFER;
+use glib::translate::ToGlibPtrMut;
 /// GtkTextBuffer — Stores attributed text for display in a GtkTextView
 
 struct_Widget!(TextBuffer);
@@ -21,6 +24,16 @@ impl TextBuffer {
         };
 
         check_pointer!(tmp_pointer, TextBuffer)
+    }
+
+    pub fn get_start_iter(&self) -> TextIter {
+        unsafe {
+            let mut target_iter = TextIter::uninitialized();
+            ffi::gtk_text_buffer_get_start_iter(
+                GTK_TEXT_BUFFER(self.unwrap_widget()), 
+                target_iter.to_glib_none_mut().0);
+            target_iter
+        }
     }
 }
 


gkoz:
Can you believe it `join` still has not made it into a release yet. It's probably on the 1.3 train. We'll have to wait a bit more.

GuillaumeGomez:
Oh come on... It's so annoying ! ><

GuillaumeGomez:
@gkoz: I added tree sortable trait. An example can be found [here](https://github.com/GuillaumeGomez/process-viewer/pull/4). However, it seems there is an issue with too many unref called from gtk like we had previously. I'm still looking to find the issue.

gtk-sys/src/enums.rs
```diff
@@ -12,7 +12,7 @@
 /// the window manager won't decorate the window with a frame, many GTK+ features that rely on the window manager will not work
 /// (e.g. resize grips and maximization/minimization).
 ///
-/// GGtkWindowPopUp is used to implement widgets such as gtk::Menu
+/// GtkWindowPopUp is used to implement widgets such as gtk::Menu
```
gkoz: Umm, this file is not used now. My bad for not deleting it.

gtk-sys/src/lib.rs
```diff
@@ -2058,7 +2058,7 @@ pub type GtkTickCallback = Option<unsafe extern "C" fn(*mut GtkWidget, *mut gdk:
 pub type GtkTranslateFunc = Option<unsafe extern "C" fn(*const c_char, gpointer) -> *mut c_char>;
 pub type GtkTreeCellDataFunc = Option<unsafe extern "C" fn(*mut GtkTreeViewColumn, *mut GtkCellRenderer, *mut GtkTreeModel, *mut GtkTreeIter, gpointer)>;
 pub type GtkTreeDestroyCountFunc = Option<unsafe extern "C" fn(*mut GtkTreeView, *mut GtkTreePath, c_int, gpointer)>;
-pub type GtkTreeIterCompareFunc = Option<unsafe extern "C" fn(*mut GtkTreeModel, *mut GtkTreeIter, *mut GtkTreeIter, gpointer) -> c_int>;
+pub type GtkTreeIterCompareFunc = unsafe extern "C" fn(*mut GtkTreeModel, *mut GtkTreeIter, *mut GtkTreeIter, gpointer) -> c_int;
```
gkoz: It's wrapped in `Option` because Rust's `fn` is not nullable while the C function pointers are.
Now I realize, this might not be the best choice of where to put that `Option` -- it might be better at the use sites because that's where `nullable` annotations are.

gtk-sys/src/lib.rs
```diff
@@ -11766,9 +11766,9 @@ extern "C" {
     pub fn gtk_tree_sortable_get_type() -> GType;
     pub fn gtk_tree_sortable_get_sort_column_id(sortable: *mut GtkTreeSortable, sort_column_id: *mut c_int, order: *mut GtkSortType) -> gboolean;
     pub fn gtk_tree_sortable_has_default_sort_func(sortable: *mut GtkTreeSortable) -> gboolean;
-    pub fn gtk_tree_sortable_set_default_sort_func(sortable: *mut GtkTreeSortable, sort_func: GtkTreeIterCompareFunc, user_data: gpointer, destroy: glib::GDestroyNotify);
+    pub fn gtk_tree_sortable_set_default_sort_func(sortable: *mut GtkTreeSortable, sort_func: *mut c_void, user_data: gpointer, destroy: *mut c_void/*glib::GDestroyNotify*/);
```
gkoz: Can't do that, sorry. ;)
With the temporary `cast_*` exception, editing the `sys` files doesn't make sense. If a mistake is found there, we should fix the generator instead. But it doesn't seem like there's a mistake here.

gtk-sys/src/enums.rs
GuillaumeGomez: I remove it then.

gtk-sys/src/lib.rs
GuillaumeGomez: I don't even use it actually.

gtk-sys/src/lib.rs
GuillaumeGomez: There is not, it's a workaround to compile.

gkoz:
The `TreeSortable` struct is redundant. I recall there was a `FFIGObject` trait or something. A trait that has `unwrap` and `wrap` methods. You can make `TreeSortableTrait` inherit `FFIGObject` that one and implement `FFIGObject` for `ListStore` (it already has the same two methods anyway). Rinse and repeat for `TreeStore`.

GuillaumeGomez:
Oh right, I forgot this. I update.

GuillaumeGomez:
FFIGObject has `unwrap_gobject` and `wrap_object` methods, it's not implemented on any of gtk objects (weird). However, `ListStore` does have `wrap_pointer` and `unwrap_pointer` methods. I'll go on this either.

gkoz:
Yeah the widgets have their own  trait (`FFIWidget` is it?) of possibly different heritage.

I know it's frustrating to just sit wait for the object reform to land. After we deal with the org issues and publishing the fresh crates, I'll concentrate on finishing that. Hopefully with half the classes made by the generator.

GuillaumeGomez:
For now, I'll pause this work if you don't mind. It's a shame, I wanted to add column sorting on the process-viewer. :'(

gkoz:
I guess I'm fine either way. I don't expect this kind of change to hinder the work on that branch.
But a part of this work would be a wasted effort. (Designing a good callback API would still be useful. That would involve studying all three kinds of callbacks there exist in the glib world, some synchronous, others not.)

PR #144, opened 2015-09-08
Remove the unused enums.rs

gkoz:


GuillaumeGomez:
I was sure you already removed it... ^^'

PR #146, opened 2015-09-23
Update gtk-sys and make some signatures more precise accordingly

gkoz:


gtk-sys/build.rs
```diff
@@ -35,7 +35,7 @@ fn main() {
     for cfg in &cfgs {
         println!("cargo:rustc-cfg={}", cfg);
     }
-    println!("cargo:cfg={}", cfgs.connect(" "));
+    println!("cargo:cfg={}", cfgs.join(" "));
```
GuillaumeGomez: I think my PR is still open for this XD

GuillaumeGomez:
Looks good to me. I wait for travis and stuff and then I merge. Thanks for your work !

gtk-sys/build.rs
gkoz: That PR seems to have been transformed into something entirely different :)

gkoz:
Whatever happened to appveyor. It might not have gotten the memo when the org was renamed. You may need to readd the repos there.

Issue #148, opened 2015-09-24
Setting EntryCompletion for Entry


LordSentox:
The Entry trait is currently lacking the function *set_completion()*. It is included in the *ffi crate*, but somehow doesn't seem to have made it into the *gtk crate*.

GuillaumeGomez:
Well saw, thanks ! Do you mind opening a PR to fix it ?

PS: [doc url](https://developer.gnome.org/gtk3/stable/GtkEntry.html#gtk-entry-set-completion)

LordSentox:
I have created one including the fix in #154

PR #154, opened 2015-09-25
Added completion setting/getting methods

LordSentox:
This pull request adds the missing functions as discussed in #148.
A byproduct was fixing the compilation, which failed due to the PopoverMenu being referenced in the lib-crate, without having been made available in the widget mod.
Hence the PR is split into two commits.

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -139,8 +139,6 @@ pub use self::font_chooser_widget::FontChooserWidget;
 #[cfg(target_os = "linux")]
 pub use self::socket::Socket;
 pub use self::event_box::EventBox;
-#[cfg(gtk_3_16)]
-pub use self::popover_menu::PopoverMenu;
 
 mod css_provider;
 mod style_context;
@@ -279,5 +277,3 @@ mod font_chooser_widget;
 #[cfg(target_os = "linux")]
 mod socket;
 mod event_box;
-#[cfg(gtk_3_16)]
-mod popover_menu;
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -139,6 +139,8 @@ pub use self::font_chooser_widget::FontChooserWidget;
 #[cfg(target_os = "linux")]
 pub use self::socket::Socket;
 pub use self::event_box::EventBox;
+#[cfg(gtk_3_16)]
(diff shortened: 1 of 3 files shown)


GuillaumeGomez:
Thanks ! Seems good to me. Do you confirm @gkoz ?

src/widgets/mod.rs
```diff
@@ -139,6 +139,8 @@ pub use self::font_chooser_widget::FontChooserWidget;
 #[cfg(target_os = "linux")]
 pub use self::socket::Socket;
 pub use self::event_box::EventBox;
+#[cfg(gtk_3_16)]
+pub use self::popover_menu::PopoverMenu;
```
gkoz: I believe this will conflict with the already merged #156

src/widgets/mod.rs
```diff
@@ -277,3 +279,5 @@ mod font_chooser_widget;
 #[cfg(target_os = "linux")]
 mod socket;
 mod event_box;
+#[cfg(gtk_3_16)]
+mod popover_menu;
```
gkoz: Same

src/widgets/popover_menu.rs
```diff
@@ -6,6 +6,7 @@
 
 use ffi;
 use cast::GTK_POPOVER_MENU;
+use glib::translate::ToGlibPtr;
```
gkoz: And this too.

gkoz:
The `entry` stuff looks good.

LordSentox:
Yeah, that was the same issue I experienced, which was why I originally included it in my fix.
I'm not really that experienced with this, should I open a new PR or can you merge only the first of my commits?

gkoz:
You can use `git rebase -i master` to discard the popover commit.

GuillaumeGomez:
@LordSentox: I'd do it like this:

```
git remote add upstream https://github.com/gtk-rs/gtk.git
git fetch upstream
git rebase -i upstream/master
git rebase -i HEAD~2 #remove the commit from your branch, take a look on google to know how it works
git push -f origin HEAD:master 
```

LordSentox:
Thanks, I'll try that next time. 
Anyway, it should be safe to merge now. Only the changes in the EntryTrait left.

GuillaumeGomez:
Perfect, thanks !

PR #156, opened 2015-09-25
Fix the build of popover_menu.

ltratt:
On a machine with gtk+-3.16 I previously got:

  src/lib.rs:221:5: 221:16 error: unresolved import `self::widgets::PopoverMenu`. There is no `PopoverMenu` in `widgets` [E0432]
  src/lib.rs:221     PopoverMenu
                   ^~~~~~~~~~~

because popover menu was being referenced in lib.rs, but the popover_menu
module was never imported. Once that was fixed, popover_menu didn't quite
compile. This commit fixes both issues.

### DIFF
--- a/src/widgets/mod.rs
+++ b/src/widgets/mod.rs
@@ -120,6 +120,8 @@ pub use self::recent_chooser_widget::RecentChooserWidget;
 pub use self::combo_box::ComboBox;
 #[cfg(gtk_3_12)]
 pub use self::popover::Popover;
+#[cfg(gtk_3_16)]
+pub use self::popover_menu::PopoverMenu;
 pub use self::combo_box_text::ComboBoxText;
 //pub use self::gtype::g_type;
 pub use self::text_mark::TextMark;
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Weird that travis didn't report it... Thanks a lot for this fix !

ltratt:
I think it's because travis has gtk 3.4, whereas this build problem would only manifest with gtk >= 3.16.

GuillaumeGomez:
I think, yes. @gkoz and I should start thinking about finding a way to update travis repo before building but it's not a priority for the moment. Thanks again for the fix ! :)

gkoz:
@GuillaumeGomez it looks like we need a policy to manually compile-check stuff that touches GTK 3.16 and upper before merging. Up to 3.14 should be handled by appveyor once it's working again. Opened #157.

GuillaumeGomez:
@gkoz: Perfect. Please remind me to take a look at appveyor this evening please ! :)

Issue #139, opened 2015-08-23
Access to TextIter from TextBuffer


pd40:
This is an enhancement request to read the contents of a TextBuffer, unless there is a mechanism to do this already..

TextView implements get_buffer() for access to [TextBufferTrait](https://github.com/rust-gnome/gtk/blob/master/src/traits/text_buffer.rs). TextBufferTrait only appears to implement set_text(). 

Is there a mechanism to get access to the TextIter of a TextBuffer?

insanebits:
It seems TextBuffer is missing get_text method as well.

Yesterday did some testing on your fix and it seems to be working :)

Can you create a pull request so we can start using it?

GuillaumeGomez:
Wo, this is an old one! Like, @insanebits asked, can you create a PR @pd40 please ? :)

pd40:
Thanks for trying it out @insanebits. 

Pull request [#159](https://github.com/gtk-rs/gtk/pull/159)

gkoz:
This should be fixed in #221

PR #159, opened 2015-09-30
#139 First pass at linking txt iterator to buffer

pd40:
First pass at adding text iterator to allow reading of TextView.

### DIFF
--- a/src/widgets/text_buffer.rs
+++ b/src/widgets/text_buffer.rs
@@ -6,7 +6,10 @@ use std::ptr;
 
 use ffi;
 use FFIWidget;
+use TextIter;
 
+use cast::GTK_TEXT_BUFFER;
+use glib::translate::ToGlibPtrMut;
 /// GtkTextBuffer — Stores attributed text for display in a GtkTextView
 
 struct_Widget!(TextBuffer);
@@ -21,6 +24,16 @@ impl TextBuffer {
         };
 
         check_pointer!(tmp_pointer, TextBuffer)
+    }
+
+    pub fn get_start_iter(&self) -> TextIter {
+        unsafe {
+            let mut target_iter = TextIter::uninitialized();
+            ffi::gtk_text_buffer_get_start_iter(
+                GTK_TEXT_BUFFER(self.unwrap_widget()), 
+                target_iter.to_glib_none_mut().0);
+            target_iter
+        }
     }
 }
 


GuillaumeGomez:
Perfect, thanks! I wait for checks status and I merge.

cc @gkoz

gkoz:
This is not your fault but what `TextIter::new` does makes no sense to me. I'd really like that to be fixed before anything relying on it was checked in.

Also the formatting seems non-standard, we use a 4-space indent.

gtk-sys/src/gtk_glue.c
```diff
@@ -143,6 +143,10 @@ GtkTextTagTable* cast_GtkTextTagTable(void* widget) {
     return GTK_TEXT_TAG_TABLE(widget);
 }
 
+GtkTextIter* cast_GtkTextIter(void* widget) {
+    return GTK_TEXT_ITER(widget);
```
gkoz: How this passed travis is a mystery to me. I don't think there is a `GTK_TEXT_ITER` macro.

Anyway the `text_buffer` patch doesn't use `GTK_TEXT_ITER` so you can just remove this `cast_` business.

GuillaumeGomez: Dark magic !

gkoz: Oh right, in C if you call any random `foo_bar(x, y, z)` the compiler assumes there exists a `int foo_bar(...)` and goes on happily. You'll get busted by the linker if the symbol is undefined but since this function isn't used in our examples, all is well. With rust+windows things are slightly different for now and that's why appveyor has caught this.

pd40:
Thanks for the feedback. Formatting fixed.

GuillaumeGomez:
@gkoz: Ok for you ?

gkoz:
We need to fix `TextIter` first.

gkoz:
The `TextIter` has been overhauled and you can follow [this example](https://github.com/gkoz/gtk-rs-gtk/blob/0e128976cf8aec65f849fc2349bc23385e05ad5a/src/widgets/text_view.rs#L82-L85) now.

gtk-sys/src/lib.rs
```diff
@@ -12041,6 +12041,7 @@ extern "C" {
     pub fn cast_GtkRecentInfo(widget: *mut GtkWidget) -> *mut GtkRecentInfo;
     pub fn cast_GtkEditable(widget: *mut GtkWidget) -> *mut GtkEditable;
     pub fn cast_GtkTextView(widget: *mut GtkWidget) -> *mut GtkTextView;
+    pub fn cast_GtkTextIter(widget: *mut GtkWidget) -> *mut GtkTextIter;
```
gkoz: This looks like another remnant of the `cast_*` experiment.

src/traits/text_buffer.rs
```diff
@@ -16,5 +16,4 @@ pub trait TextBufferTrait: ::WidgetTrait {
                 text.len() as i32)
         }
     }
-
```
gkoz: No point to touch this file, right?

pd40:
Thanks @gkoz for the overhaul. Will patch things up & resubmit.

gkoz:
Looks good. Can you squash the commits into one please?

GuillaumeGomez:
Perfect, thanks !

PR #160, opened 2015-09-30
added remove_column method for TreeView

insanebits:
Hey guys, it seems that `remove_column` was missing for the `TreeView`. 

It also needs `get_columns` but I'm not sure how you guys handle `GList`'s so left for someone else to implement that.


gkoz:
Thanks for the PR! The change looks good to me.

GuillaumeGomez:
Seems good to me too, I merge. Thanks for your work !

PR #165, opened 2015-10-03
Adding support for GtkStackSidebar

jmvrbanac:


src/widgets/stack_sidebar.rs
```diff
@@ -0,0 +1,44 @@
+// Copyright 2013-2015, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+//! A controller for GtkStackSidebar
```
GuillaumeGomez: "A controller for StackSideBar".

src/widgets/stack_sidebar.rs
```diff
@@ -0,0 +1,44 @@
...
+
+use ffi;
+use cast::{GTK_STACK_SIDEBAR, GTK_STACK};
+use FFIWidget;
+
+/// GtkStackSidebar —  An automatic sidebar widget
```
GuillaumeGomez: "An automatic sidebar widget" (GtkStackSidebar is unneeded here)

GuillaumeGomez:
Thanks a lot for this PR ! I just wait you to update the two comments and @gkoz's approval before merging.

jmvrbanac:
@GuillaumeGomez This is my first contribution to this project, so for forgive my ignorance on standards. I modeled the naming based on the GTK documentation for  [GtkStackSidebar](https://developer.gnome.org/gtk3/stable/GtkStackSidebar.html). From on your comments, It looks like there is a separation between the GTK names and the Rust names. Is there a documented naming format I can refer to so that I don't make this mistake again? Thanks!

GuillaumeGomez:
The goal is to provide a naming as nice as possible. So when you use a type, I think everyone prefers:

```Rust
let a = gtk::Button::new();
```

Than:

```Rust
let a = gtk::GtkButton::new();
```

So we try to follow this "rule" for documentation. Otherwise no, the code you wrote is really good, no need to worry !

About my comment about the documentation, I added the doc on most of the crates already, however, it takes a lot of time and it's really a pain in the ass (haha). So I was thinking about asking people to add doc when they add a new element. Like this, it'll reduce the size of what remains to be added.

gkoz:
About the naming: @GuillaumeGomez  we can't really do "nice" now because the generator. So I say just translate the names verbatim, stripping the `Gtk` suffix. Once the generated APIs are there someone might create a "nice" higher-level crate.

> @gkoz: BTW, should we force people to add documentation when they add a new element ?

This might be too mean because the manual documentation work should probably start from the point where we're already using the generator and it copies the docs from the girs. Then we would frown at the docs taken out of C context and rewrite some of them neatly in separate files.

GuillaumeGomez:
@gkoz: I think I'll let you write the missing documentation parts (*machiavelic laugh*) !

jmvrbanac:
@GuillaumeGomez if the goal is to just strip off the Gtk part, do you still want me to switch everything to ```StackSideBar``` instead of the spelling similar to the Gtk docs of ```StackSidebar``` ? I just want to make sure I'm making the right changes here :) .

gkoz:
Keep the spelling the same as original please, just strip off `Gtk`.

GuillaumeGomez:
@gkoz explained it better than me.

gkoz:
@GuillaumeGomez I'm going to compile-test all the PRs in a chroot with GTK 3.18.

GuillaumeGomez:
Oh nice ! Thanks a lot ! I have one on gtk and one on gdk that need it.

jmvrbanac:
Changes made. Let me know if I need to change anything else. Thanks!

GuillaumeGomez:
@jmvrbanac: Thanks !
@gkoz: Is it good or you ?

src/widgets/stack_sidebar.rs
```diff
@@ -0,0 +1,44 @@
...
+    pub fn get_stack(&self) -> Option<::Stack> {
+        let tmp_pointer = unsafe {
+            ffi::gtk_stack_sidebar_get_stack(GTK_STACK_SIDEBAR(self.pointer))
+        };
+
+        if tmp_pointer.is_null() {
+            None
+        } else {
+            Some(::FFIWidget::wrap_widget(tmp_pointer as *mut _))
+        }
```
gkoz: It seems this bit could be handled with `check_pointer!` like in `new` and cut down on the verbosity.

src/widgets/stack_sidebar.rs
```diff
@@ -0,0 +1,44 @@
...
+/// An automatic sidebar widget
+struct_Widget!(StackSidebar);
+
+impl StackSidebar {
+    pub fn new() -> Option<StackSidebar> {
+        let tmp_pointer = unsafe { ffi::gtk_stack_sidebar_new() };
+        check_pointer!(tmp_pointer, StackSidebar)
+    }
+
+    pub fn set_stack(&self, stack: ::Stack) {
```
gkoz: It'd be nicer to import `Stack` to avoid repeated leading `::`.

src/lib.rs
```diff
@@ -218,7 +218,8 @@ pub use self::widgets::{
 #[cfg(gtk_3_16)]
 /// GTK Widgets for versions since GTK 3.16
 pub use self::widgets::{
-    PopoverMenu
+    PopoverMenu,
+    StackSidebar
```
gkoz: Putting a trailing comma here is a better style.

gkoz:
:+1: on my end.

@GuillaumeGomez we might briefly consider the `new() -> Option` issue here as well, but I'm just not sure I want to push against consistency.

GuillaumeGomez:
I think it would be better to do all this change at once to remain consistent on the API.

PR #166, opened 2015-10-03
Add GLArea widget

GuillaumeGomez:


src/widgets/gl_area.rs
```diff
@@ -0,0 +1,212 @@
...
+        unsafe { ffi::gtk_gl_area_set_auto_render(GTK_GL_AREA(self.pointer), to_gboolean(auto_render)) }
+    }
+
+    /// Returns whether the area is in auto render mode or not.
+    pub fn get_auto_render(&self) -> bool {
+        unsafe { to_bool(ffi::gtk_gl_area_get_auto_render(GTK_GL_AREA(self.pointer))) }
+    }
+
+    /// Retrieves the required version of OpenGL set using GLArea::set_required_version().
+    pub fn get_required_version(&self, major: &mut i32, minor: &mut i32) {
```
gkoz: Same as https://github.com/gtk-rs/gdk/pull/75#discussion_r41091922

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,212 @@
...
+///     }
+/// }
+/// ```
+///
+/// If you need to change the options for creating the GdkGLContext you should use the "create-context" signal.
+struct_Widget!(GLArea);
+
+impl GLArea {
+    /// Creates a new GLArea widget.
+    pub fn new() -> Option<GLArea> {
```
gkoz: We will break this after the reform, removing the `Option`. Just checking if you consciously choose consistency with older widgets over forward-compatibility.

GuillaumeGomez: If this returns a null pointer, what will we do ? Panicking ? I'm against this !

GuillaumeGomez:
Updated.

src/widgets/gl_area.rs
gkoz: We will panic on any constructor returning a `NULL` because they aren't supposed to.

GuillaumeGomez: But the code should be able to continue to run anyway. Forcing this behavior isn't a good idea for me. If you don't have enough memory to allocate a new one, the previous ones still exist and work. Panicking because of this isn't a good idea !

gkoz: We've settled this long ago. For all intents and purposes `g_object_new` ["cannot ever fail"](https://developer.gnome.org/gobject/stable/howto-gobject-construction.html). It's no different than `Box::new` failing to allocate. Going for `Option` was a mistake and now everyone has useless `unwrap`s all over the place.

gkoz:
They'll probably also want a way to connect the `render` signal.

GuillaumeGomez:
Yup, didn't think of this. I'll add it tomorrow however.

src/signal.rs
```diff
@@ -1351,3 +1351,34 @@ extern "C" fn tree_view_column_trampoline(this: *mut GtkTreeViewColumn,
         f: &Box<Fn(TreeViewColumn) + 'static>) {
     f(TreeViewColumn::wrap_pointer(this))
 }
+
+#[cfg(gtk_3_16)]
```
gkoz: Due to the need for `cfg`s it's probably better to put this whole addition into a submodule like done above.

src/signal.rs
```diff
@@ -1351,3 +1351,34 @@ extern "C" fn tree_view_column_trampoline(this: *mut GtkTreeViewColumn,
...
+            connect(self.unwrap_pointer() as *mut _,"rendered",
+                transmute(gl_area_trampoline_res), into_raw(f) as *mut _)
+        }
+    }
+}
+
+#[cfg(gtk_3_16)]
+extern "C" fn gl_area_trampoline(this: *mut GtkGLArea, context: *mut gdk::GdkGLContext,
+        f: &Box<Fn(GLArea, gdk::GLContext) + 'static>) {
+    f(GLArea::wrap_widget(this), gdk::GLContext::wrap(context))
```
gkoz: This would be along the lines of
```rust
f(GLArea::wrap_widget(this), from_glib_none(context))
```

GuillaumeGomez:
@gkoz: Updated. Can you test it please ?

src/signal.rs
```diff
@@ -1351,3 +1351,39 @@ extern "C" fn tree_view_column_trampoline(this: *mut GtkTreeViewColumn,
...
+                let f: Box<Box<Fn(GLArea, i32, i32) + 'static>> = Box::new(Box::new(f));
+                connect(self.unwrap_pointer() as *mut _,"rendered",
+                    transmute(gl_area_trampoline_res), into_raw(f) as *mut _)
+            }
+        }
+    }
+
+    extern "C" fn gl_area_trampoline(this: *mut GtkGLArea, context: *mut gdk::GdkGLContext,
+            f: &Box<Fn(GLArea, gdk::GLContext) + 'static>) {
+        f(GLArea::wrap_widget(this), gdk::GLContext::wrap(context))
```
gkoz: This should be
```
f(GLArea::wrap_widget(this), from_glib_none(context))
```
similar to https://github.com/GuillaumeGomez/gtk/blob/master/src/signal.rs#L586

src/signal.rs
```diff
@@ -1351,3 +1351,39 @@ extern "C" fn tree_view_column_trampoline(this: *mut GtkTreeViewColumn,
         f: &Box<Fn(TreeViewColumn) + 'static>) {
     f(TreeViewColumn::wrap_pointer(this))
 }
+
+pub trait GLAreaSignals {
```
gkoz: missing cfg?

GuillaumeGomez: Not really. The trait will be declared but will remain unused. I can add a cfg if you want but it's not really necessary.

gkoz: Doesn't it reference the types hidden behind cfg?

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,216 @@
...
+        let tmp_pointer = unsafe { ffi::gtk_gl_area_new() };
+        check_pointer!(tmp_pointer, GLArea)
+    }
+
+    /// Retrieves the gdk::GLContext used by area.
+    pub fn get_context() -> Option<gdk::GLContext> {
+        let tmp_pointer = unsafe { ffi::gtk_gl_area_get_context() };
+        check_pointer!(tmp_pointer, gdk::GLContext)
```
gkoz: gdk is reformed, the old traits don't work on it, use `from_glib_none(ffi::gtk_gl_area_get_context())`

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,216 @@
...
+
+impl GLArea {
+    /// Creates a new GLArea widget.
+    pub fn new() -> Option<GLArea> {
+        let tmp_pointer = unsafe { ffi::gtk_gl_area_new() };
+        check_pointer!(tmp_pointer, GLArea)
+    }
+
+    /// Retrieves the gdk::GLContext used by area.
+    pub fn get_context() -> Option<gdk::GLContext> {
```
gkoz: I believe `Option` is not needed here, it's never null

gkoz: Umm, disregard this. On a closer look it can be null

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
+// Copyright 2013-2015, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+//! A widget for custom drawing with OpenGL
+
+use libc::{c_uint, c_int};
+
+use CalendarDisplayOptions;
+use cast::GTK_GL_AREA;
```
gkoz: Oops, are the cast functions missing?

src/widgets/gl_area.rs
gkoz: `use gdk` or add leading `::`

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
...
+impl GLArea {
+    /// Creates a new GLArea widget.
+    pub fn new() -> Option<GLArea> {
+        let tmp_pointer = unsafe { ffi::gtk_gl_area_new() };
+        check_pointer!(tmp_pointer, GLArea)
+    }
+
+    /// Retrieves the gdk::GLContext used by area.
+    pub fn get_context() -> Option<gdk::GLContext> {
+        unsafe { from_glib_none(ffi::gtk_gl_area_get_context()) }
```
gkoz: `use glib::translate::from_glib_none`

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
...
+    ///
+    /// If has_alpha is false there will be no alpha channel, and the buffer will fully replace
+    /// anything below the widget.
+    pub fn set_has_alpha(&self, has_alpha: bool) {
+        unsafe { ffi::gtk_gl_area_set_has_alpha(GTK_GL_AREA(self.pointer), to_gboolean(has_alpha)) }
+    }
+
+    /// Returns whether the area has an alpha component.
+    pub fn has_alpha(&self) -> bool {
+        unsafe { to_bool(ffi::gtk_gl_area_has_alpha(GTK_GL_AREA(self.pointer))) }
```
gkoz: forgot `get_` in the name

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// If has_depth_buffer is true the widget will allocate and enable a depth buffer for the target
+    /// framebuffer. Otherwise there will be none.
+    pub fn set_has_depth_buffer(&self, has_depth_buffer: bool) {
+        unsafe { ffi::gtk_gl_area_set_has_depth_buffer(GTK_GL_AREA(self.pointer), to_gboolean(has_depth_buffer)) }
```
gkoz: `get_`

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /// Returns whether the area has a depth buffer.
+    pub fn has_depth_buffer(&self) -> bool {
+        unsafe { to_bool(ffi::gtk_gl_area_has_depth_buffer(GTK_GL_AREA(self.pointer))) }
```
gkoz: `get_`

src/widgets/gl_area.rs
```diff
@@ -0,0 +1,215 @@
...
+    /// and when rendering to the window the result will be composited over whatever is below the widget.
+    ///
+    /// If has_alpha is false there will be no alpha channel, and the buffer will fully replace
+    /// anything below the widget.
+    pub fn set_has_alpha(&self, has_alpha: bool) {
+        unsafe { ffi::gtk_gl_area_set_has_alpha(GTK_GL_AREA(self.pointer), to_gboolean(has_alpha)) }
+    }
+
+    /// Returns whether the area has an alpha component.
+    pub fn has_alpha(&self) -> bool {
```
gkoz: `get_` should probably be in the method names as well, or the generator will break this

src/signal.rs
```diff
@@ -1351,3 +1357,34 @@ extern "C" fn tree_view_column_trampoline(this: *mut GtkTreeViewColumn,
...
+            connect(self.unwrap_widget() as *mut _,"rendered",
+                transmute(gl_area_trampoline_res), into_raw(f) as *mut _)
+        }
+    }
+}
+
+#[cfg(gtk_3_16)]
+extern "C" fn gl_area_trampoline(this: *mut GtkGLArea, context: *mut ::gdk_ffi::GdkGLContext,
+        f: &Box<Fn(GLArea, ::gdk::GLContext) + 'static>) {
+    f(GLArea::wrap_widget(this), ::gdk::GLContext::from_glib_none(context))
```
gkoz: `from_glib_none` is supposed to be the free function here but you'll have to import it.

GuillaumeGomez: I tried to but it seems I "cant't". Really weird error.

gkoz: [This one](https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L376) See how cairo context is handled in `connect_draw`.

gkoz:
Actually [gtk-bootstrap](https://github.com/gkoz/gtk-bootstrap) lets you avoid both the travis dance and installing a full OS in a chroot.

PR #169, opened 2015-10-06
Update the build script to play nice with GTK 3.4

gkoz:


GuillaumeGomez:
GTK 3.4 was ignored before ?

gkoz:
There is a slight inconsistency between the various build scripts that causes breakage on 3.4 (the sys crates emit gtk_3_4 and this one doesn't). This should fix that.

GuillaumeGomez:
Ok, thanks ! :)

PR #176, opened 2015-10-15
Use the `glib_wrapper!` macro and wrap TreeIter

gkoz:
Requires https://github.com/gtk-rs/glib/pull/67

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -47,6 +47,7 @@ extern crate gdk_pixbuf_sys as gdk_pixbuf_ffi;
 extern crate gtk_sys as ffi;
 extern crate cairo_sys as cairo_ffi;
 extern crate pango_sys as pango_ffi;
+#[macro_use]
 extern crate glib;
 extern crate gdk;
 extern crate cairo;
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1223,15 +1223,20 @@ mod tree_view {
 
     extern "C" fn iter_path_trampoline(this: *mut GtkTreeView, iter: *mut GtkTreeIter,
             path: *mut GtkTreePath, f: &Box<Fn(TreeView, &mut TreeIter, TreePath) + 'static>) {
-        f(FFIWidget::wrap_widget(this as *mut _), &mut TreeIter::wrap_pointer(iter),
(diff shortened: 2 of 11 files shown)


gkoz:
It seems unlikely that all the methods really need `TreeIter` to be mutable but it's the GTK developers' call.

GuillaumeGomez:
Well, I guess they have data that needs to be changed in order to have the status on read-only data... A bit weird but since I have no idea on how it works internally...

gkoz:
Or they just didn't bother putting consts in the signatures because C isn't big on immutability.

GuillaumeGomez:
Do they in the C++ binding ?

gkoz:
It seems so. Compare https://github.com/GNOME/gtk/blob/master/gtk/gtktreemodel.h#L325-L326 and
 https://github.com/GNOME/gtkmm/blob/master/gtk/src/treemodel.hg#L186

GuillaumeGomez:
Then let's do it as well. If it's autogenerated, can we set an exception for this case ?

gkoz:
Okay.

We could add a way to assert constness in the gir config. Ideally we'd probably try to push the consts upstream.

EPashkin:
Other variant do separate program that modified `c:type` in girs for listed cases.

gkoz:
The `gi-rust` project seems to be going down the path of transforming the xmls before processing them.

Adding an generalized quirks section to the config should be possible though. Unifying the existing blacklisting and non-nullability settings in the process.

gkoz:
Changed many `TreeIter` references back to immutable. Fixed more signatures I missed the last time. 

Added safety checks to `reorder`. There aren't any checks on the other side: [behold](https://github.com/GNOME/gtk/blob/master/gtk/gtktreestore.c#L2305-L2311).

gkoz:
@GuillaumeGomez you're going to like https://github.com/gtk-rs/gtk/commit/46f3cf1dcb5de5706cfcd464aa3a6c7d3b8a6e3a, this will likely solve your process viewer crashes.

GuillaumeGomez:
Oh yes ! Finally !

gkoz:
All right, it's green!

PR #177, opened 2015-10-15
Remove MenuItemTrait::activate, which duplicates the one from WidgetTrait

### LINKED ISSUE #168: MenuItem::activate conflicts with Widget::activate
Both of those traits implement activate method and trying to call it results in error. Not exactly sure how it should be fixed because they're calling different functions(`gtk_widget_activate` and `gtk_menu_item_activate`).

gkoz:
Closes #168


GuillaumeGomez:
Perfect, thanks !

PR #184, opened 2015-10-26
`=` in dependencies is too restrictive

oli-obk:
with `~` all versions up to but not including `0.4.0` are okay.


GuillaumeGomez:
@gkoz: Ping.

gkoz:
This was necessary when I tried to keep compatibility with older `rustc`. Since the checks pass, it's not any more I guess. I'd change to just `"0.3"` then.

gkoz:
Anyway `"^0.3.9"` and `"~0.3.9"` seem equivalent so I'm merging as is.

PR #185, opened 2015-11-02
Implement signal trait for GLArea

mjkoo:
Looked into what was needed to get a simple example with GtkGLArea working, had to change the signal names and expose the appropriate trait similar to other widgets. Was able to set up opengl using a dummy invisible glutin window and draw to the widget after setting this up.

Can clean up my example and submit a PR to the examples repo if you're interested. Right now it just uses gl::ClearColor to paint the whole widget red, looking into getting a triangle up akin to the example at https://www.bassi.io/articles/2015/02/17/using-opengl-with-gtk/

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -55,12 +55,6 @@ use {
     WidgetHelpType,
 };
 
-#[cfg(gtk_3_16)]
-use {
-    gdk,
-    GLArea,
-};
-
 /// Whether to propagate the signal to other handlers
 #[derive(Clone, Copy, Debug, Default, Eq, PartialEq)]
 pub struct Inhibit(pub bool);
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Thanks for the PR ! We would be totally interested in an OpenGL example.

cc @gkoz.

src/signal.rs
```diff
@@ -55,6 +55,12 @@ use {
     WidgetHelpType,
 };
 
+#[cfg(gtk_3_16)]
+use {
+    gdk,
```
gkoz: This conditional import of `gdk` could be problematic. Since we're importing a lot of stuff from `gdk` above it's okay to import `GLContext` as well.

gkoz:
You don't need to add a trait there, it would've been enough to make the methods public (apparently, I've made a poor job of reviewing this code when it was checked in).

src/signal.rs
mjkoo: Sure, actually removing the trait as mentioned in your other comment meant I could remove both of those.

mjkoo:
Thanks gkoz, I removed the trait. I'm relatively new to rust and was following the example for the calendar widget below GLArea in src/signal.rs, could I bother you for an ELI5 why it's required for one and not the other? Or is the CalendarSignals trait also superfluous?

gkoz:
You're doing very good!
A trait is needed in case the class has subclasses. Take `WidgetSignals` for example: each widget needs to have those methods. When the class is a leaf in the hierarchy we usually don't add additional traits but just implement inherent methods. I guess the existence `CalendarSignals` contradicts that.

mjkoo:
Thanks! That makes sense, I wasn't sure if the trait was useful in client code (subclassing widgets in apps?) being the reason for its inclusion, I grepped for `CalendarSignals` originally and didn't see another use from the library itself. Thanks for the explanation!

PR #186, opened 2015-11-03
[WIP] Status icons

lidavidm:
This is just to get the code for #6 visible. It's not complete.

TODO:

- [x] Make sure `gdk::Screen` type in `get_geometry` is correct
- [ ] Implement `GIcon` for `gio` (where is a good place to put this?)
- [ ] Implement `GtkMenu`
- [ ] Write an example for https://github.com/gtk-rs/examples

`Image` is missing its `GIcon`-dependent member functions as well.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -43,6 +43,7 @@ use {
     ScrollType,
     SpinButton,
     StateFlags,
+    StatusIcon,
     TextDirection,
     ToolButton,
     Tooltip,
@@ -1481,5 +1482,116 @@ mod calendar {
 
     extern "C" fn void_trampoline(this: *mut GtkCalendar, f: &Box<Fn(Calendar) + 'static>) {
         f(FFIWidget::wrap_widget(this as *mut _));
+    }
+}
+
+pub trait StatusIconSignals {
+    fn connect_activate<F: Fn(StatusIcon) + 'static>(&self, f: F) -> u64;
+    fn connect_button_press_event<F: Fn(StatusIcon, &EventButton) -> bool + 'static>(&self, f: F) -> u64;
+    fn connect_button_release_event<F: Fn(StatusIcon, &EventButton) -> bool + 'static>(&self, f: F) -> u64;
+    fn connect_popup_menu<F: Fn(StatusIcon, u32, u32) + 'static>(&self, f: F) -> u64;
+    fn connect_query_tooltip<F: Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>(&self, f: F) -> u64;
+    fn connect_scroll_event<F: Fn(StatusIcon, &EventScroll) -> bool + 'static>(&self, f: F) -> u64;
+    fn connect_size_changed<F: Fn(StatusIcon, i32) -> bool + 'static>(&self, f: F) -> u64;
+}
+
+mod status_icon {
+    use super::into_raw;
+    use StatusIcon;
+    use libc::{c_int, c_uint};
+    use std::mem::transmute;
+    use ffi::{GtkStatusIcon, GtkTooltip};
+    use gdk::{EventButton, EventScroll};
+    use glib::signal::connect;
+    use glib::translate::*;
+    use glib_ffi::gboolean;
+    use Tooltip;
+
+    impl super::StatusIconSignals for StatusIcon {
+        fn connect_activate<F: Fn(StatusIcon) + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon) + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "activate",
+                    transmute(void_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_button_press_event<F: Fn(StatusIcon, &EventButton) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, &EventButton) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "button-press-event",
+                    transmute(event_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_button_release_event<F: Fn(StatusIcon, &EventButton) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, &EventButton) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "button-release-event",
+                    transmute(event_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_popup_menu<F: Fn(StatusIcon, u32, u32) + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, u32, u32) + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "popup-menu",
+                    transmute(popup_menu_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_query_tooltip<F: Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "query-tooltip",
+                    transmute(query_tooltip_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_scroll_event<F: Fn(StatusIcon, &EventScroll) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, &EventScroll) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "scroll-event",
+                    transmute(event_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_size_changed<F: Fn(StatusIcon, i32) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, i32) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "size-changed",
+                    transmute(size_changed_trampoline), into_raw(f) as *mut _)
+            }
+        }
+    }
+
+    extern "C" fn void_trampoline(this: *mut GtkStatusIcon, f: &Box<Fn(StatusIcon) + 'static>) {
+        unsafe {
+            f(from_glib_none(this));
+        }
+    }
+
+    extern "C" fn event_trampoline(this: *mut GtkStatusIcon, event: *mut EventButton, f: &Box<Fn(StatusIcon, &EventScroll) -> bool + 'static>) -> gboolean {
+        unsafe {
+            f(from_glib_none(this), transmute(event)).to_glib()
+        }
+    }
+
+    extern "C" fn popup_menu_trampoline(this: *mut GtkStatusIcon, button: c_uint, activate_time: c_uint, f: &Box<Fn(StatusIcon, u32, u32) + 'static>) {
+        unsafe {
+            f(from_glib_none(this), button, activate_time);
+        }
+    }
+
+    extern "C" fn query_tooltip_trampoline(this: *mut GtkStatusIcon, x: c_int, y: c_int, keyboard_mode: gboolean, _tooltip: *mut GtkTooltip, f: &Box<Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>) -> gboolean {
+        unsafe {
+            f(from_glib_none(this), x, y, from_glib(keyboard_mode), Tooltip).to_glib()
+        }
+    }
+
+    extern "C" fn size_changed_trampoline(this: *mut GtkStatusIcon, size: c_int, f: &Box<Fn(StatusIcon, i32) -> bool + 'static>) -> gboolean {
+        unsafe {
+            f(from_glib_none(this), size).to_glib()
+        }
     }
 }
--- a/src/widgets/menu.rs
+++ b/src/widgets/menu.rs
@@ -1,0 +1,169 @@
+// Copyright 2013-2015, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use std::ptr;
+
+use cast::GTK_MENU;
+use gdk;
+use glib::translate::*;
+use traits::FFIWidget;
+use ffi;
+
+struct_Widget!(Menu);
+impl_TraitWidget!(Menu);
+
+impl Menu {
+    pub fn new() -> Option<Menu> {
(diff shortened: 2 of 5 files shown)


gkoz:
> Make sure `gdk::Screen` type in get_geometry is correct

This should be straightforward, `Screen` uses the same kind of wrappers.

> Implement `GIcon` for `gio` (where is a good place to put this?)

If you don't need `GIcon` for your project, feel free to just ignore it for now. There seem to be plenty other ways to set the icon.

> Implement GtkMenu

This can be done with the [generated file](https://github.com/gkoz/gtk-rs-gtk/blob/object_reform_gir/src/auto/menu.rs) as well but will require porting to the current deprecated wrappers system the `master` branch uses...

lidavidm:
> but will require porting to the current deprecated wrappers system the master branch uses...

Just to make sure, do you mean the `struct_Widget!`/`check_pointer!` stuff as used in, for example, [button.rs](https://github.com/gtk-rs/gtk/blob/master/src/widgets/button.rs)?

Also, since `GtkMenu` is a superclass, it should have a trait, if I understand correctly?

gkoz:
> Just to make sure, do you mean the `struct_Widget!`/`check_pointer!` stuff as used in, for example, button.rs?

Yep. `T: Upcast<Widget>` would translate to `T: TraitWidget`; luckily, `cast.rs` already has `GTK_MENU`.

> Also, since `GtkMenu` is a superclass, it should have a trait, if I understand correctly?

Technically true but could be slightly pointless without adding `RecentChooserMenu` so it's okay to ignore this concern for now. It doesn't pay to add bindings for every little class manually because the generator should take that over soon.

lidavidm:
Alright, wrapped part of `Menu`. Still working out the types/casts needed for `Menu.popup`, and `from_glib_none`/`check_pointer!` don't seem to help when the return type is `T: WidgetTrait` instead of a specific struct.

src/widgets/menu.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+    }
+
+    pub fn popdown(&self) {
+        unsafe {
+            ffi::gtk_menu_popdown(GTK_MENU(self.unwrap_widget()));
+        }
+    }
+
+    // pub fn popup<T: ::WidgetTrait, U: ::WidgetTrait, V>(&self, parent_menu_shell: Option<&T>, parent_menu_item: Option<&U>, func: fn(menu: &Menu, x: &mut i32, y: &mut i32, push_in: &mut bool, data: &mut V), data: V, button: u32, activate_time: u32) {
```
gkoz: We could go for the easier option here first: pass `None` as `func` and `ptr::null_mut()` as `data`. The positioning function has a reasonable default doesn't it?

src/widgets/menu.rs
```diff
@@ -0,0 +1,192 @@
...
+    }
+
+    pub fn set_title(&self, title: &str) {
+        unsafe {
+            ffi::gtk_menu_set_title(GTK_MENU(self.unwrap_widget()), title.to_glib_none().0);
+        }
+    }
+
+    // Needs the appropriate FromGlibPtrContainer impl
+    // pub fn get_for_attach_widget<T: ::WidgetTrait>(widget: &T) -> Vec<T> {
```
gkoz: This seems pretty obscure and difficult to implement right now, so drop it? ;)

lidavidm:
Ok, did the easy thing with regards to `Menu.popup`. Right now we can't connect signals to `StatusIcon` since it's not a `Widget`. I tried using a `MenuButton` to test the menus, but the menu items don't display; the status icon shows up fine though.

Sample: https://gist.github.com/lidavidm/8573ccaa759dc2bb5807

src/widgets/menu.rs
```diff
@@ -0,0 +1,180 @@
...
+                Some(ref p) => p.unwrap_widget(),
+                None => ::std::ptr::null_mut()
+            };
+            let parent_menu_item = match parent_menu_item {
+                Some(ref p) => p.unwrap_widget(),
+                None => ::std::ptr::null_mut()
+            };
+            ffi::gtk_menu_popup(
+                GTK_MENU(self.unwrap_widget()),
+                parent_menu_shell,
```
gkoz: Matches are too cumbersome for this, after importing `std::ptr` handling the optional comes down to:
```
parent_menu_shell.map(|p| p.unwrap_widget()).unwrap_or(ptr::null_mut())
```
which will be much easier to convert to `.to_glib_none().0` when the time comes.

src/widgets/status_icon.rs
```diff
@@ -0,0 +1,242 @@
...
+    pub fn new_from_stock(stock_id: &str) -> StatusIcon {
+        unsafe {
+            from_glib_full(ffi::gtk_status_icon_new_from_stock(stock_id.to_glib_none().0))
+        }
+    }
+
+    pub fn get_geometry(&self, screen: &gdk::Screen, area: &mut RectangleInt) -> Option<Orientation> {
+        unsafe {
+            let mut res = mem::uninitialized();
+            if from_glib(ffi::gtk_status_icon_get_geometry(
```
gkoz: We have three outparams here so they all have to be in the return value: 
```rust
get_geometry(&self) -> Option<(gdk::Screen, RectangleInt, Orientation)> {
...
    let mut screen = ptr::null_mut();
...
        &mut screen,
...
    Some((from_glib_none(screen), area, orientation))
...
```

src/widgets/status_icon.rs
```diff
@@ -0,0 +1,242 @@
...
+        }
+    }
+
+    pub fn set_visible(&self, visible: bool) {
+        unsafe {
+            ffi::gtk_status_icon_set_visible(self.to_glib_none().0, visible.to_glib());
+        }
+    }
+
+    pub fn position_menu(menu: &::Menu, x: &mut i32, y: &mut i32, user_data: &StatusIcon) -> bool {
```
gkoz: I'd comment this one out because it's pointless with our limited version of `popup`.

gkoz:
> Right now we can't connect signals to `StatusIcon` since it's not a `Widget`

Yep, it has its own set of signals, most of them seemingly can be copied from `WidgetSignals` (you don't have to add a trait, it's okay to implement them on the struct itself like it's done at the bottom of `signal.rs` for other structs).

gkoz:
> I tried using a MenuButton to test the menus, but the menu items don't display; the status icon shows up fine though.

Apparently, despite `menu_button.set_popup(&menu)` the menu isn't considered the window's child and isn't touched by `window.show_all()`. `menu.show_all()` does the trick.

lidavidm:
Alright, signals implemented and fixed `get_geometry`. At this point right-clicking and displaying a menu works! https://gist.github.com/lidavidm/8573ccaa759dc2bb5807

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+    impl super::StatusIconSignals for StatusIcon {
+        fn connect_activate<F: Fn(StatusIcon) + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon) + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "activate",
+                    transmute(activate_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_button_press_event<F: Fn(StatusIcon, &EventAny) -> bool + 'static>(&self, f: F) -> u64 {
```
gkoz: It seems the event type should be `EventButton` similarly to `WidgetSignals`.

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, &EventAny) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "button-press-event",
+                    transmute(event_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_button_release_event<F: Fn(StatusIcon, &EventAny) -> bool + 'static>(&self, f: F) -> u64 {
```
gkoz: Ditto.

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+
+        fn connect_query_tooltip<F: Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "query-tooltip",
+                    transmute(query_tooltip_trampoline), into_raw(f) as *mut _)
+            }
+        }
+
+        fn connect_scroll_event<F: Fn(StatusIcon, &EventAny) -> bool + 'static>(&self, f: F) -> u64 {
```
gkoz: `EventScroll`?

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+        fn connect_size_changed<F: Fn(StatusIcon, i32) -> bool + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(StatusIcon, i32) -> bool + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0 as *mut _, "size-changed",
+                    transmute(size_changed_trampoline), into_raw(f) as *mut _)
+            }
+        }
+    }
+
+    extern "C" fn activate_trampoline(this: *mut GtkStatusIcon, f: &Box<Fn(StatusIcon) + 'static>) {
```
gkoz: This should be named `void_trampoline` for consistency with sibling modules.

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+    }
+
+    extern "C" fn popup_menu_trampoline(this: *mut GtkStatusIcon, button: u32, activate_time: u32, f: &Box<Fn(StatusIcon, u32, u32) + 'static>) {
+        unsafe {
+            let icon: StatusIcon = from_glib_none(this);
+            f(icon, button, activate_time);
+        }
+    }
+
+    extern "C" fn query_tooltip_trampoline(this: *mut GtkStatusIcon, x: i32, y: i32, keyboard_mode: bool, _tooltip: *mut GtkTooltip, f: &Box<Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>) {
```
gkoz: This signal returns a `gboolean`.

src/signal.rs
```diff
@@ -1483,3 +1484,113 @@ mod calendar {
...
+        unsafe {
+            f(from_glib_full(this), x, y, keyboard_mode, Tooltip).to_glib();
+        }
+    }
+
+    extern "C" fn size_changed_trampoline(this: *mut GtkStatusIcon, size: i32, f: &Box<Fn(StatusIcon, i32) -> bool + 'static>) {
```
gkoz: This signal returns a `gboolean`.

src/widgets/menu.rs
```diff
@@ -0,0 +1,176 @@
...
+        let tmp_pointer = unsafe { ffi::gtk_menu_new() };
+        check_pointer!(tmp_pointer, Menu)
+    }
+
+    //#[cfg(gtk_3_4)]
+    //pub fn new_from_model<T: Upcast<gio::MenuModel>>(model: &T) -> Menu {
+    //    unsafe { TODO: call ffi:gtk_menu_new_from_model() }
+    //}
+
+    // Should be implemented in a MenuShell trait eventually
```
gkoz: Wait, doesn't `MenuShellTrait`, which is `impl`'d below already have this method?

gkoz:
About `position_menu`, If the default `Menu::popup` behavior proves inadequate we might add a `StatusIcon::popup(&self, menu: &Menu, button: u32, activate_time: u32)` method, that would take care of passing the correct `func`.

src/signal.rs
gkoz: Oh and I'd prefer we stayed literal and kept `c_int` instead of `i32` (and `c_uint` instead of `u32` where appropriate). Speaking of c-types, `bool` is very different from `gboolean`.

GuillaumeGomez: It's weird that gtk didn't use the official C `bool` type yet.

lidavidm:
Changes:

- Removed `Menu.append` because `MenuShellTrait` implements it
- Return `gboolean` where appropriate
- `i32 -> c_int` and `u32 -> c_uint` in signal handlers
- Use appropriate `Event` struct
- Renamed `activate_trampoline` to `void_trampoline`

src/signal.rs
```diff
@@ -1483,3 +1484,115 @@ mod calendar {
...
+    }
+
+    extern "C" fn popup_menu_trampoline(this: *mut GtkStatusIcon, button: c_uint, activate_time: c_uint, f: &Box<Fn(StatusIcon, u32, u32) + 'static>) {
+        unsafe {
+            let icon: StatusIcon = from_glib_none(this);
+            f(icon, button, activate_time);
+        }
+    }
+
+    extern "C" fn query_tooltip_trampoline(this: *mut GtkStatusIcon, x: c_int, y: c_int, keyboard_mode: bool, _tooltip: *mut GtkTooltip, f: &Box<Fn(StatusIcon, i32, i32, bool, Tooltip) -> bool + 'static>) -> gboolean {
```
gkoz: `keyboard_mode` here should be `gboolean`. C.f. https://developer.gnome.org/gtk3/stable/GtkStatusIcon.html#GtkStatusIcon-query-tooltip

src/widgets/status_icon.rs
```diff
@@ -0,0 +1,240 @@
...
+        }
+    }
+
+    pub fn new_from_file(filename: &str) -> StatusIcon {
+        unsafe {
+            from_glib_full(ffi::gtk_status_icon_new_from_file(filename.to_glib_none().0))
+        }
+    }
+
+    pub fn new_from_gicon(icon: *mut gio_ffi::GIcon) -> StatusIcon {
```
gkoz: A function that takes a raw pointer would probably have to be unsafe.

lidavidm: Ah, I think it's best to remove this for now, since Icon isn't implemented.

lidavidm:
Changes:

- Use gboolean in signal handler
- Remove `new_from_gicon`

src/signal.rs
```diff
@@ -1483,3 +1484,115 @@ mod calendar {
...
+
+    extern "C" fn void_trampoline(this: *mut GtkStatusIcon, f: &Box<Fn(StatusIcon) + 'static>) {
+        unsafe {
+            f(from_glib_none(this));
+        }
+    }
+
+    extern "C" fn event_trampoline(this: *mut GtkStatusIcon, event: *mut EventButton, f: &Box<Fn(StatusIcon, &EventScroll) -> bool + 'static>) -> gboolean {
+        unsafe {
+            f(from_glib_full(this), transmute(event)).to_glib()
```
gkoz: Oh, one other thing I've overlooked before is that all trampolines should use `from_glib_none(this)` because the library doesn't expect us to decrease the reference count.

src/signal.rs
```diff
@@ -1483,3 +1484,115 @@ mod calendar {
...
+        }
+    }
+
+    extern "C" fn popup_menu_trampoline(this: *mut GtkStatusIcon, button: c_uint, activate_time: c_uint, f: &Box<Fn(StatusIcon, u32, u32) + 'static>) {
+        unsafe {
+            let icon: StatusIcon = from_glib_none(this);
```
gkoz: Binding `icon` here seems inconsistent with the other trampolines.

gkoz:
Aside from the last two notes, this seems good to go, we can take care of `GIcon` and the outstanding methods later. Would you consider squashing some of the fixes so we'd have about 2-4 commits?

lidavidm:
Alright, turned it into two commits, one implementing `StatusIcon`, the other implementing `Menu`.

gkoz:
Thanks.
:+1: 
@GuillaumeGomez

GuillaumeGomez:
@lidavidm: Thanks for your work. I merge !
@gkoz: Thanks for taking care of this one :)

lidavidm:
@gkoz @GuillaumeGomez thank you both for all your help & comments!

Issue #188, opened 2015-11-06
Enforce the requirement to call gtk::init


gkoz:
It's unsafe to let the user do anything before calling `gtk::init`. See also #3.

I don't think we're going to make sweeping changes like in #34 just yet so for the time being I'm thinking of adding thread local booleans to `gdk` and `gtk`
```rust
thread_local!(static INITIALIZED: Cell<bool> = Cell::new(false));
```
and asserts to all constructors and free functions
```rust
assert!(INITIALIZED.with(|v| v.get()), "Not the main thread or failed to call `gtk::init`");
```

(Perhaps with another static to distinguish the two cases)

Unresolved questions: What about `idle_add`? Should this reach into `glib` as well?

GuillaumeGomez:
It's hard to say. If we start to move functions from one lib to another, it can get confusing for users. Otherwise I agree with your other changes.

gkoz:
Fixed in #189

PR #189, opened 2015-11-17
Add safety checks

### LINKED ISSUE #187: Catch panics in callbacks
Panicing into C code is undefined behavior but gtk-rs doesn't currently handle panics in callbacks. Until `catch_panic` (or `recover` or whatever it ends up being called) is stabalized, gtk-rs should probably just abort on panic (using a destructor bomb).

gkoz:
* Assert that GTK is initialized in constructors and free functions.
 * Prevent unwinding across FFI.
 * Make `main_quit` safe.

Fix #187 and #188

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ addons:
     - libgtk-3-dev
 script:
   - rustc --version
+  - ./check_init_asserts
   - mkdir .cargo
   - echo 'paths = ["."]' > .cargo/config
   - git clone -q --depth 50 -b pending https://github.com/rust-gnome/examples _examples
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: rust
 rust:
-  - nightly
+  - beta
   - stable
 env:
   - GTK=3.4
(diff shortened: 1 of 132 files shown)


GuillaumeGomez:
And here comes the monstruous thing... I forgot to ask : does your script takes in account the fact the macro is already called or does it just add it without checking ?

gkoz:
The script is a fancy specialized version of grep. It will fail the test if a non-method is missing the assert. Macro invocations were added semi-automatically by a variation of that regex.

gkoz:
[This](https://gist.github.com/gkoz/f2d4a2119ee93ec7dfd0) is the script, that helps me do these kinds of sweeping 'refactorings'. A long time ago when starting to tinker with rgtk I was intending to make a graphical tool in Rust for this but this script does the job mostly.

GuillaumeGomez:
We should think about adding a repo for all these scripts, don't you think ?

gkoz:
Dunno. The replacer script is pretty generic and spartan. Anything else?

GuillaumeGomez:
The `rustdoc-stripper` when the regenerate part will be done ?

gkoz:
Yeah perhaps.
Oops, I've just tried running the examples for the first time and looks like I'll be fixing up the PRs :fearful:

gkoz:
Added callback guard. This completes the safety fixes I had in mind for now.

PR #190, opened 2015-11-20
get_child_at function for Grid

LordSentox:
The function was missing, so I added it.

### DIFF
--- a/src/traits/container.rs
+++ b/src/traits/container.rs
@@ -44,4 +44,5 @@ pub trait ContainerTrait: ::WidgetTrait {
             ffi::gtk_container_set_border_width(GTK_CONTAINER(self.unwrap_widget()), border_width as c_uint);
         }
     }
+
 }
--- a/src/widgets/grid.rs
+++ b/src/widgets/grid.rs
@@ -54,6 +54,18 @@ impl Grid {
         }
     }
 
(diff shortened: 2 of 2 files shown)


gkoz:
Thank you for contributing!

Unfortunately the wrappers used in this branch can't handle checking the actual type of the widget so either the function has to be marked `unsafe` or return `Option<Widget>`. The branch, that would support downcasting properly, is under development.

gkoz:
I think I'll go and mark the existing similar occurrences `unsafe`, making this one `unsafe` would be consistent with that.

LordSentox:
I see. I was confused for a second, because I had modeled this function after the get_child function of bin, which was not marked unsafe.
Should I mark the function unsafe then, or would you prefer it to be in your commit, when you change the other functions alongside with it?

gkoz:
I'd prefer you changed it but let's see what @GuillaumeGomez of the overall plan.

GuillaumeGomez:
Hi @LordSentox, thanks for PR ! Like @gkoz said, I'd prefer you to set the function as `unsafe`. It's a temporary state, but I'd prefer it this way now.

gkoz:
The better way to remove that commit would be by doing an interactive rebase.

Don't bother with handling lists and stuff unless you require a particular function. Most of it has been solved, just has to make its way into this branch.

PR #206, opened 2015-12-07
Add gtk::Window::new() example

joshcurtis:
I found it frustrating to not have any examples to go along with the documentation so I added a link to the GTK site and added an example of how to create a window. If this is useful, I can add examples and links for some other widgets.

### DIFF
--- a/src/widgets/window.rs
+++ b/src/widgets/window.rs
@@ -16,6 +16,20 @@ use WindowType;
 */
 struct_Widget!(Window);
 
+/// Go to the  GTK website for complete
+/// [documentation](https://developer.gnome.org/gtk3/stable/GtkWindow.html).
+///
+/// # Examples
+///
+/// ```
+/// use gtk::traits::*;
+///
+/// gtk::init();
+/// let window = gtk::Window::new(gtk::WindowType::Toplevel).unwrap();
+/// window.set_window_position(gtk::WindowPosition::Center);
+/// window.show_all();
+/// ```
+///
 impl Window {
     pub fn new(window_type: WindowType) -> Option<Window> {
         assert_initialized_main_thread!();


src/widgets/window.rs
```diff
@@ -16,6 +16,20 @@ use WindowType;
...
+///
+/// ```
+/// use gtk::traits::*;
+///
+/// gtk::init();
+/// let window = gtk::Window::new(gtk::WindowType::Toplevel).unwrap();
+/// window.set_window_position(gtk::WindowPosition::Center);
+/// window.show_all();
+/// ```
+///
```
GuillaumeGomez: This line is useless.

GuillaumeGomez:
The documentation url is in the [README file](https://github.com/gtk-rs/gtk/blob/master/README.md#documentation). So I don't really see why we should merge your PR. Soory about that.

cc @gkoz

joshcurtis:
The README file doesn't have a link to GTK's about the window class. The page for the (window)[http://gtk-rs.org/docs/gtk/struct.Window.html] documentation doesn't have any examples on how to use the rust bindings. 

I think the link to the GTK documentation is useful for people who aren't familiar with the GtkWindow, and the example is useful to show people one way to use the library.

GuillaumeGomez:
The README file gives a link to the docs. After, it's up to you to look for what you want. We can't give a link for every structs and traits available in gtk-rs.

GuillaumeGomez:
Hum... It actually might be a good idea to add more examples directly in the doc (by rereading it, it seems to be a big lack on gtk crate).

However, @gkoz is currently refactoring an enormous part of the gtk crate, and so, I think we should wait for him to finish before adding more doc (the unwrap will disappear for example, and I don't want to update all code examples once again ^^').

gkoz:
I'd say the more immediate issue is that we're moving the documentation itself around (see https://github.com/gtk-rs/rustdoc-comments) and it's not entirely clear what will go where. I believe the starting point would be to use the generator to fill those external documentation files with upstream documentation, for instance `Window` [would have this](https://developer.gnome.org/gtk3/stable/GtkWindow.html#GtkWindow.description):

> A `Window` is a toplevel window which can contain other widgets. Windows normally have decorations that are under the control of the windowing system and allow the user to manipulate the window (resize it, move it, close it,...).
...

Any customizations and examples would then be done on top of that.

So, @joshcurtis your effort is appreciated but the impact will be bigger after we sort out the documentation repo story.

joshcurtis:
Ok, I'll wait until the documentation is sorted out before adding anything.

GuillaumeGomez:
Since this PR is super old, I'll close it for now. Feel free to reopen an up-to-date one. :)

PR #208, opened 2015-12-12
Added "changed" signal to TextBuffer and Entry

aethertap:
Okay, one commit, one file changed, hopefully that works better. I added support for the "changed" signal to the EntrySignals trait, and added a TextBufferSignals trait with the "changed" signal.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -40,6 +40,7 @@ use {
     Dialog,
     DirectionType,
     Entry,
+    TextBuffer,
     MovementStep,
     Range,
     ScrollType,
@@ -822,6 +823,7 @@ pub trait EntrySignals {
     fn connect_move_cursor<F: Fn(Entry, MovementStep, i32, bool) + 'static>(&self, f: F) -> u64;
     fn connect_insert_at_cursor<F: Fn(Entry, &str) + 'static>(&self, f: F) -> u64;
     fn connect_preedit_changed<F: Fn(Entry, &str) + 'static>(&self, f: F) -> u64;
+    fn connect_changed<F: Fn(Entry) + 'static>(&self, f: F) -> u64;
 }
 
 mod entry {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Very nice, thanks ! (You could have made a `git push -f` on your `dev` branch but whatever haha).

Ok for you @gkoz ?

gkoz:
There doesn't seem to be a `changed` in the [`GtkEntry` signals list](https://developer.gnome.org/gtk3/stable/GtkEntry.html#GtkEntry.signals).

src/signal.rs
```diff
@@ -811,6 +812,7 @@ mod widget {
 
 }
 
+
```
gkoz: It'd be nice to avoid spurious additions of blank lines like this.

src/signal.rs
```diff
@@ -915,6 +918,13 @@ mod entry {
                     transmute(string_trampoline), Box::into_raw(f) as *mut _)
             }
         }
+        fn connect_changed<F: Fn(Entry) + 'static>(&self, f: F) -> u64 {
```
gkoz: But leaving a blank line between functions would be consistent with the rest of the module.

csoriano89:
It's either notify:text in GtkEntry or search-changed in GtkSearchEntry

aethertap:
My apologies for whitespace inconsistency. As far as getting notified when the content changes, the signal in GTK is part of the [Editable](https://developer.gnome.org/gtk3/unstable/GtkEditable.html) interface that's implemented by Entry (which I didn't actually realize until reading your comment that it's not in Entry). Given that, this PR probably needs to be reworked and I'm sorry for wasting your time.

Since the changed signal comes from Editable, I'm thinking what I should have done was to add an EditableSignals trait to signal.rs. I don't see that in there, but before I go down that path is this something you'd be interested in, or am I still off target?

gkoz:
Oh, signals on an interface could be a problem at the moment. I guess it's alright to add `connect_changed` to all implementors of `Editable` we have then. When the object reform, which has proper up/downcasting support, lands, this situation will be improved.

aethertap:
I added an EntryCompletionSignals trait to support the "match-selected" and "action-activated" signals for EntryCompletion, but I have it in a different branch ([entrycompletion](https://github.com/aethertap/gtk/commit/d8fac83f6107f3ad303bb901b5333577e5e8788d)) for now. Would it be better to fold in with this PR or keep it separate?

gkoz:
Good job, thanks!

I'll merge this and the other branch can be its own PR. Please watch out for whitespace style consistency (spaces after punctuation and around operators, e.g. `foo: bar`, `foo, bar` and `bar + baz`) there.

PR #211, opened 2015-12-15
Update version number in README's Cargo instructions

pshendry:
A minor point obviously, but I followed the README blindly and got bit because 0.0.4 won't build some reason. Not sure why that was, but in any case 0.0.6 built just fine.

Specifying 0.0 still allows for 0.1 to break APIs, but grabs the latest 0.0.* patch version. Seems like the desired behaviour, and it avoids the issue of forgetting to update this bit again when future versions come out.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ Examples of gtk applications can be found in the [examples repository](https://g
 To include gtk as a cargo dependency you have to add it to your Cargo.toml:
 ```Toml
 [dependencies]
-gtk = "0.0.4"
+gtk = "0.0"
 ```
 
 ## Use __gtk__


GuillaumeGomez:
Hi, thanks for this! However, I'm a bit reluctant to this. The change needs to be done, but it's *generally* preferred to give a precise crate version.

What do you think of this @gkoz?

pshendry:
Happy to change it to 0.0.6 if you prefer, perhaps also with a comment above the version line in Cargo.toml reminding to update the README when the version is bumped? Having to remember to review the docs/README on a patch update seems a little arduous.

GuillaumeGomez:
That's why I don't really know what solution to choose. I think for now I'll take your initial proposition and if needed, we'll update.

So I merge, thanks for your contribution !

pshendry:
No trouble at all. Thanks for all the hard work on this, having Gtk bindings is great :)

GuillaumeGomez:
Well, if it's useful to someone, then it means it worthes it. ^^

gkoz:
Note that in the `0.0.x` territory, `x` is considered a major version for the purposes of semver, so perhaps we should be bumping the version appropriately, as suggested initially.

pshendry:
That README version bump didn't happen though, twice in a row. Which is understandable, it's easy to miss, but I was aiming for a solution which reduces the chance of that happening next time in addition to fixing it right now.

My gut feel is that the typical user of a 0.0.* library is interested in receiving updates, breaking or not, which makes it a reasonable default. Only by 0.1 would I be expecting any API stability as a user.

Anyway, not looking to bikeshed over it, so if you prefer something else I'm happy to submit it!

Paul

gkoz:
Since we have the gtk-rs.org site now (and the version is updated properly there), I wouldn't mind culling the readme, moving any user-centric information to the site and linking there. The readme would then be aimed at contributors and those who want to follow the master branch (git urls). @GuillaumeGomez how about that?

GuillaumeGomez:
@gkoz: This is an excellent idea !
@pshendry: Do you want to do it ?

PR #217, opened 2015-12-31
Add MenuBar

trolleyman:


GuillaumeGomez:
Thanks for your PR! I took a quick look and it seems correct but we'll also need @gkoz's confirmation.

trolleyman:
Oh there's just one thing - I can't find a way to have a 'proper' menu bar. This one is not native, but I found [this](http://gtk-rs.org/docs/gtk_sys/fn.gtk_application_set_menubar.html) function in the docs, and was wondering if this is the way to create a proper menu bar?
For reference, MenuBar looks like this:

image

And native menus look like this:

image

trolleyman:
Also Happy New Year!

gkoz:
Okay doing `GMenuModel` definitely needs the object reform to be in place. I'm wrapping it up shortly and this issue can then progress.

GuillaumeGomez:
@trolleyman: Thanks, you too! And just like @gkoz said, you'll have to be little bit patient, sorry about that.

trolleyman:
Ok cool, thanks for doing this. Is [this](https://github.com/gkoz/gtk-rs-gtk/tree/object_reform) the object reform repo?

gkoz:
The latest now is https://github.com/gkoz/gtk-rs-gtk/tree/object_reform_gir, it takes advantage of https://github.com/gkoz/gir, and there are some updates I haven't pushed yet

gkoz:
I've looked into this and for the `GMenu` and friends bindings to be useful there need to be `GVariant` bindings. It's funny I've just finished dealing with `GValue` and here's another "variant datatype".

gkoz:
But anyway adding `MenuBar` would go through `Gir.toml` now. See #232 for guidance.

GuillaumeGomez:
I think we can close this issue now. Thanks anyway @trolleyman!

wrnice:
Any news about this ? I am missing MenuBar a lot !

GuillaumeGomez:
I think we could add them now since GValue binding has been done. Or something else is still missing @gkoz?

PR #221, opened 2016-01-11
Object reform

gkoz:
The (project of the year 2015) [object reform](https://github.com/gtk-rs/gtk/pull/59) is getting ready for merging.

While there remain some handmade modules, this reform is powered by [code generation](https://github.com/gkoz/gir) and I'm grateful to @EPashkin for working with me on making this automated refactoring possible.

I've tried to minimize breaking changes but there still is a lot, some caused by human errors in the old code, some by making the signatures closer to their stated or apparent contracts, some by attempting to make APIs more Rust-y. A small number of functions will disappear for a while.

I'd like to know about any breaking changes that shouldn't be there.

Note that at this point we're not covering the whole GTK API, the first goal was a subset of it intersecting with our existing codebase. After this is in place, more generated bindings will appear.

Requires https://github.com/gtk-rs/glib/pull/86. Examples fixes at https://github.com/gtk-rs/examples/pull/56.


gsingh93:
Wow, good job guys. Looking forward to testing this out.

GuillaumeGomez:
It'll take me a while to review (it'll land tomorrow or after tomorrow if everything's fine for me). Congrats to you @gkoz and @EPashkin, this is really impressive !

gkoz:
Oh, the new objects framework documentation is not too good right now, I'll add an overview tomorrow.

src/auto/about_dialog.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    pub fn set_website(&self, website: Option<&str>) {
+        unsafe {
+            ffi::gtk_about_dialog_set_website(self.to_glib_none().0, website.to_glib_none().0);
+        }
+    }
+
+    pub fn set_website_label(&self, website_label: &str) {
```
GuillaumeGomez: Why this `&str` isn't wrapped inside an `Option`?

src/auto/about_dialog.rs
```diff
@@ -0,0 +1,219 @@
...
+            ffi::gtk_about_dialog_set_website_label(self.to_glib_none().0, website_label.to_glib_none().0);
+        }
+    }
+
+    pub fn set_wrap_license(&self, wrap_license: bool) {
+        unsafe {
+            ffi::gtk_about_dialog_set_wrap_license(self.to_glib_none().0, wrap_license.to_glib());
+        }
+    }
+
```
GuillaumeGomez: There is an extra empty line in every struct by the way (keep me this error, I'll fix it on the generator directly! :D ).

src/auto/cell_area.rs
```diff
@@ -0,0 +1,341 @@
...
+        unsafe {
+            FromGlibPtrContainer::from_glib_none(ffi::gtk_cell_area_get_focus_siblings(self.to_glib_none().0, renderer.to_glib_none().0))
+        }
+    }
+
+    fn get_preferred_height<T: Upcast<Widget>>(&self, context: &CellAreaContext, widget: &T) -> (i32, i32) {
+        unsafe {
+            let mut minimum_height = mem::uninitialized();
+            let mut natural_height = mem::uninitialized();
+            ffi::gtk_cell_area_get_preferred_height(self.to_glib_none().0, context.to_glib_none().0, widget.to_glib_none().0, &mut minimum_height, &mut natural_height);
```
GuillaumeGomez: Same for variables. I'd like to insert an empty line between variable(s) declaration and code. I'll see on the generator directly.

GuillaumeGomez:
Very impressive work @gkoz and @EPashkin ! Thanks a lot to both of you. Except the two issues there are on the generator, the code is great.

@gkoz: I'll wait your answer about the unwrapped `&str` before merging.

src/auto/about_dialog.rs
EPashkin: By https://developer.gnome.org/gtk2/stable/GtkAboutDialog.html#gtk-about-dialog-set-website-label in not allow Null.
But code same as for `set_website` so it possible mistake in gir.
Can be overridden in Gir.toml (see `Gtk.AppChooserDialog` as example)

GuillaumeGomez: We're supposed to wrap gtk3 ;) (I'm referencing to your url which should be https://developer.gnome.org/gtk3/stable/GtkAboutDialog.html#gtk-about-dialog-set-website-label). It doesn't say anything about the fact that it cannot be `NULL` however. I don't really mind but it seems a bit strange.

EPashkin: There many same error in .gir files so we add override in project Gir.toml if it really needed.
By way, I say wrong example: wanted to say `set_invisible_char` in `Gtk.Entry`.

GuillaumeGomez: Oh I see. Have you fix these errors in official GIR repository?

EPashkin: It need to be fixed in current projects `Gir.toml`. Minor problem that after change `Gir.toml` full regen needed so IMHO better don't fix single error and collect some.

src/auto/cell_area.rs
gkoz: I don't see why you'd want that. I've never seen such a style guideline in any Rust project.

This isn't ANSI C where the declarations are grouped at the top. There doesn't seem to be a logical reason to separate variable initialization and use here (I can see blank lines between logically separate blocks of code). The ’let` keyword should be prominent enough.

GuillaumeGomez: It's just a personal preference. However, if no one wants it, they I'll just live with it. Like I said, just a personal preference.

src/auto/about_dialog.rs
gkoz: This does seem like an upstream mistake because they [handle the `NULL` label case](https://github.com/GNOME/gtk/blob/master/gtk/gtkaboutdialog.c#L1048-L1051).
I'll add an override.

GuillaumeGomez: Do you want me to wait for you update?

src/auto/cell_area.rs
gkoz: Personal preference is understandable but breaking down what seems like a logically singular operation of calling a function impacts readability (like e.g. an, out of place comma) so could you come up with some at least semi-rational justification... ;)

gkoz:
Updated `set_website_label`.

GuillaumeGomez:
The "error" has been fixed. Everything's good. Thanks again to both of you @gkoz and @EPashkin!

GuillaumeGomez:
We'll now need to update travis and appveyor scripts.

gkoz:
Anything wrong with the scripts? We can rerun the builds on examples now.

GuillaumeGomez:
I did. Take a look.

PR #224, opened 2016-01-13
Update of the generator which remove unneeded extra backlines

GuillaumeGomez:
@gkoz: I have shared feelings about adding empty line between functions declaration: it makes reading easier, but files longer. What do you think?


gkoz:
I guess considering we'll be putting the docs there, blank lines between functions make sense. But this should probably be better reflected in the notes then. Also make sure you're using `gir` from `master` when making the last version of the PR (i.e. regenerate after the gir PR is merged).

GuillaumeGomez:
Ok, I update the commit message on `gir`.

gkoz:
> Also make sure you're using `gir` from `master` when making the last version of the PR (i.e. regenerate after the gir PR is merged)

This will pick up other changes to gir though making you commit message here incomplete yet again :(

GuillaumeGomez:
That's not a real problem, I'll just regenerate the sources and change the commit message here as well.

GuillaumeGomez:
@gkoz: Updated with the gir master branch.

src/auto/actionable.rs
```diff
@@ -17,19 +17,25 @@ glib_wrapper! {
...
+
     //#[cfg(gtk_3_4)]
     //fn set_action_target_value(&self, target_value: /*Ignored*/&glib::Variant);
+
     #[cfg(gtk_3_4)]
     fn set_detailed_action_name(&self, detailed_action_name: &str);
 }
 
 impl<O: Upcast<Actionable>> ActionableExt for O {
+
```
gkoz: Bummer. We've missed this case.

src/auto/bin.rs
```diff
@@ -21,10 +21,10 @@ pub trait BinExt {
 }
 
 impl<O: Upcast<Bin>> BinExt for O {
+
```
GuillaumeGomez: `impl` case is broken. I fix it.

GuillaumeGomez:
Updated.

gkoz:
Yay! Thanks.

PR #226, opened 2016-01-13
Signals evolution

gkoz:
* Change signal handler signatures to take all wrapper types by reference
* Change the first argument of all signal handlers to `&Self`


GuillaumeGomez:
Weird PR. Github doesn't appreciate your commits it seems.

gkoz:
I may have rebased incorrectly

gkoz:
Looks like the module is too big for github to show the diffs :disappointed:

PR #230, opened 2016-01-17
Add gir and gir-files submodules and implement `make gir`

gkoz:


gir
```diff
@@ -0,0 +1 @@
+Subproject commit 463de47aa99618470e1d6343da34fa2cc211b5a0
```
GuillaumeGomez: What is this?

gir-files
```diff
@@ -0,0 +1 @@
+Subproject commit 11e0e6d1ffcca587f42d91c66064c0c9d21497da
```
GuillaumeGomez: And this?

gir
EPashkin: Its hash of current commit in submodule.

EPashkin:
IMHO adding submodules not very good idea now, it created too strong link.

gkoz:
@EPashkin I must be missing the downsides. Using the crate does not involve submodules, they can be ignored. But they can avoid confusion if some incompatible changes to either gir or gir-files while someone learns to run the generator in this repo. Adding stuff to `Gir.toml` is more accessible this way.

EPashkin:
Problem for me was multiply copies of `gir` and `gir-files` (already has 2 of it, symlinked though).
For crate user sub-modules unneeded and unnoticed, so its presence does not matter.
For "novice" `gtk` crate-developers submodules is good help as mentioned by @gkoz and this reason outweigh others.
Not sure if it stay same with predicting adding other gir-generated modules to `gtk-rs` :sweat_smile:

Issue #213, opened 2015-12-23
Images in TreeView?


panpiotrs:
Is it possible to create something like this right now or is code necessary for that not yet implemented? I'm in my third day of learning rust, I know how to work with ordinary text columns, but making list with images is beyond me, can you help a little bit?
![Screenshot](http://i.stack.imgur.com/c2g54.png)

GuillaumeGomez:
Yes, I already did something similar. However I don't really have time (holidaaaaays !). You'll certainly have to wait a few days before having an answer (at least from me), sorry.

panpiotrs:
No problem, I will wait and learn rust in other ways, thank you and happy holidays! :)

GuillaumeGomez:
Ok, I added it in the examples repo [here](https://github.com/gtk-rs/examples/blob/pending/src/treeview.rs#L85-L117). Good luck!

PR #232, opened 2016-01-18
Add CellRendererProgress for TreeViews

honorabrutroll:
This is a fairly simple but useful PR, just adds CellRendererProgress to Gir.toml and adds the generated files. The results can be seen [here](https://github.com/honorabrutroll/examples/blob/pending/src/treeview.rs#L19).


honorabrutroll:
Note: something similar can probably done to add CellRendererPixbuf, which could address issue #213 very easily.

gkoz:
Thanks for taking the generator for a spin! Sadly it doesn't handle properties so these cell renderer objects may still be quite inadequate.
Would you mind trying to add the whole lot of missing `CellRenderer*` objects to`Gir.toml`?
Also the current policy is to keep changes to autogenerated files in separate commits.

honorabrutroll:
In my example it was able to at least render the progress bar with the desired percentage, so I believe the properties aren't that important at the moment except for very specialized use cases. Adding the missing ```CellRenderer*``` objects should be fairly easy. So just to clarify, I would first add the objects to ```Gir.toml``` in one commit and then add the autogenerated files in a separate commit?

gkoz:
Yes, that's right!

honorabrutroll:
Alright, from my searches of the pygtk documentation, I'm getting these ```CellRenderer*``` objects to add in:
* ```CellRendererAccel```
* ```CellRendererCombo```
* ```CellRendererPixBuf```
* ```CellRendererSpin```
* ```CellRendererSpinner```

Does this list look right to you?

gkoz:
The complete hierarchy is here: https://developer.gnome.org/gtk3/stable/ch02.html Looks like there are
```text
    │   ├── GtkCellRenderer
    │   │   ├── GtkCellRendererText
    │   │   │   ├── GtkCellRendererAccel
    │   │   │   ├── GtkCellRendererCombo
    │   │   │   ╰── GtkCellRendererSpin
    │   │   ├── GtkCellRendererPixbuf
    │   │   ├── GtkCellRendererProgress
    │   │   ├── GtkCellRendererSpinner
    │   │   ╰── GtkCellRendererToggle
```

honorabrutroll:
Whups, should've used reset instead of revert.

honorabrutroll:
Fixed.

Gir.toml
```diff
@@ -26,6 +26,12 @@ generate = [
     "Gtk.CellEditable",
     "Gtk.CellLayout",
     "Gtk.CellRendererText",
+    "Gtk.CellRendererAccel",
```
gkoz: Could you preserve the alphabetical order please?

gkoz:
Perfect.
ping @GuillaumeGomez

GuillaumeGomez:
@honorabrutroll: Thanks for you PR!
@gkoz: Thanks for the review (why weren't these files not included at first?)!

gkoz:
I had largely concentrated on not losing the existing APIs rather than including everything to avoid further delaying the merge.

PR #235, opened 2016-01-19
AppVeyor: Use libgtk+ from msys2

### LINKED ISSUE #234: Use msys2 for the AppVeyor build
I'm opening this to ask if it would be appreciated if I would change appveyor.yml to use msys2 to do the builds.
Advantages:
 - The [official guide](http://www.gtk.org/download/windows.php) to build on Windows recommends using msys2
 - Probably simpler build script

Disadvantages:
 - Only one GTK version supported

rnestler:
This is only to test if the build would work with this configuration.
It would fix #234

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,59 +1,18 @@
-platform:
-  - x86
-  - x64
 environment:
   matrix:
-    - RUST_VERSION: 1.5.0
-      GTK: gnome
-    - RUST_VERSION: 1.5.0
-      GTK: win64
-      FEATURES: gtk_3_14
-    #- RUST_VERSION: nightly
-      #GTK: gnome
-
-matrix:
-  #fast_finish: true
-  allow_failures:
-    - platform: x86
-      GTK: win64
-
+    - TARGET: i686-pc-windows-gnu
+      RUST_VERSION: 1.5.0
+      BITS: 32
+    - TARGET: x86_64-pc-windows-gnu
+      RUST_VERSION: 1.5.0
+      BITS: 64
 install:
-  - if "%PLATFORM%-%GTK%" == "x86-win64" ( exit 1 )
-  - if "%PLATFORM%" == "x86" SET RUST_PLATFORM=i686-pc-windows-gnu
-  - if "%PLATFORM%" == "x64" SET RUST_PLATFORM=x86_64-pc-windows-gnu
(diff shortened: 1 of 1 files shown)


gkoz:
What can I say, it does seem to work :)

GuillaumeGomez:
@gkoz: So what do you say? Do you want we merge it or not?

gkoz:
Okay, I guess I'm fine with having just this mode in appveyor. I'd like to avoid unnecessary verbosity though:

appveyor.yml
```diff
@@ -1,59 +1,19 @@
...
-      }
-  - cmd: SET PATH=%MINGW%\bin;C:\Deps\gtk\bin;C:\Rust\bin;%PATH%
-
-cache:
-  C:\Deps
+  - ps: Start-FileDownload "https://static.rust-lang.org/dist/rust-${env:RUST_VERSION}-${env:TARGET}.exe"
+  - rust-%RUST_VERSION%-%TARGET%.exe /VERYSILENT /NORESTART /COMPONENTS="Rustc,Cargo,Std" /DIR="C:\Rust"
+  - SET PATH=C:\Rust\bin;C:\Program Files\Git\cmd;C:\Program Files\Git\usr\bin
+  - SET PATH=C:\msys64\mingw%BITS%\bin;C:\msys64\usr\bin;%PATH%
+  - echo %PATH%
```
gkoz: I don't think we need this

appveyor.yml
```diff
@@ -61,7 +21,6 @@ build_script:
   - echo paths = ["."] > .cargo\config
   - git clone -q --depth 50 -b pending https://github.com/gtk-rs/examples _examples
   - cd _examples
-  - "echo Features: %FEATURES%"
-  - cargo build --features "%FEATURES%"
+  - cargo build --verbose
```
gkoz: The `echo` and `--verbose` are not necessary now but the `--features gtk_3_14` is.

appveyor.yml
rnestler: The `echo %PATH%`? Yeah I can remove that.

appveyor.yml
rnestler: OK

gkoz:
Thanks, looks good :+1:

GuillaumeGomez:
Thanks!

PR #238, opened 2016-01-23
Make `Value` APIs safer and remove redundancies

gkoz:


gkoz:
While meddling with `set_value` it seems I should probably change it to `set`...

GuillaumeGomez:
I agree. Since similar changes like this have been made all around, why not here as well?

PR #244, opened 2016-01-29
Use safe event wrappers in signal handler signatures

gkoz:
The wrappers don't have any safe getters yet but they implement `AsRef`
for the respective FFI structs so all the event fields remain available
via `as_ref()`.

[breaking change]


gkoz:
https://github.com/gtk-rs/gdk/pull/95

PR #245, opened 2016-02-03
Implement `embed-lgpl-docs` and `purge-lgpl-docs` features

gkoz:
Part of #8

`embed-lgpl-docs`: embed documentation comments contained in the gtk-rs-lgpl-docs package into the sources.
`purge-lgpl-docs`: remove all LGPL doc comments from the source.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,10 @@ addons:
 script:
   - rustc --version
   - ./check_init_asserts
+  - cargo test --features embed-lgpl-docs
+  # catch any sneaked in lgpl docs
+  - cargo build --features purge-lgpl-docs
+  - git diff -R --exit-code
   - mkdir .cargo
   - echo 'paths = ["."]' > .cargo/config
   - git clone -q --depth 50 -b pending https://github.com/rust-gnome/examples _examples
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Maybe specify that such features exist and even add a little example in the README is needed here.

gkoz:
I wouldn't want to endorse these in any way before the docs repo is populated and ready to accept PRs from users wanting to improve the docs.

GuillaumeGomez:
Then we'd better wait before merging this PR.

gkoz:
After thinking about format changes a little it might be preferable to make `gtk-rs-lgpl-docs` depend on `stripper_lib` and move the substance of the code from this PR there. The code won't be duplicated across build scripts then and the lgpl-docs crate will always depend on the compatible version of stripper_lib (one that understands the particular cmts format).

GuillaumeGomez:
Since stripper_lib has been created, why not using it? ^^

gkoz:
Implemented the idea from my last comment.

gkoz:
I think this is good to go.

GuillaumeGomez:
Thanks!

PR #247, opened 2016-02-08
Minor cleanups and removals

gkoz:
`LockButton` is impossible to use until there're meaningful `GPermission` bindings.

Similarly, `notify` and `notify-children` signals require work on `GParamSpec`.


GuillaumeGomez:
Thanks!

Issue #243, opened 2016-01-28
[RFC] Version autodetection and API stability


gkoz:
I'm getting more worried about the decision to detect the library version instead of setting it with a feature (#120). Beside being magic (magic can be confusing and frustrating) this is also a forward compatibility hazard.

Adding an `enum` variant is a breaking change in Rust because of `match` exhaustiveness rules. Newer GTK releases sometimes do add them. If such enum variants are under `#[cfg]` this will e.g. cause code that built with GTK 3.16 libraries to stop building on a GTK 3.18 system. If all known today enum variants are exposed, there still can appear new ones and we'd have to bump major version when adding a variant. Imagine going from 1.0 to 2.0 because of something like `GDK_TOUCHPAD_SWIPE`.

I'm considering bringing back version features and perhaps keeping autodetection as another feature.

If I'm not mistaken this could have a noticeable upside: now that we don't build any shims our only external dependency is `pkg-config` used to detect versions. Dropping [by default] `pkg-config` would make development packages (`libgtk3-dev` and the like) unnecessary.

GuillaumeGomez:
I had a solution between all this: creating a little script which then displays something like:

```Shell
> ./detect_version
The gtk version detected on your computer is 3.13. You should run cargo with the GTK_3_13 feature (except if you have another version).
```

What do you think of this?

gkoz:
Thing is, you should not give cargo any random version you see on your local computer. You should be considering the versions your app users might have on their systems.

gkoz:
I mean, the point mostly is that the user needs to deliberately select a version. If they just don't care, well then we can still offer detection as a feature.

gkoz:
> If I'm not mistaken this could have a noticeable upside: now that we don't build any shims our only external dependency is pkg-config used to detect versions. Dropping [by default] pkg-config would make development packages (libgtk3-dev and the like) unnecessary.

I've just realized that the `pkg-config` can be avoided regardless because all of the libraries expose some version functions e.g. `glib_{major|minor|micro}_version`.

gkoz:
I'm implementing version features with a slightly different syntax than before, e.g.
```toml
[features]
"3.6" = ["gtk-sys/3.6", "gdk/3.6"]
"3.8" = ["3.6", "gtk-sys/3.8", "gdk/3.8"]
"3.10" = ["3.8", "gtk-sys/3.10", "gdk/3.10"]
[...]

PR #248, opened 2016-02-12
Switch to version features instead of autodetection

gkoz:
Implement the reinstatement of version features outlined in https://github.com/gtk-rs/gtk/issues/243.

- Installed library version detection no longer impacts conditional compilation.
- Each crate defaults to supporting a particular minimal upstream version.
- Higher (newer) versions need to be enabled via cargo features e.g. `--features 3.16`, `features = ["3.16"]`.
- The selected version is checked against `pkg-config` aborting the build if it's too high.
- The use of `pkg-config` can be overridden by setting `GTK_LIB_DIR` env variable, which translates into `-L $GTK_LIB_DIR` linker argument. No version checks are done in this case.

[breaking change]


gkoz:
Naturally this can't be tested by our CI tools. It's been preliminary tested locally but we'll only know after merging everything.

PR #252, opened 2016-02-22
Remove GTK_LIB_DIR from appveyor it's no longer required

gkoz:
Also update the changelog a bit


gkoz:
Eh, I didn't mean to delete `socket.rs` in this PR :/

PR #254, opened 2016-02-24
Fix generic signal trampolines

gkoz:
This was supposed to be a part of #226.
The bug wasn't harmful because all object wrappers have the same representation.
Thanks to @EPashkin for noticing suspicious code.


GuillaumeGomez:
Nicely done. Thanks to both of you!

PR #257, opened 2016-03-01
Add some documentation and gdk-pixbuf dependency

gkoz:


src/lib.rs
```diff
@@ -2,6 +2,81 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+//! # GTK+ 3 bindings
+//!
+//! This library contains safe Rust bindings for [GTK+ 3](http://www.gtk.org), a
+//! multi-platform GUI toolkit. It's a part of [Gtk-rs](http://gtk-rs.org/).
```
GuillaumeGomez: I think it's better to write: "It's part of" instead of "It's a part of".

src/lib.rs
```diff
@@ -2,6 +2,81 @@
...
+//!
+//! `Cargo.toml` example:
+//!
+//! ```toml
+//! [dependencies.gtk]
+//! version = "0.x.y"
+//! features = ["3.16"]
+//! ```
+//!
+//! Take care when choosing which version to target: some of your users might
```
GuillaumeGomez: "Take care" formulation sounds weird for me.

src/lib.rs
gkoz: Parts as in libraries are countable so the article is necessary.

src/lib.rs
gkoz: Is that enough to draw attention or should I perhaps make this sentence bold for greater effect? I feel it might be important to make users stop and think before choosing the version.

GuillaumeGomez: Nice idea! Something like "the greater the needed version, the lesser the users' number" could help. I didn't pay attention enough to this point but it's indeed really important.

gkoz:
Updated the versions paragraph.

GuillaumeGomez:
All good, thanks!

PR #258, opened 2016-03-02
Don't pass any borrowed wrappers as &mut references, it's unsound

gkoz:
A closure could use `mem::swap` to take the wrapper out of a mutable reference and wind up with a dangling pointer.


GuillaumeGomez:
Perfect, I'm taking a look. Thanks!

PR #261, opened 2016-03-04
Add TreeSortable missing functions

GuillaumeGomez:


src/tree_sortable.rs
```diff
@@ -0,0 +1,106 @@
...
+use std::process;
+
+use glib_ffi::gpointer;
+use {TreeSortable, TreeIter};
+use ffi::{GtkTreeIter, GtkTreeSortable};
+
+pub trait TreeSortableExt {
+    // -- manual --
+    fn set_default_sort_func<F>(&self, sort_func: F)
+    where F: FnMut(&TreeSortable, &TreeIter, &TreeIter) -> i32 + 'static;
```
gkoz: I think we could make this more flexible with regard to receiver, [like in signals](https://github.com/gtk-rs/gtk/blob/master/src/signal.rs#L135): `FnMut(&Self, ...`, so the user would be in control of the exact type.

Also the closure should probably return `cmp::Ordering`

src/tree_sortable.rs
```diff
@@ -0,0 +1,106 @@
...
+
+impl Drop for CallbackGuard {
+    fn drop(&mut self) {
+        if thread::panicking() {
+            process::exit(101);
+        }
+    }
+}
+
+macro_rules! callback_guard {
```
gkoz: It might be better to export the macro from `mod signals` and use it here.

src/tree_sortable.rs
```diff
@@ -0,0 +1,106 @@
...
+    #[inline]
+    fn has_default_sort_func(&self) -> bool {
+        Auto::has_default_sort_func(self)
+    }
+
+    fn set_default_sort_func<F>(&self, sort_func: F)
+        where F: FnMut(&TreeSortable, &TreeIter, &TreeIter) -> i32 + 'static {
+        unsafe {
+            ffi::gtk_tree_sortable_set_default_sort_func(self.to_glib_none().0,
+                                                         transmute(trampoline),
```
gkoz: I used to do transmutes like this myself but now I'd prefer to move to a different convention: make the trampoline's signature compatible with this function (`user_data: gpointer`) and use either `transmute` or `as` in the trampoline itself.

src/tree_sortable.rs
GuillaumeGomez: Oh right, I totally forgot to use `cmp::ordering`...

However, isn't it better to let the user handles the value he gets back? By returning the type he wants, it means we have to use `get_value` directly in our function (or maybe I misunderstood something?). If he wants to get back `TreeIter`, I don't see how we could do it.

gkoz: This is about the first parameter. If I have `store: TreeStore` then `store.set_default_sort_func()` will take `FnMut(&TreeStore, &TreeIter, &TreeIter) -> Ordering`. If I have `store: TreeSortable`, it will be `FnMut(&TreeSortable, &TreeIter, &TreeIter) -> Ordering`.

gkoz:
I think we can be pretty sure the comparisons won't be called recursively and just skip the `RefCell`, make it `Box<Box<FnMut...>>`.

src/tree_sortable.rs
GuillaumeGomez: Oh right! Sorry, I totally misunderstood. :-/

Then yes, just need to add a generic parameter and we're good to go!

gkoz:
How about changing `i32` to `SortColumn`? The automatic trait would become mostly useless but we'd have a more strongly-typed API for this trait.

gkoz:
> we can be pretty sure the comparisons

Uh, actually no, there's no evidence of this yet. :(

GuillaumeGomez:
For as far as I know, the `sort_column_id` corresponds to the position of the column inside the its view. Or perhaps you mean adding an unit struct to wrap the `i32`?

gkoz:
Would `enum SortColumn { None, Default, Some(u32) }` reflect sort_column_id semantics? It's either unsorted, sorted with the default function or by some column with non-negative index.

gkoz:
Can you test what happens if you change the sort column id from a comparator closure? I don't see any thing preventing recursion in libgtk. So the `RefCell` will probably panic. In that case the closures need to be `Fn`.

GuillaumeGomez:
I'll try once I'm back home.

gkoz:
I've tested calling `set_sort_column_id` from the comparator recursively and caused `RefCell` to panic. So `Fn` it must be after all.

src/tree_sortable.rs
```diff
@@ -0,0 +1,144 @@
...
+}
+
+pub trait TreeSortableExt {
+    // -- manual --
+    fn set_default_sort_func<F, T: TreeSortableExt>(&self, sort_func: F)
+    where F: FnMut(&T, &TreeIter, &TreeIter) -> Ordering + 'static;
+    fn set_sort_func<F, T: TreeSortableExt>(&self, sort_column_id: SortColumn, sort_func: F)
+    where F: FnMut(&T, &TreeIter, &TreeIter) -> Ordering + 'static;
+    // -- auto --
+    fn get_sort_column_id(&self) -> Option<(SortColumn, SortType)>;
```
gkoz: If the signature is different it's misleading to file this under "auto"

src/tree_sortable.rs
```diff
@@ -0,0 +1,144 @@
...
+        Auto::has_default_sort_func(self)
+    }
+
+    fn set_default_sort_func<F, T: TreeSortableExt>(&self, sort_func: F)
+        where F: FnMut(&T, &TreeIter, &TreeIter) -> Ordering + 'static {
+        unsafe {
+            ffi::gtk_tree_sortable_set_default_sort_func(self.to_glib_none().0,
+                                                         transmute(trampoline),
+                                                         into_raw(sort_func),
+                                                         Some(destroy_closure)) }
```
gkoz: Trailing closing brace.

src/tree_sortable.rs
```diff
@@ -0,0 +1,144 @@
...
+
+    fn set_sort_func<F, T: TreeSortableExt>(&self, sort_column_id: SortColumn,
+                                            sort_func: F)
+        where F: FnMut(&T, &TreeIter, &TreeIter) -> Ordering + 'static {
+        unsafe {
+            ffi::gtk_tree_sortable_set_sort_func(self.to_glib_none().0,
+                                                 sort_column_id.to_glib(),
+                                                 transmute(trampoline),
+                                                 into_raw(sort_func),
+                                                 Some(destroy_closure)) }
```
gkoz: Ditto

src/tree_sortable.rs
```diff
@@ -0,0 +1,144 @@
...
+impl FromGlib<i32> for SortColumn {
+    #[inline]
+    fn from_glib(val: i32) -> SortColumn {
+        if val == ffi::GTK_TREE_SORTABLE_DEFAULT_SORT_COLUMN_ID {
+            SortColumn::Default
+        } else if val == ffi::GTK_TREE_SORTABLE_UNSORTED_SORT_COLUMN_ID {
+            SortColumn::Unsorted
+        } else {
+            SortColumn::Column(val as u32)
+        }
```
mmstick: How about using a match instead of an `if else if` match?

```
match val {
    ffi::GTK_TREE_SORTABLE_DEFAULT_SORT_COLUMN_ID => SortColumn::Default,
    ffi::GTK_TREE_SORTABLE_UNSORTED_SORT_COLUMN_ID => SortColumn::Unsorted,
    _ => SortColumn::Column(val as u32)
}

Comment (GuillaumeGomez) [inline src/tree_sortable.rs]: Indeed, well saw.

File: src/tree_sortable.rs (hunk 4 above)
Comment (GuillaumeGomez) [inline src/tree_sortable.rs]: Why that? We even declare functions which aren't in the other trait, so I'm not too bothered by this.

Comment (gkoz) [inline src/tree_sortable.rs]: Everything under "auto" heading I would assume comes by copy-pasting from the auto trait. But this isn't the case here.

File: src/tree_sortable.rs [hunk 8]
```diff
@@ -0,0 +1,132 @@
...
+
+use SortType;
+use ffi;
+use glib::object::IsA;
+use glib::translate::*;
+use auto::traits::TreeSortableExt as Auto;
+use std::cell::RefCell;
+use std::mem::transmute;
+use std::cmp::Ordering;
+use std::i32::MAX;
```
gkoz: Does `i32::max_value()` not work? Anyway importing just `MAX` is confusing.

src/tree_sortable.rs
GuillaumeGomez: I'll move the modified functions into manual then.

src/tree_sortable.rs
```diff
@@ -0,0 +1,132 @@
...
+                SortColumn::Column(x as u32)
+            }
+        }
+    }
+}
+
+pub trait TreeSortableExt {
+    // -- manual --
+    fn set_default_sort_func<F, T: TreeSortableExt>(&self, sort_func: F)
+    where F: FnMut(&T, &TreeIter, &TreeIter) -> Ordering + 'static;
```
gkoz: This does not look like the signature I suggested and the example from mod signal.

src/prelude.rs
```diff
@@ -57,6 +57,7 @@ pub use app_chooser::AppChooserExt;
 pub use color_chooser::ColorChooserExt;
 pub use dialog::DialogExt;
 pub use tree_model::TreeModelExt;
+pub use tree_sortable::{SortColumn, TreeSortableExt};
```
gkoz: We don't need `SortColumn` in the prelude, it's not important enough. But it should be reexported in `lib.rs`.

src/tree_sortable.rs
```diff
@@ -0,0 +1,131 @@
...
+use signal::CallbackGuard;
+
+use glib_ffi::gpointer;
+use {TreeSortable, TreeIter};
+use ffi::{GtkTreeIter, GtkTreeSortable};
+
+pub enum SortColumn {
+    Default,
+    Unsorted,
+    Column(u32),
```
gkoz: Considering there already is "Column" in the enum name, maybe `Id(u32)`? That would rhyme with the method names.

Alternatively we could name this variant `Index` and drop `_id` from the methods. Then the method names would nicely match the type.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+impl ToGlib for SortColumn {
+    type GlibType = i32;
+
+    #[inline]
+    fn to_glib(&self) -> i32 {
+        match *self {
+            SortColumn::Default => ffi::GTK_TREE_SORTABLE_DEFAULT_SORT_COLUMN_ID,
+            SortColumn::Unsorted => ffi::GTK_TREE_SORTABLE_UNSORTED_SORT_COLUMN_ID,
+            SortColumn::Index(x) => {
+                assert!(x <= i32::max_value() as u32, "invalid column index");
```
gkoz: The message could be more specific, it's not just invalid, it's too big.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
+// This file was generated by gir (17af302) from gir-files (11e0e6d)
+// DO NOT EDIT
```
gkoz: This probably shouldn't be here.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+use {TreeSortable, TreeIter};
+use ffi::{GtkTreeIter, GtkTreeSortable};
+
+pub enum SortColumn {
+    Default,
+    Unsorted,
+    Index(u32),
+}
+
+impl ToGlib for SortColumn {
```
gkoz: I like to hide these with `#[doc(hidden)]`

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+                SortColumn::Index(x as u32)
+            }
+        }
+    }
+}
+
+pub trait TreeSortableExt {
+    // -- manual --
+    fn set_default_sort_func<F>(&self, sort_func: F)
+    where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static;
```
gkoz: It would probably be more readable to indent `where` in the trait definition.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn set_sort_column_id(&self, sort_column_id: SortColumn, order: SortType);
+    // -- auto --
+    fn has_default_sort_func(&self) -> bool;
+    fn sort_column_changed(&self);
+}
+
+unsafe extern "C" fn trampoline<T>(this: *mut GtkTreeSortable, iter: *mut GtkTreeIter,
+                                   iter2: *mut GtkTreeIter,
+                                   f: &mut Box<Fn(&T, &TreeIter, &TreeIter) -> Ordering>)
+    where T: IsA<TreeSortable> {
```
gkoz: ...but keep it unindented elsewhere.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+    fn get_sort_column_id(&self) -> Option<(SortColumn, SortType)>;
+    fn set_sort_column_id(&self, sort_column_id: SortColumn, order: SortType);
+    // -- auto --
+    fn has_default_sort_func(&self) -> bool;
+    fn sort_column_changed(&self);
+}
+
+unsafe extern "C" fn trampoline<T>(this: *mut GtkTreeSortable, iter: *mut GtkTreeIter,
+                                   iter2: *mut GtkTreeIter,
+                                   f: &mut Box<Fn(&T, &TreeIter, &TreeIter) -> Ordering>)
```
gkoz: If we forgo closure mutability the borrow of `f` has to be immutable.

But as per an earlier comment, I'd rather the type here were `gpointer`, cast to `&Box` in the body.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+    where T: IsA<TreeSortable> {
+    callback_guard!();
+    f(&TreeSortable::from_glib_none(this).downcast_unchecked(), &from_glib_borrow(iter),
+      &from_glib_borrow(iter2));
+}
+
+unsafe extern "C" fn destroy_closure<T>(ptr: gpointer)
+    where T: IsA<TreeSortable> {
+    callback_guard!();
+    Box::<Box<Fn(&TreeSortable, &TreeIter, &TreeIter) -> Ordering + 'static>>::from_raw(ptr as *mut _);
```
gkoz: `T` is not used here.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+}
+
+fn into_raw<F, T: TreeSortableExt>(func: F) -> gpointer
```
gkoz: Looks like neither `into_raw` nor `destroy_closure` need any bounds on `T`.

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+    Box::into_raw(func) as gpointer
+}
+
+impl<O: IsA<TreeSortable>> TreeSortableExt for O {
+    #[inline]
+    fn get_sort_column_id(&self) -> Option<(SortColumn, SortType)> {
+        match Auto::get_sort_column_id(self) {
+            Some((id, ty_)) => Some((from_glib(id), ty_)),
+            None => None,
+        }
```
gkoz: This implementation does not agree with function's documentation: https://developer.gnome.org/gtk3/stable/GtkTreeSortable.html#gtk-tree-sortable-get-sort-column-id

src/tree_sortable.rs
```diff
@@ -0,0 +1,134 @@
...
+                x as i32
+            }
+        }
+    }
+}
+
+impl FromGlib<i32> for SortColumn {
+    #[inline]
+    fn from_glib(val: i32) -> SortColumn {
+        match val {
```
gkoz: Travis tells me this function should begin with
```
skip_assert_initialized!();
```

gkoz:
After taking a closer look at `get/set_sort_column_id` it seems an even more precise treatment of `SortColumn` is needed:

```rust
pub enum SortColumn {
    Default,
    Index(u32),
}

trait TreeSortableExt {
    fn get_sort_column(&self) -> Option<(SortColumn, SortType)>;
    fn set_sort_column(&self, sort_column: Option<(SortColumn, SortType)>);
}
```

Why? The functions deal with `id` and `SortType` together but `SortType` is meaningless in the unsorted mode so there needs to be an `Option` somewhere despite `SortColumn` having an `Unsorted` variant. If we pull the option out of the tuple, `None` means "unsorted".

A variation on this scheme avoids `Option` in the `set_` method:

```rust
    fn set_sort_column(&self, sort_column: SortColumn, order: SortType);
    fn set_unsorted(&self);
```

src/tree_sortable.rs
GuillaumeGomez: I don't understand what's wrong here.

gkoz: `Auto::get_sort_column_id` is implemented under an incorrect assumption. It will never return a negative id.

src/lib.rs
```diff
@@ -285,6 +285,7 @@ mod recent_data;
 mod recent_info;
 mod rectangle;
 mod requisition;
+#[macro_use]
```
gkoz: #262 has made this unnecessary.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+use std::mem::{self, transmute};
+use std::cmp::Ordering;
+
+use glib_ffi::gpointer;
+use {TreeSortable, TreeIter};
+use ffi::{GtkTreeIter, GtkTreeSortable};
+
+pub enum SortColumn {
+    Default,
+    Unsorted,
```
gkoz: I think we agreed to remove this variant.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+}
+
+impl<O: IsA<TreeSortable>> TreeSortableExt for O {
+    #[inline]
+    fn get_sort_column_id(&self) -> Option<(SortColumn, SortType)> {
+        unsafe {
+            let mut sort_column_id = mem::uninitialized();
+            let mut order = mem::uninitialized();
+            let ret = from_glib(ffi::gtk_tree_sortable_get_sort_column_id(self.to_glib_none().0, &mut sort_column_id, &mut order));
+            if ret || sort_column_id != ffi::GTK_TREE_SORTABLE_UNSORTED_SORT_COLUMN_ID {
```
gkoz: `ret` is redundant because you have to look at `id` anyway.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+                                   iter2: *mut GtkTreeIter,
+                                   f: gpointer) -> i32
+where T: IsA<TreeSortable> {
+    callback_guard!();
+    let f : &Box<Fn(&T, &TreeIter, &TreeIter) -> Ordering> = transmute(f);
+    f(&TreeSortable::from_glib_none(this).downcast_unchecked(), &from_glib_borrow(iter),
+      &from_glib_borrow(iter2)).to_glib()
+}
+
+unsafe extern "C" fn destroy_closure(ptr: gpointer) {
```
gkoz: This might make no difference in practice but I'd prefer it to be generic over `T` like `into_raw`.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+    #[inline]
+    fn has_default_sort_func(&self) -> bool {
+        Auto::has_default_sort_func(self)
+    }
+
+    fn set_default_sort_func<F>(&self, sort_func: F)
+        where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static {
+        unsafe {
+            ffi::gtk_tree_sortable_set_default_sort_func(self.to_glib_none().0,
+                                                         transmute(trampoline::<Self>),
```
gkoz: Now that `trampoline` takes `gpointer`, the `transmute` should not be needed.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+        Auto::set_sort_column_id(self, SortColumn::Unsorted.to_glib(), SortType::Ascending)
+    }
+
+    fn set_sort_func<F>(&self, sort_column_id: SortColumn,
+                        sort_func: F)
+        where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static {
+        unsafe {
+            ffi::gtk_tree_sortable_set_sort_func(self.to_glib_none().0,
+                                                 sort_column_id.to_glib(),
+                                                 transmute(trampoline::<Self>),
```
gkoz: As above

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn has_default_sort_func(&self) -> bool;
+    fn sort_column_changed(&self);
+}
+
+unsafe extern "C" fn trampoline<T>(this: *mut GtkTreeSortable, iter: *mut GtkTreeIter,
+                                   iter2: *mut GtkTreeIter,
+                                   f: gpointer) -> i32
+where T: IsA<TreeSortable> {
+    callback_guard!();
+    let f : &Box<Fn(&T, &TreeIter, &TreeIter) -> Ordering> = transmute(f);
```
gkoz: Space before `:`

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+    fn set_sort_column_id(&self, sort_column_id: SortColumn, order: SortType) {
+        Auto::set_sort_column_id(self, sort_column_id.to_glib(), order)
+    }
+
+    fn set_unsorted(&self) {
+        Auto::set_sort_column_id(self, SortColumn::Unsorted.to_glib(), SortType::Ascending)
+    }
+
+    fn set_sort_func<F>(&self, sort_column_id: SortColumn,
+                        sort_func: F)
```
gkoz: Looks like this line can be unwrapped now.

src/tree_sortable.rs
```diff
@@ -0,0 +1,146 @@
...
+        where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static {
```
gkoz: And but I'd prefer `where` aligned with `fn` everywhere outside of `trait Foo { }`

src/tree_sortable.rs
GuillaumeGomez: It's to avoid a potential error I did it this way. But if you prefer, I can just remove `ret`.

gkoz: If (unlikely) a new variant (e.g. `-3`) appears this will fail silently instead of tripping the assertion.

src/tree_sortable.rs
GuillaumeGomez: it is, it's expecting to take an `Option` so the `transmute` call is unavoidable.

gkoz: Come on it's done two lines below without a transmute.

gkoz:
I've just had an idea: we might actually be able to cut down on all that `Fn(Foo, Bar)` boilerplate. If we define 
```rust
pub trait TreeIterCompareFunc<T>: Fn(&T, &TreeIter, &TreeIter) -> Ordering + 'static { }
```
this trait might make everything more readable.

gkoz:
We can merge as is and then I'll play around with various trait-based approaches.

gkoz:
Thanks for implementing this!

PR #262, opened 2016-03-05
Documentation, RecentData fix, callback_guard tweaks

gkoz:


gkoz:
The last commit makes it unnecessary to `#[macro_use]` on `mod signal` in #261.

PR #268, opened 2016-03-08
Add Application and ApplicationWindow, add TreeView::get_selection quirk

gkoz:
The examples build will fail because of ca24ec0


GuillaumeGomez:
Thanks! Can you squash the two last commits please?

gkoz:
Travis wants to update submodules by default. Interesting... I guess I'll have to merge the gir PR after all.

> Can you squash the two last commits please?

Any reason to do that? I'm not a fan of putting not exactly related changes in one commit.

GuillaumeGomez:
The "update gir" and the "regen" ones could be only one. I don't see the point to have both of them.

gkoz:
Regeneration follows three commits, each influencing the generated code.

gkoz:
Updated. The examples build failure handled in https://github.com/gtk-rs/examples/pull/75

PR #265, opened 2016-03-07
Add insert_with_values to ListStore and TreeStore

gkoz:

PR #269, opened 2016-03-08
Fix invalid insert_with_values loop

GuillaumeGomez:
Fixes #265


gkoz:
The `set` methods are where the bug originated.

gkoz:
Heh, it's not fair that this PR fails the tests because #268 hasn't been merged :)

GuillaumeGomez:
Isn't it? :p

PR #271, opened 2016-03-09
Mention event getters in the changelog

gkoz:


CHANGELOG.md
```diff
@@ -31,8 +31,8 @@
   specific receiver types][gtk226-2].
 
 - New [safe event bindings][gdk95] are [passed to signal handlers][gtk244].
-  These bindings don't have many getters and setters implemented yet but the low
-  level event structs are still reachable [via `as_ref`][examples68].
+  These bindings have [safe getters][gdk107] and the low-level event structs are
```
GuillaumeGomez: The "and" doesn't seem right here. We should mark it more strongly. Replacing it with "but", "while", "yet" or any equivalent to strenghten the two parts of the sentence.

gkoz:
Good point!

GuillaumeGomez:
Thanks!

PR #272, opened 2016-03-10
Implement Application::new, make it an alternative to gtk::init

gkoz:


src/application.rs
```diff
@@ -0,0 +1,27 @@
+use Application;
+use ffi;
+use glib::ApplicationFlags;
+use glib::translate::*;
+use rt;
+
+impl Application {
+    #[cfg(feature = "3.6")]
+    pub fn new(application_id: Option<&str>, flags: ApplicationFlags) -> Result<Application, ()> {
+        skip_assert_initialized!();
```
GuillaumeGomez: "And if you are using GtkApplication, you don't have to call any of the initialization functions either; the “startup” handler does it for you.".

GuillaumeGomez:
Thanks! And well saw for the useless calls to `gtk_init`.

gkoz:
To elaborate, if we'd let `Application::new` be the first to call `gtk_init` we would still have to guard against having already called `init` from another thread and it would kill the process upon failure.

GuillaumeGomez:
That's what I was looking for: if the double call of `gtk_init` is a failure. That's annoying.

gkoz:
Uh, actually the "another thread" check happens too late now :( see #273

fbruetting:
Hey guys, could you please explain how a Rust noob could accomplish this? Sounds silly, but I have a hard time specifying ApplicationFlags, I tried several versions of `use self::glib::ApplicationFlags` like above, but no matter how I try it, everytime I receive an "unresolved import".

In the [Gtk-rs manual](http://gtk-rs.org/docs/gtk/struct.Application.html#method.new), the parameter is [linked](http://gtk-rs.org/docs/gio/struct.ApplicationFlags.html), but there is no description. I also found no code I could steal – all programs I found use `gtk::init()`.

EPashkin:
@fbruetting, these flags defined in gio. Try this:
```
use gio::ApplicationFlags;
use gtk::Application;

let app = Application::new("app", ApplicationFlags::APPLICATION_FLAGS_NONE).unwrap();
```

fbruetting:
With
```
extern crate gtk;
extern crate gio;
use self::gtk::prelude::*;
use self::gtk::Application;
use self::gio::ApplicationFlags;
…
let application = Application::new (Some("<bla>"), ApplicationFlags::APPLICATION_FLAGS_NONE).unwrap ();
```
I get:
```
error: no associated item named `APPLICATION_FLAGS_NONE` found for type `gui::gtk3::gio::ApplicationFlags` in the current scope
  --> src/gui/gtk3/mod.rs:34:74
   |
34 | 	let application = Application::new (Some("de.frankbruetting.Kommando"), ApplicationFlags::APPLICATION_FLAGS_NONE).unwrap ();
   | 	                                                                        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

I don't know why I have to put `self::` in front of the usage commands, but this way everything worked yet – could this be the reason?

EPashkin:
Seems I don't understand how bitflags works. We need use not `ApplicationFlags` but constant itself
```
use gio::APPLICATION_FLAGS_NONE;
let application = Application::new ("<bla>", APPLICATION_FLAGS_NONE).unwrap ();
```

EPashkin:
Maybe you need `self::` if your `extern crate ..` not in crate root file.

PR #273, opened 2016-03-10
Add the "different threads" check to `init`

gkoz:
Move `gdk::set_initialized` call to `set_initialized`.

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -76,6 +76,7 @@ pub unsafe fn set_initialized() {
     else if is_initialized() {
         panic!("Attempted to initialize GTK from two different threads.");
     }
+    gdk::set_initialized();
     INITIALIZED.store(true, Ordering::Release);
     IS_MAIN_THREAD.with(|c| c.set(true));
 }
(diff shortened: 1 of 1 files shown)


src/rt.rs
```diff
@@ -93,10 +94,14 @@ pub unsafe fn set_initialized() {
 /// initialized otherwise an Err is returned.
 pub fn init() -> Result<(), ()> {
     skip_assert_initialized!();
+    if is_initialized_main_thread() {
+        return Ok(());
+    }
+    else if is_initialized() {
```
GuillaumeGomez: Can you move the `else if` one line up please?

gkoz: I'd have to fix the style in the whole file then.

GuillaumeGomez: Arf, didn't pay attention to the rest of the file. Motivated to do it?

GuillaumeGomez: I never expected it to go anywhere, that's way beyond my expectations!

GuillaumeGomez:
Well, we'll go back on code formatting later. Thanks @gkoz!

PR #276, opened 2016-03-14
Serious commit

GuillaumeGomez:
cc @gkoz


gkoz:
So looks like this is going to make the 0.0.7 release. I'm not a fan of completely meaningless commit messages though.

GuillaumeGomez:
You'll have to wait this evening then. ^^

GuillaumeGomez:
Updated.

gkoz:
Thanks!

PR #277, opened 2016-03-15
Change version features to vx_y format

gkoz:


gkoz:
These PRs can't pass tests due to cargo manifest changes. Is suggest merging all but https://github.com/gtk-rs/examples/pull/78 and restarting the tests on that one.

PR #279, opened 2016-03-17
Tree model filter

GuillaumeGomez:
cc @gkoz


Gir.toml
```diff
@@ -612,6 +612,45 @@ status = "generate"
...
+    [[object.function]]
+    name = "clear_cache"
+    [[object.function]]
+    name = "get_model"
+    [[object.function]]
+    name = "refilter"
+    [[object.function]]
+    name = "set_visible_column"
+    [[object.function.parameter]]
+        name = "column"
```
EPashkin: This and all before unneeded as it contains only "filter" params without any "action" params (like 'const')

Gir.toml
```diff
@@ -612,6 +612,45 @@ status = "generate"
...
+        [[object.function.parameter]]
+        name = "filter_iter"
+        const = true
+    [[object.function]]
+    name = "convert_path_to_child_path"
+        [[object.function.parameter]]
+        name = "filter_path"
+        const = true
+    [[object.function]]
+    name = "convert_path_to_child_path"
```
EPashkin: This duplicate of previous

Gir.toml
```diff
@@ -612,6 +612,45 @@ status = "generate"
     ignore = true
 
 [[object]]
+name = "Gtk.TreeModelFilter"
+status = "generate"
+    [[object.function]]
+    name = "clear_cache"
+    [[object.function]]
+    name = "get_model"
```
gkoz: It seems this widget should always have an associated model so we could make this function non-nullable.

GuillaumeGomez: For other functions as well, I wanted you to confirm if the `Option` wrapping was needed.

src/auto/tree_model_filter.rs
```diff
@@ -0,0 +1,78 @@
...
+
+    pub fn convert_child_iter_to_iter(&self, child_iter: &TreeIter) -> Option<TreeIter> {
+        unsafe {
+            let mut filter_iter = TreeIter::uninitialized();
+            let ret = from_glib(ffi::gtk_tree_model_filter_convert_child_iter_to_iter(self.to_glib_none().0, filter_iter.to_glib_none_mut().0, mut_override(child_iter.to_glib_none().0)));
+            if ret { Some(filter_iter) } else { None }
+        }
+    }
+
+    pub fn convert_child_path_to_path(&self, child_path: &TreePath) -> TreePath {
```
gkoz: I wonder why this doesn't return `Option<TreePath>`

gkoz:
The $64,000 question is where did the constructor go? :mag: There could be a bug in the generator.

GuillaumeGomez:
I updated the code and fixed what @EPashkin found out. Now remains the constructor and `convert_child_path_to_path` issues.

src/tree_model_filter.rs
```diff
@@ -0,0 +1,80 @@
...
+use ffi;
+use glib::object::{Downcast, IsA};
+use glib::translate::*;
+use std::mem::transmute;
+
+use glib_ffi::{gboolean, gpointer};
+use {TreeIter, TreeModel, TreeModelFilter};
+use ffi::{GtkTreeIter, GtkTreeModel};
+
+unsafe extern "C" fn trampoline<T>(this: *mut GtkTreeModel, iter: *mut GtkTreeIter,
```
gkoz: Since we don't have a `TreeModelFilterExt` trait here there's no point making these functions generic.

GuillaumeGomez:
Updated.

EPashkin:
By https://developer.gnome.org/gtk3/stable/GtkTreeModelFilter.html in has constructor `gtk_tree_model_filter_new`, but in gir it lies in `GtkTreeModel`. Very strange.

gkoz:
I've located the constructor, it's a [`TreeModel` method](https://github.com/gtk-rs/gtk/blob/821cd9eef6e73844415a501c91b2db465f2e2898/src/auto/tree_model.rs#L76-L80). This isn't a gir bug, gir is told to do that (`Gtk-3.0.gir` line `118268` in `gir-files` master).

See also https://bugzilla.gnome.org/show_bug.cgi?id=646742.

I think we should fix `filter_new` to make the return type non-nullable and add a manual `new` constructor, that would return `TreeModelFilter`.

EPashkin:
If `gtk_tree_model_filter_new` really create only `TreeModelFilter` IMHO better hide function in `TreeModel` and add manual as @gkoz propose

EPashkin:
But anyway it needed be tested on "real" example.

GuillaumeGomez:
@EPashkin: I like the fact that we can get the filter with `TreeModel`. So I think it's better to keep the method there and also add the missing one on `TreeModelFilter`.

As for the "real" example, I'll add it on my process-viewer as soon as it's merged.

gkoz:
Yeah, hiding `filter_new` probably is a better idea. It being a `tree_model` method really is a misnomer, `TreeModel` has no such method so it can't be "implemented" by any class, it's [just a function](https://github.com/GNOME/gtk/blob/c94afa01c4bdc175d715c3e8573b02d8cb33325a/gtk/gtktreemodelfilter.c#L3773).

EPashkin:
@GuillaumeGomez  Problem at having constructor in `TreeModel` is that it returns `TreeModel` so we need add converter function and it theoretically can fail if it get not `TreeModelFilter`.
IMHO hide possibility of failing from end user.

GuillaumeGomez:
@EPashkin: Ok, that's weird. I didn't check because I thought that it returned `TreeModelFilter`. No point to have it then.

GuillaumeGomez:
So what do we do about this PR @gkoz?

gkoz:
Well, the mystery of missing `Option` in `convert_child_path_to_path` has been solved in https://github.com/gtk-rs/gir/pull/231. I guess we need to start by regenerating the crate (probably in a separate PR) and seeing its impact.

And I think we concluded that a constructor needs to be implemented manually, while `TreeModel::filter_new` set to be ignored.

GuillaumeGomez:
I understood it the other way around (that the auto would generate it). My bad. I change it.

src/tree_model_filter.rs
```diff
@@ -0,0 +1,86 @@
...
+
+fn modify_into_raw<F, T>(func: F) -> gpointer
+    where F: Fn(&T, &TreeIter, &Value, u32) + 'static {
+    let func: Box<Box<Fn(&T, &TreeIter, &Value, u32) + 'static>> =
+        Box::new(Box::new(func));
+    Box::into_raw(func) as gpointer
+}*/
+
+impl TreeModelFilter {
+    pub fn new(child_model: &TreeModel, root: &mut TreePath) -> TreeModelFilter {
```
GuillaumeGomez: Here I return a `TreeModelFilter` whereas the gnome doc says it returns a `TreeModel`. I found it more logical this way. What do you think of it @gkoz?

GuillaumeGomez:
@gkoz: The check_init_asserts script fails but this method cannot be called without another Widget which has it. We need to add an exception maybe? :)

src/tree_model_filter.rs
EPashkin: Converting `TreeModel` to `TreeModelFilter` is only reason to writing manual `new`.
`root: Option<&TreePath>` as it can be NULL and don't seems need `mut`

EPashkin:
If you can, change commit name `Add TreeModelFilter` to `Add manual part for TreeModelFilter`.
Also IMHO need add `pub use tree_model_filter::TreeModelFilterExt;` to `prelude.rs`

gkoz:
> Also IMHO need add `pub use tree_model_filter::TreeModelFilterExt;` to prelude.rs

It's not an interface, there shouldn't be a trait.

src/tree_model_filter.rs
gkoz: Yeah the return type seems right and `root` should indeed be an immutable reference (see how the generator uses `mut_override`).

GuillaumeGomez:
Updated.

EPashkin:
Hide/comment `TreeModelExt::filter_new` in Gir.toml and auto\tree_model.rs

EPashkin:
IMHO better squash commits to one as `src/tree_model_filter.rs` changed in both.
@gkoz Maybe add `assert_initialized_main_thread!();` to `new` to stop `check_init_asserts` ?

gkoz:
My preference is having a "Regen" commit separate from others (some day we might have a bot do those commits). That commit should only touch files in `src/auto` of course.

The constructors probably needs `skip_assert_initialized!();`

src/tree_model_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+
+unsafe extern "C" fn destroy_closure(ptr: gpointer) {
+    callback_guard!();
+    Box::<Box<Fn(&TreeModel, &TreeIter) -> bool + 'static>>::from_raw(ptr as *mut _);
+}
+
+fn into_raw<F>(func: F) -> gpointer
+    where F: Fn(&TreeModel, &TreeIter) -> bool + 'static {
+    let func: Box<Box<Fn(&TreeModel, &TreeIter) -> bool + 'static>> =
+        Box::new(Box::new(func));
```
gkoz: This function feels like too much boilerplate. Maybe just move this line into constructor?

src/tree_model_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+    Box::into_raw(func) as gpointer
+}
+
+/*unsafe extern "C" fn modify_trampoline<T>(this: *mut GtkTreeModel, iter: *mut GtkTreeIter,
```
gkoz: I don't think we should check in the commented out code.

src/tree_model_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+}
+
+fn modify_into_raw<F, T>(func: F) -> gpointer
+    where F: Fn(&T, &TreeIter, &Value, u32) + 'static {
+    let func: Box<Box<Fn(&T, &TreeIter, &Value, u32) + 'static>> =
+        Box::new(Box::new(func));
+    Box::into_raw(func) as gpointer
+}*/
+
+impl TreeModelFilter {
```
gkoz: Can you move the `impl` above the `extern` functions so the first thing the reader sees in the file would be the public part not the implementation details?

src/tree_model_filter.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn new(child_model: &TreeModel, root: Option<&TreePath>) -> TreeModelFilter {
```
gkoz: One thing I missed here is this should probably be

```rust
pub fn new<T: IsA<TreeModel>(child_model: &T, root: Option<&TreePath>) -> TreeModelFilter
```

src/tree_model_filter.rs
GuillaumeGomez: The point is to avoid to rewrite the function I already wrote once while we wait for these functions to be uncommented.

src/tree_model_filter.rs
GuillaumeGomez: I'm wondering... This code doesn't really have its place into the constructor, it's nice thing to keep this out (at least for me).

gkoz: Oops, not the constructor. The method, which uses it.

src/tree_model_filter.rs
gkoz: There are other places to store unfinished code.

GuillaumeGomez: Where?

src/tree_model_filter.rs
GuillaumeGomez: Well, still the same for me. :-/

src/tree_model_filter.rs
gkoz: In your local branch, in your repo, etc.

gkoz:
I'm pretty sure you have't regenerated the code after editing `Gir.toml`. This demonstrates why it's beneficial to have a separate "Regen" commit: it's easier (i.e. less error-prone) to discard and recreate it than to amend other commits. Another concern is in case I had my PR ready you could get merge conflicts in the `src/auto` directory. Again instead of dealing with merge conflicts you could just drop that commit.

GuillaumeGomez:
Regenerated as well.

EPashkin:
`TreeModelExt::filter_new` still generated, instead commenting it in `gir.toml` need change it to
```
    [[object.function]]
    name = "filter_new"
    # Replaced by TreeModelFilter::new
    ignore = true
```

GuillaumeGomez:
@EPashkin: Indeed, regenerated. Thanks.

EPashkin:
Changing `gir.toml` better do in first commit.

EPashkin:
Not deleted `filter_new` in manual tree model so build failed.

EPashkin:
Before you start fix it, plz, do as @gkoz suggested: just create branch like `tree_model_filter_tmp` with current code, and don't push it to server,
then remove commented parts from `src/tree_model_filter.rs`.
And don't add any `src/auto/*` changes to first commit.

GuillaumeGomez:
Done. I splitted out all non-gir generated changes to the first commit.

EPashkin:
Thanks. Seems all ok. Except needed some `unwrap` in examples.

gkoz:
I don't understand the decision to regenerate with a newer (387ece8) gir than the one pinned by the submodule (0d8699e). It introduces changes that aren't in scope for this PR and some of them need to be overridden in `Gir.toml`.

EPashkin:
Then I add examples PR, that need for this PR to builds.

GuillaumeGomez:
@EPashkin: Thanks! :)
@gkoz: I didn't have a copy of gir on my computer so I redownloaded the repo and didn't think to "reset --hard".

gkoz:
I'm preparing a separate PR for gir update, which will deal with the `Option`, so this can be generated with the currently pinned version and ignore the `Option` issue.

EPashkin:
Oh, then https://github.com/gtk-rs/examples/issues/80 was wrong idea?

gkoz:
A bit too early perhaps but it's all mostly done: https://github.com/gtk-rs/gir/pull/232

EPashkin:
@GuillaumeGomez you don't will reRegen with version 0d8699e ? or don't have time? If any problem we can talk about it on gitter today.

GuillaumeGomez:
@EPashkin: I didn't know it was done, I'll do it once I'm home.

EPashkin:
Seems problem bigger that I though: before reregen need remove `Option` in manual `TreeModelExt::get_path` to build `gtk` but after this build examples still failing by my 
premature PR.
We have 2 options:
1. Merge this after fixing with ignore build errors in examples.
2. Merge @gkoz 's PR first and rebase this to it, but it need some work on merging changes in `Gir.toml` and other conflicts.

GuillaumeGomez:
The first solution seems more easy to do.

EPashkin:
IMHO it same complexity as required some git'fu
Though second have more work for gkoz

gkoz:
That PR already passes the test, here it seem you just need to discard the `get_path` changes.

EPashkin:
After merge your?

EPashkin:
Now we have only second options: rebase this on new master :wink:

GuillaumeGomez:
Updated.

gkoz:
Looks good, thanks!

Issue #278, opened 2016-03-15
Fix "zero-sized fn" nightly warnings


gkoz:
Example:
```
src/signal.rs:93:70: 93:79 warning: `extern "C" fn(&core::cell::RefCell<Box<core::ops::FnMut() -> glib::source::Continue + 'static>>) -> i32 {signal::trampoline}` is now zero-sized and has to be cast to a pointer before transmuting to `core::option::Option<unsafe extern "C" fn(*mut libc::c_void) -> i32>`, #[warn(transmute_from_fn_item_types)] on by default

src/signal.rs:93         glib_ffi::g_idle_add_full(glib_ffi::G_PRIORITY_DEFAULT_IDLE, transmute(trampoline),

                                                                                      ^~~~~~~~~

src/signal.rs:93:70: 93:79 warning: this was previously accepted by the compiler but is being phased out; it will become a hard error in a future release!

src/signal.rs:93:70: 93:79 note: for more information, see issue #19925 <https://github.com/rust-lang/rust/issues/19925>
```

We should make our extern fns match the expected signatures to make transmutes unnecessary.

GuillaumeGomez:
Can't we just cast as `*mut c_void`? It would do the same after all.

gkoz:
Transmuting `fn` is too fragile anyway. More error-prone than transmuting or casting individual arguments.

gkoz:
Um, well, speaking of signals there may not be actual signatures to adhere to. I'm not sure now. So this issue needs some investigation.

EPashkin:
https://github.com/rust-lang/rust/issues/19925 stated that we can do even `transmute(size_changed_trampoline as usize)`.
As for @gkoz concern about transmuting individual arguments IMHO it not applied here, except for `&Box<Fn>` argument that really not error-prone.
May be better solution will be pass some structure with some id and function pointer that will be checked before using function but it needed some storage with unclear rule for erasing elements.
As we need transmute trampoline anyway IMHO better add cast sooner.

gkoz:
It looks like we'll have to use the `as usize` workaround for signal trampolines because there's no other choice there. Callbacks with defined types (like `GtkTreeModelFilterVisibleFunc`) and `g_idle_add` can avoid transmuting `fn`s.

EPashkin:
I add PR for gtk and glib then.

EPashkin:
Also calling  `g_idle_add_full` and other in glib also needed casting.

gkoz:
[...]

PR #281, opened 2016-03-17
fixed some clippy warnings

llogiq:
Mostly this changes `_.map(..).unwrap_or(_)`  to `_.map_or(_, ..)` in a few places.

There are a *lot* of warnings suggesting adding `Default` impls for most types (which may or may not be useful, I have not looked too closely into the codebase), also unrelated to clippy, there are a lot of transmute warnings on nightly, but I'm unsure if this is in generated code.

### DIFF
--- a/src/list_store.rs
+++ b/src/list_store.rs
@@ -37,7 +37,7 @@ impl ListStore {
             let mut iter = TreeIter::uninitialized();
             ffi::gtk_list_store_insert_with_valuesv(self.to_glib_none().0,
                 iter.to_glib_none_mut().0,
-                position.map(|n| n as c_int).unwrap_or(-1),
+                position.map_or(-1, |n| n as c_int),
                 mut_override(columns.as_ptr() as *const c_int),
                 values.to_glib_none().0,
                 columns.len() as c_int);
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Thanks for your work!

> There are a lot of warnings suggesting adding Default impls for most types (which may or may not be useful, I have not looked too closely into the codebase)

A lot of structs need arguments for their initialization so it'd be better to just avoid implementing `Default` when it's not needed.

> also unrelated to clippy, there are a lot of transmute warnings on nightly, but I'm unsure if this is in generated code.

This is in generated code. @gkoz and I are working on it.

@gkoz: PR is good for me.

llogiq:
Regarding `Default` impls, clippy only warns when there's a constructor method without arguments (this could reasonably be a default). So this doesn't target structs that need arguments. However, the question whether implementing `Default` makes sense hinges on the question if those arg-less constructors really return meaningful values.

gkoz:
It's difficult to comment on `Default` without specific examples. I'm not sure I see benefits of implementing `Default` for any widgets; note that they would have a [potential to panic](https://github.com/gtk-rs/gtk/blob/master/src/auto/label.rs#L24) which doesn't seem like proper `Default` behavior.

The transmute warnings are in manual code. The relevant issue is #278.

Thanks for pointing out `map_or`.

llogiq:
You're welcome. :+1:

PR #284, opened 2016-03-19
Cast trampolines before transmuting

EPashkin:
Fix warning introduced https://github.com/rust-lang/rust/issues/19925
See https://github.com/gtk-rs/gtk/issues/278

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -90,7 +90,7 @@ pub fn idle_add<F>(func: F) -> u32
     where F: FnMut() -> Continue + 'static {
     assert_initialized_main_thread!();
     unsafe {
-        glib_ffi::g_idle_add_full(glib_ffi::G_PRIORITY_DEFAULT_IDLE, transmute(trampoline),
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Good too. Thanks @EPashkin!

src/signal.rs
```diff
@@ -90,7 +90,7 @@ pub fn idle_add<F>(func: F) -> u32
     where F: FnMut() -> Continue + 'static {
     assert_initialized_main_thread!();
     unsafe {
-        glib_ffi::g_idle_add_full(glib_ffi::G_PRIORITY_DEFAULT_IDLE, transmute(trampoline),
+        glib_ffi::g_idle_add_full(glib_ffi::G_PRIORITY_DEFAULT_IDLE, transmute(trampoline as usize),
```
gkoz: We can apply to these the same approach as in `glib` right?

EPashkin:
Updated 3 first functions, but others still needed `transmute` as `connect` accept only `GCallback` (function without parameters)

gkoz:
Thanks!

PR #288, opened 2016-03-28
Make some return types nullable, add StyleProvider

gkoz:


EPashkin:
Strange still error
```
src/treeview.rs:112:20: 112:29 error: no method named `get_depth` found for type `core::option::Option<gtk::auto::tree_path::TreePath>` in the current scope
```

EPashkin:
It need rebase after 279 merge and then regen again to hide `TreeModel::filter_new`.

PR #289, opened 2016-03-31
ComboBox::format-entry-text  and Scale::format-value signals

EPashkin:
Tried update examples for this (https://github.com/epashkin/examples/tree/connect_format_value),
but it crashes, so there some error in this code.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1814,3 +1814,32 @@ mod status_icon {
         f(&from_glib_none(this), size).to_glib()
     }
 }
+
+pub trait ScaleSignals {
+    fn connect_format_value<F: Fn(&Self, f64) -> String + 'static>(&self, f: F) -> u64;
+}
+
+mod scale {
+    use Scale;
+    use libc::{c_double, c_char};
+    use std::mem::transmute;
+    use ffi::GtkScale;
+    use glib::signal::connect;
+    use glib::translate::*;
+
+    impl super::ScaleSignals for Scale {
+        fn connect_format_value<F: Fn(&Self, f64) -> String + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(&Self, f64) -> String + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0, "format-value",
+                    transmute(format_value as usize), Box::into_raw(f) as *mut _)
+            }
+        }
+    }
+
+    unsafe extern "C" fn format_value(this: *mut GtkScale, value: c_double,
+                                      f: &Box<Fn(&Scale, f64) -> String + 'static>) -> *mut c_char {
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -1796,3 +1812,31 @@ mod status_icon {
...
+    use libc::{c_double, c_char};
+    use std::mem::transmute;
+    use ffi::GtkScale;
+    use glib::signal::connect;
+    use glib::translate::*;
+
+    impl super::ScaleSignals for Scale {
+        fn connect_format_value<F: Fn(&Self, f64) -> String + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<F>> = Box::new(Box::new(f));
```
gkoz: The trouble is in this line: the inner box needs to be a trait object `Box<Fn(&Scale, f64) -> String + 'static>`, `Box<F>` is just a boxed struct.

EPashkin:
Updated.

EPashkin:
"`Box<F>` is just a boxed struct" =>  "`Box<F>` is just a boxed function"?
It works, thanks.
Add PR to examples after this (already updated branch).

gkoz:
> "Box<F> is just a boxed struct" => "Box<F> is just a boxed function"?

I suppose, `F` is either a function or a struct, since a closure is an anonymous struct, that stores the captured environment. Apparently, it can be zero-sized if nothing is captured. As we've recently discovered, on nightly functions are zero-sized too. [Play](http://is.gd/88Aneq).

Boxing a zero-sized value yields a special invalid pointer value (which appears to be `0x1` at this time).

gkoz:
r+ from me

GuillaumeGomez:
@EPashkin: Before we merge, can you try to get lines to less than 100 column please?

EPashkin:
Fixed

src/signal.rs
```diff
@@ -1058,6 +1068,15 @@ mod combobox {
         callback_guard!();
         f(&ComboBox::from_glib_none(this).downcast_unchecked(), scroll_type);
     }
+
+    unsafe extern "C" fn format_entry_text_trampoline<T>(
+        this: *mut GtkComboBox,
+        path: *mut c_char, f: &Box<Fn(&T, &str) -> String + 'static>) -> *mut c_char
+    where T: IsA<ComboBox> {
```
GuillaumeGomez: I never remember how @gkoz wants where indentation. Let's wait for his advice. :)

gkoz: `where` is all right but the way parameters are wrapped seems a bit random: the first one is one per line and the rest are not. I don't think we do it this way elsewhere in this module.

EPashkin:
Updated.

GuillaumeGomez:
Thanks! (waiting for travis at least)

PR #294, opened 2016-04-07
Bump versions

gkoz:


EPashkin:
Don't want before this remove `IsA<Buildable>` constraint in `Builder::get_object` ?

EPashkin:
Or its just forgotten bump after 0.7 publishing?

gkoz:
The goal is to make the version different from crates.io to avoid that kind of confusion, not to follow semver in the master branch (that wouldn't work too well anyway).

gkoz:
The examples build in travis hits the limitation of cargo path overrides but the gtk part seems to build cleanly.

EPashkin:
IMHO no way update gtk system versions without broken builds, so why just not merge all modules and update `examples` last for real check?

gkoz:
I'll take the liberty to merge this because the rest of PRs were approved and keeping this one open will just break someone's build.

GuillaumeGomez:
Absolutely, you did well. Sorry, I didn't see it... :-/

PR #295, opened 2016-04-07
Remove the IsA<Buildable> bound on Builder::get_object

gkoz:


gkoz:
BTW, since `Buildable` isn't a useful bound and is only relevant when implementing subclasses, it seems we should remove it.

EPashkin:
Agree than `Buildable` mainly for "internal" use and can disabled

PR #296, opened 2016-04-07
Panic if libgtk-3 appears to be built with enable-debug=no

### LINKED ISSUE #270: Dealing with G_DISABLE_CHECKS
GTK+ and other GLib-based libraries [rely](https://github.com/GNOME/gtk/search?q=g_return_if_fail&type=Code) [heavily](https://github.com/GNOME/gtk/search?q=g_return_val_if_fail&type=Code) on [g_return_if_fail](https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-if-fail)/[g-return-val-if-fail](https://developer.gnome.org/glib/stable/glib-Warnings-and-Assertions.html#g-return-val-if-fail) checks for safety.

> ```C
#define             g_return_if_fail(expr)
```

> Verifies that the expression `expr` , usually representing a precondition, evaluates to `TRUE`. If the function returns a value, use `g_return_val_if_fail()` instead.

These shunts guard against undefined behavior.

> If `expr` evaluates to `FALSE`, the current function should be considered to have undefined behaviour (a programmer error). The only correct solution to such an error is to change the module that is calling the current function, so that it avoids this incorrect call.

> To make this undefined behaviour visible, if `expr` evaluates to `FALSE`, the result is usually that a critical message is logged and the current function returns.

But they're optional.

> If `G_DISABLE_CHECKS` is defined then the check is not performed.

If `G_DISABLE_CHECKS` is set, this crate becomes unable to uphold the safety guarantees. We can't control that option because for technical and licensing reasons `gtk-sys` links with `libgtk-3` dynamically. Even detecting it in a straightforward way doesn't seem possible.

Although the majority of those checks (`NULL` and object type checks) are superseded by Rust's type system, a significant number (potentially hundreds) remain necessary. Duplicating them all, beside being another handful of tedious error-prone work, would mean targeting not a documented interface but an implementation.

Ideas?

gkoz:
Safety of this crate's API relies on `g_return_if_fail` checks being enabled. Refuse to work with library built with `enable-debug=no`, which exposes undefined behavior all over the place.
Kudos to @efyang for discovering ways to implement this check.
Close #270.

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -3,9 +3,10 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use std::cell::Cell;
+use std::env;
 use std::ptr;
 use std::sync::atomic::{AtomicBool, ATOMIC_BOOL_INIT, Ordering};
-use libc::c_uint;
+use libc::{c_char, c_int, c_uint};
 use ffi;
 use glib::translate::*;
 use gdk;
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Oh nice! So you finally found out how to do it. :)

gkoz:
Yeah, we figured this out in the issue, I just kept putting off implementing it.

PR #300, opened 2016-04-08
Add signals for MenuItem

### LINKED ISSUE #297: GtkMenuItem doesn't seem to provide .connect_activate()
There doesn't seem to be a way to implement a 'click' event for MenuItems, from what I can tell?

```
src/hymntoninkasi.rs:350:19: 350:35 error: no method named `connect_activate` found for type `gtk::MenuItem` in the current scope
src/hymntoninkasi.rs:350         unitsitem.connect_activate ( move |_| {
                                           ^~~~~~~~~~~~~~~~
src/hymntoninkasi.rs:350:19: 350:35 note: the method `connect_activate` exists but the following trait bounds were not satisfied: `gtk::MenuItem : glib::IsA<gtk::Entry>`, `gtk::MenuItem : glib::IsA<gtk::Button>`
```

Edit: Sorry I just noticed there might already be an issue open for this.

cheers

Chris

GuillaumeGomez:
Fixes #297.

cc @gkoz

### DIFF
--- a/src/dialog.rs
+++ b/src/dialog.rs
@@ -28,8 +28,7 @@ impl Dialog {
         ret.add_buttons(buttons);
         ret
     }
-
-    }
+}
 
 pub trait DialogExt {
     // -- manual --
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -1843,3 +1843,64 @@ mod scale {
         f(&from_glib_none(this), value).to_glib_full()
     }
 }
+
+pub trait MenuItemSignals {
+    fn connect_activate<F>(&self, activate_func: F) -> u64
+        where F: Fn(&Self) + 'static;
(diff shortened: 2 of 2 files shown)


gkoz:
Oh that's a good catch by @EPashkin: `toggle-size-request` takes a pointer, not an `int`. Should we perhaps discard both `toggle-*` signals because they're undocumented and probably aren't supposed to be used by applications?

GuillaumeGomez:
They're documented actually ([here](https://developer.gnome.org/gtk3/stable/GtkMenuItem.html#GtkMenuItem-toggle-size-allocate)).

gkoz:
The machine has done its job yes but the meaning of the signals and their arguments is completely opaque. Which suggests they're rather just virtual methods / implementation details.

PR #301, opened 2016-04-09
Signals

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -29,14 +29,18 @@ use gdk::{
 use cairo::{Context, RectangleInt};
 
 use {
+    CellEditable,
+    CellRenderer,
     Continue,
     DeleteType,
     DirectionType,
     MovementStep,
+    Rectangle,
     ScrollType,
     StateFlags,
     TextDirection,
     TreeIter,
+    TreeModel,
     TreePath,
     TreeSelection,
     TreeViewColumn,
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -1903,4 +1928,170 @@ mod menu_item {
...
+            }
+        }
+    }
+
+    unsafe extern "C" fn trampoline<T>(this: *mut GtkAboutDialog, c_str: *const c_char,
+                                       f: &Box<Fn(&T, &str) + 'static>)
+    where T: IsA<AboutDialog> {
+        callback_guard!();
+        let buf = CStr::from_ptr(c_str).to_bytes();
+        let string = str::from_utf8(buf).unwrap();
```
gkoz: This seems to reimplement `CStr::to_str`.

src/signal.rs
```diff
@@ -1903,4 +1928,170 @@ mod menu_item {
...
+    unsafe extern "C" fn apply_trampoline<T>(this: *mut GtkCellArea,
+                                             model: *mut GtkTreeModel,
+                                             iter: *mut GtkTreeIter,
+                                             is_expander: gboolean,
+                                             is_expanded: gboolean,
+                                             f: &Box<Fn(&T, &TreeModel, &TreeIter, bool, bool) + 'static>)
+    where T: IsA<CellArea> {
+        callback_guard!();
+        f(&CellArea::from_glib_none(this).downcast_unchecked(),
+          &TreeModel::from_glib_none(model).downcast_unchecked(),
```
gkoz: We're not generic over `TreeModel` so this downcast isn't doing anything.

src/signal.rs
```diff
@@ -1903,4 +1928,170 @@ mod menu_item {
...
+          &TreeIter::from_glib_none(iter),
```
gkoz: From the existing trampolines in this module it seems that `TreeIter` should be converted with `from_glib_borrow`.

src/signal.rs
EPashkin: Also for not generic also works without `TreeModel::`

src/signal.rs
```diff
@@ -1903,4 +1928,170 @@ mod menu_item {
...
+
+    unsafe extern "C" fn remove_trampoline<T>(this: *mut GtkCellArea,
+                                              renderer: *mut GtkCellRenderer,
+                                              editable: *mut GtkCellEditable,
+                                              f: &Box<Fn(&T, &CellRenderer, &CellEditable) + 'static>)
+    where T: IsA<CellArea> {
+        callback_guard!();
+        f(&CellArea::from_glib_none(this).downcast_unchecked(),
+          &CellRenderer::from_glib_none(renderer).downcast_unchecked(),
+          &CellEditable::from_glib_none(editable).downcast_unchecked());
```
gkoz: Just generally it only makes sense to downcast `this` because the rest of arguments  aren't specializable.

src/signal.rs
GuillaumeGomez: I just copy/paste some code above in this file.

GuillaumeGomez:
I removed unneeded downcasts.

src/signal.rs
gkoz: This comment still applies.

GuillaumeGomez: Oh indeed, fixed.

src/signal.rs
```diff
@@ -1903,4 +1928,170 @@ mod menu_item {
...
+    use glib::signal::connect;
+    use glib::translate::*;
+    use IsA;
+    use libc::c_char;
+    use std::ffi::CStr;
+    use std::str;
+
+    impl super::CellAreaSignals for CellArea {
+        fn connect_add_editable<F>(&self, add_editable_func: F) -> u64
+        where F: Fn(&Self, &CellRenderer, &CellEditable, &Rectangle, &str) + 'static {
```
gkoz: Meh, I've missed the fact that the string parameter is a path. [Here's how we handle that](https://github.com/gtk-rs/gtk/blob/c40bc7abe7f922957d59c1df360d55f8e5687996/src/signal.rs#L1150-L1165). Applies to `focus-changed` as well.

gkoz: Actually since we're not borrowing the path it's probably better to pass it by value not by reference like in that example.

GuillaumeGomez:
Updated.

gkoz:
Thanks!

PR #305, opened 2016-04-16
Signals

GuillaumeGomez:
cc @gkoz

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -2305,3 +2305,37 @@ mod cell_renderer {
           path);
     }
 }
+
+pub trait CellRendererTextSignals {
+    fn connect_edited<F: Fn(&Self, &TreePath, &str) + 'static>(&self, f: F) -> u64;
+}
+
+mod cell_renderer_text {
+    use std::mem::transmute;
+    use std::str;
+    use glib::signal::connect;
+    use glib::translate::*;
+    use libc::c_char;
+    use std::ffi::CStr;
+    use ffi::{GtkCellRendererText, gtk_tree_path_new_from_string};
+    use {CellRendererText, TreePath};
+
+    impl super::CellRendererTextSignals for CellRendererText {
+        fn connect_edited<F: Fn(&Self, &TreePath, &str) + 'static>(&self, f: F) -> u64 {
+            unsafe {
+                let f: Box<Box<Fn(&Self, &TreePath, &str) + 'static>> = Box::new(Box::new(f));
+                connect(self.to_glib_none().0, "edited",
+                    transmute(trampoline as usize), Box::into_raw(f) as *mut _)
+            }
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -2305,3 +2306,154 @@ mod cell_renderer {
...
+
+mod container {
+    use std::mem::transmute;
+    use glib::signal::connect;
+    use glib::translate::*;
+    use ffi::{GtkContainer, GtkWidget};
+    use {Container, Widget};
+
+    impl super::ContainerSignals for Container {
+        fn connect_add<F: Fn(&Self, &Widget) + 'static>(&self, f: F) -> u64 {
```
GuillaumeGomez: I'm not sure here if `Widget` shouldn't be `IsA<Widget>`.

EPashkin:
https://developer.gnome.org/gtk3/stable/GtkCellRendererText.html ,
https://developer.gnome.org/gtk3/stable/GtkCheckMenuItem.html ,
https://developer.gnome.org/gtk3/stable/GtkContainer.html ,
have children, so `<T>`

EPashkin:
`GtkContainer` signals can be generated https://gist.github.com/EPashkin/977773a192bf633bb8413a96a1930870, but currently don't contains `IsWidget<T>` (IMHO needed)

EPashkin:
..(forgot add main idea), so if `IsA<Widget>` not needed and no need in this signals currently, IMHO not need add signals as it simply.

gkoz:
`IsA<Widget>` couldn't work: the handler can't assume anything about the types of the child widgets, we can't predict anything statically.
More generally, I suspect that the container signals are not really supposed to be used by applications. Note the absence of documentation. I haven't tested this but I have the impression that `Box` doesn't emit any of `Container` signals (e.g. packing a widget might not emit the `add` signal).

GuillaumeGomez:
@gkoz: Is it really an issue? At least we'll have them *if* anyone needs them.

@EPashkin: I updated the PR, do you see anything else? (I made `CellRendererText`, `CheckMenuItem` and `Container` "serializable".)

EPashkin:
Thanks, sees it all ok.

gkoz:
> Is it really an issue? At least we'll have them if anyone needs them.

If someone needs them but they don't match reasonable expectations it is an issue. Can you test with a few containers to see if `add` and `remove` are fired as expected?

Regarding someone needing any obscure signal, I think it might be possible to make a dynamically-typed interface for connecting any signals.

GuillaumeGomez:
@gkoz: It's working (tested on VerticalLayout in process-viewer [here](https://github.com/GuillaumeGomez/process-viewer/tree/test_signal)). :)

GuillaumeGomez:
So, can we merge it?

gkoz:
Thanks for testing these signals @GuillaumeGomez

Issue #307, opened 2016-04-17
Applying new enums and bitflags to gtk-rs modules.


EPashkin:
https://github.com/gtk-rs/gir/pull/229 added generation of enums and bitflags and changes in code that assumes that all of it has `.to_glib(_xxx)` and `from_glib(_xxx)` functions.
For this need add changes to all `gtk-rs` modules so I create this issue for tracking it progress.

- [x] gtk -> 310
- [x] gdk -> 115

/*Currently no change needed in other modules
- [ ] pango
- [ ] cairo ?
- [ ] glib
- [ ] gio ?
- [ ] examples

*/
Also I propose don't merge to any PR that unrelated to this issue until it was fixed.

EDIT: Goal this issue is get buildable `gtk` with generated enums,
we were lucky and it needed only changes in `gdk`,
so after 155 merged I update 310 and it must pass build.

EPashkin:
@gkoz You doing changes yourself or want help?

gkoz:
We should probably start from `gio` and gradually go up to `gtk` along the dependency tree, converting other crates to generated mode. I'll put up the gio crate and then I don't mind someone else trying to tinker with any of the crates.

EPashkin:
Strange, but needed only changes in `gtk` and `gdk`, so I add PRs with auto regens.

EPashkin:
Updated issue.

EPashkin:
Now `gtk` updated and builds (not on travis) so I close this

PR #310, opened 2016-04-18
Enums

EPashkin:
Part of #307
Do not merge until all parts done


EPashkin:
Build of this PR IMHO better restarted after merging `gdk` part, it must builds.

gkoz:
A big question here is are there any enums that appear as return values, that we don't want to have the `__Nonexhaustive` member, i.e. keep exhaustive matches possible for them?

The error enums to add can be found by the `glib:error-domain` attribute. We'll at least want `BuilderError`, `FileChooserError`, `RecentChooserError`.

EPashkin:
Agree that `__Nonexhaustive` unintuitive. Also on detailed look I don't understand what return enum's `from_glib` if ffi-value not in enum.

I add gtk's errors after gdk's PR merged to test build.

EPashkin:
Checked that adding all 7 errors don't fail build.

EPashkin:
Now I fully don't understand reason add `__Nonexhaustive` to enums as it checked only in `to_glib` and 
"defend" C++ code from rust.

gkoz:
It's true that `from_glib` doesn't handle unexpected variants *now*, because enums in the sys crates are incapable of representing unexpected values. Once we finish this transition and stop reexporting enums from sys, they can be changed to `i32` typedefs or newtypes and then `from_glib` conversions can have the fallback arms.

gkoz:
Shouldn't we remove all enum and flags reexports from `lib.rs`? Also this seems to have build errors unrelated to the gdk PR.

EPashkin:
Strange, I see that all 15 errors from gdk.

gkoz:
Might need to enable the `v3_16` feature to reproduce.

EPashkin:
Though I misses `gtk` enums needed with `gtk_3_18` feature

EPashkin:
@gkoz I ready to push fixed.
Is there any problem with gdk's PR? IMHO better merge it before push here, so we get normal build test.

gkoz:
Fine, let's do this.

EPashkin:
What to do with ./check_init_asserts errors?

gkoz:
Oops, I wondered why this isn't an issue but it actually is... The generator should be inserting the 'skip' macro invocations.

gkoz:
We could merge and fix that later.

GuillaumeGomez:
@gkoz: I let you handle this PR (and this issue as well :p).

EPashkin:
AppVeyor build succeeded

gkoz:
Thanks @EPashkin for handling this. I'd rather we removed other reexports right away if they seem unused.

EPashkin:
IMHO I removed all enums reexports, missed anything?

gkoz:
Rigth, sorry, this only applies to gdk then.

A concern here is preserving a way to get a `i32` from `IconSize` until we override the type. We might recommend using `to_glib` for the time being...

PR #314, opened 2016-04-20
Implement Into<i32> for IconSize and ResponseType

gkoz:
Preserve a way to pass these to functions that take `i32` until the generator supports rewriting types.

### DIFF
--- a/src/enums.rs
+++ b/src/enums.rs
@@ -1,0 +1,15 @@
+use glib::translate::ToGlib;
+
+impl Into<i32> for ::IconSize {
+    fn into(self) -> i32 {
+        skip_assert_initialized!();
+        self.to_glib() as i32
+    }
+}
+
+impl Into<i32> for ::ResponseType {
+    fn into(self) -> i32 {
+        skip_assert_initialized!();
+        self.to_glib() as i32
+    }
+}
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Don't we have a trait generator for enums now?

gkoz:
This is a temporary band-aid, normally there's no need to convert enums to `i32`. Eventually functions like `DialogExt::add_button` should take an enum instead of `i32` but it's not practical yet.

GuillaumeGomez:
Oh ok. I let you handle with travis errors.

gkoz:
D'oh, I forgot the macro here as well!

GuillaumeGomez:
There are now a lot of errors. I don't tjink they come from this PR however.

EPashkin:
If from this PR in gtk, just log not good.

EPashkin:
Must be `skip_assert_initialized!();`

EPashkin:
+1 after fix

EPashkin:
@GuillaumeGomez it seems fixed

GuillaumeGomez:
Perfect! Thanks a lot @EPashkin!

EPashkin:
Not me but gkoz need thanking, he just forgot to say, that it fixed

GuillaumeGomez:
Haha, indeed! Sorry for the confusion.

So thanks @EPashkin for notifying me and thanks a lot @gkoz for your work! :)

PR #318, opened 2016-04-23
Add missing signals for SpinButton

GuillaumeGomez:
cc @EPashkin


src/signal.rs
```diff
@@ -1706,34 +1707,111 @@ mod editable {
...
+                let f: Box<Box<Fn(&SpinButton, ScrollType) + 'static>> =
+                    Box::new(Box::new(change_value_func));
+                connect(self.to_glib_none().0, "change_value",
+                        transmute(change_trampoline as usize), Box::into_raw(f) as *mut _)
+            }
+        }
 
-    impl SpinButton {
-        pub fn connect_input<F: Fn(&SpinButton, &mut f64) -> Result<bool, ()> + 'static>(&self, f: F) -> u64 {
+        fn connect_input<F>(&self, f: F) -> u64
```
GuillaumeGomez: Could you take a look here @gkoz?

src/signal.rs
```diff
@@ -1706,34 +1707,111 @@ mod editable {
...
+
+    unsafe extern "C" fn change_trampoline(this: *mut GtkSpinButton,
+                                           scroll: ScrollType,
+                                           f: &Box<Fn(&SpinButton, ScrollType) + 'static>) {
+        callback_guard!();
+        f(&from_glib_none(this), scroll)
     }
 
-    unsafe extern "C" fn input_trampoline(this: *mut GtkSpinButton, new_value: *mut c_double, f: gpointer) -> c_int {
+    unsafe extern "C" fn input_trampoline(this: *mut GtkSpinButton,
```
GuillaumeGomez: And here too please @gkoz.

src/signal.rs
```diff
@@ -1706,34 +1707,111 @@ mod editable {
     }
 }
 
+pub trait SpinButtonSignals {
```
EPashkin: IMHO trait don't needed

GuillaumeGomez: It's just to be consistent with the other signals in this file.

EPashkin: Them may be `mod gl_area` need fixing?

GuillaumeGomez: If it's in this file, then yes. Want me to do it in this PR as well? Or I open a new one?

EPashkin: It in this file, only one without `xxSignals`.
Not sure, that it really need be fixed. IMHO better wait for gkoz.

gkoz:
Regarding the `input` signal. Should it be `Fn(&SpinButton) -> Option<Result<f64, ()>>`?
That is, first of all the signal handler has a choice of either trying to convert the value or just reacting to the signal. If it doesn't convert it `None` is returned (this corresponds to `Inhibit(false)`). If it does attempt conversion, then it's `Some(result)` (inhibiting the default handler).

EPashkin:
@gkoz Your closure variant much better.
@GuillaumeGomez Change it please, and I fix example.

GuillaumeGomez:
Updated.

src/signal.rs
```diff
@@ -1706,34 +1707,117 @@ mod editable {
...
-            Err(_) => GTK_INPUT_ERROR,
+        f(&from_glib_none(this), scroll)
+    }
+
+    unsafe extern "C" fn input_trampoline(this: *mut GtkSpinButton,
+                                          new_value: *mut c_double,
+                                          f: &Box_<Fn(&SpinButton) -> Option<Result<f64, ()>> + 'static>)
+                                          -> c_int {
+        callback_guard!();
+        if let Some(v) = f(&from_glib_none(this)) {
```
EPashkin: Not better add one  `match` ?

EPashkin:
Also needed ignoring generating signals and Regen.

src/signal.rs
GuillaumeGomez: You gave me an idea actually! Hold on!

GuillaumeGomez:
@EPashkin: In just one match now. :D

src/signal.rs
```diff
@@ -1706,34 +1707,114 @@ mod editable {
...
             unsafe {
-                let f: Box_<Box_<Fn(&SpinButton, &mut f64) -> Result<bool, ()> + 'static>> = Box_::new(Box_::new(f));
+                let f: Box_<Box_<Fn(&SpinButton) -> Option<Result<f64, ()>> + 'static>> = Box_::new(Box_::new(f));
                 connect(self.to_glib_none().0, "input",
                         transmute(input_trampoline as usize), Box_::into_raw(f) as *mut _)
             }
         }
+
+        fn connect_output<F>(&self, output_func: F) -> u64
+        where F: Fn(&SpinButton) -> bool + 'static {
```
gkoz: It looks like the return type here has to be `Inhibit`.

EPashkin:
@GuillaumeGomez It much better.

src/signal.rs
GuillaumeGomez: Then `Inhibit` it'll be.

EPashkin: Why `Inhibit` ?
```
Returns
TRUE if the value has been displayed
```

EPashkin: Due `_gtk_boolean_handled_accumulator` in code?

GuillaumeGomez: So... Do I change it or not?

gkoz: > Due `_gtk_boolean_handled_accumulator` in code?

Yes, the documentation fails to make the semantics clear sometimes but the use of this accumulator strongly suggests `Inhibit`.

GuillaumeGomez:
I regenerated and added the `Inhibit` return.

gkoz:
Out of curiosity why does the generator not generate `connect_wrapped`? The signature seems basic and I don't see any `ignore` directives.

gkoz:
Um looks like it actually does generate four signals.

EPashkin:
@GuillaumeGomez fix this please, it must be `[[object.signal]]`
```
[[object]]
name = "Gtk.StatusIcon"
status = "generate"
    [[object.function]]
    pattern = ".+" #not all generated
    ignore = true
```

EPashkin:
"Regen" again with different version of `gir` :(

GuillaumeGomez:
Updated `Gir.toml` file and regenerated.

EPashkin:
Thanks, but add "Update gir" commit please if you regen with `gir (af5277e)`

GuillaumeGomez:
I don't have the "original' regen commit, the new one is with `gir (af5277e)`.

EPashkin:
Seems we have some misunderstanding.
I asked you updating git sub-module, same as you do in 64fcc0730c8a84b239ad3ea9c210aebd62e23c35

GuillaumeGomez:
Totally misunderstood, indeed. ^^'

EPashkin:
Other misunderstanding: we need ignore generation of spinbutton's signals.

GuillaumeGomez:
All updated.

EPashkin:
Thanks. @gkoz merge this please

gkoz:
Thanks!

PR #328, opened 2016-05-03
Fix const_err warning

gkoz:
Rely on glib::translate instead of a hacky one-off

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -6,7 +6,7 @@ use std::cell::Cell;
 use std::env;
 use std::ptr;
 use std::sync::atomic::{AtomicBool, ATOMIC_BOOL_INIT, Ordering};
-use libc::{c_char, c_int, c_uint};
+use libc::{c_int, c_uint};
 use ffi;
 use glib::translate::*;
 use gdk;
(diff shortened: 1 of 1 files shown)


src/rt.rs
```diff
@@ -135,13 +135,10 @@ fn pre_init() -> bool {
...
-            b"\0" as *const u8 as *mut c_char,
-            b"--gtk-debug=misc\0" as *const u8 as *mut c_char,
-        ];
-        let mut argc = args.len() as c_int;
-        let mut argv = args.as_mut_ptr();
-        let ret = from_glib(ffi::gtk_parse_args(&mut argc, &mut argv));
+        let argv = ["", "--gtk-debug=misc"];
+        let mut argc = argv.len() as c_int;
+        let mut argv_stash = argv.to_glib_none();
+        let ret = from_glib(ffi::gtk_parse_args(&mut argc, &mut argv_stash.0));
```
EPashkin: `&mut argv.to_glib_none().0`?

EPashkin:
:+1:

src/rt.rs
gkoz: Possible but too dense magic already.

EPashkin: Ok

GuillaumeGomez:
So no magic, ok! :D

Thanks @gkoz!

PR #330, opened 2016-05-07
Add methods to Builder object

GuillaumeGomez:


src/builder.rs
```diff
@@ -57,4 +57,31 @@ impl Builder {
                 .and_then(|obj| obj.downcast().ok())
         }
     }
+
+    pub fn add_from_file(&self, file_name: &str) -> bool {
```
EPashkin: Why not `&Path`?

EPashkin: `new_from_file` IMHO need `&Path` also (or some `T: AsPath`)

GuillaumeGomez: Indeed! Well saw!

GuillaumeGomez:
Changed files path to `&Path`.

EPashkin:
IMHO to more generic `AsPath` better but :+1: anyway.
`xxx_resource` also use file path? I though it some internal path.

src/builder.rs
```diff
@@ -26,13 +28,13 @@ impl Builder {
     }
 
     #[cfg(feature = "v3_10")]
-    pub fn new_from_file(file_name: &str) -> Builder {
+    pub fn new_from_file(file_path: &Path) -> Builder {
```
gkoz: Elsewhere we use `T: AsRef<Path>`.

src/builder.rs
```diff
@@ -26,13 +28,13 @@ impl Builder {
...
         assert_initialized_main_thread!();
-        unsafe { from_glib_full(ffi::gtk_builder_new_from_file(file_name.to_glib_none().0)) }
+        unsafe { from_glib_full(ffi::gtk_builder_new_from_file(file_path.to_glib_none().0)) }
     }
 
     #[cfg(feature = "v3_10")]
-    pub fn new_from_resource(resource_path: &str) -> Builder {
+    pub fn new_from_resource(resource_path: &Path) -> Builder {
```
gkoz: The resource paths should probably remain as strings, I don't think resources can handle arbitrary encodings.

src/builder.rs
```diff
@@ -57,4 +59,31 @@ impl Builder {
                 .and_then(|obj| obj.downcast().ok())
         }
     }
+
+    pub fn add_from_file(&self, file_path: &Path) -> bool {
```
gkoz: We have means to implement proper error handling now.

GuillaumeGomez: We do?! :open_mouth:

src/builder.rs
GuillaumeGomez: I don't see why it couldn't. The description says it's a path to a resource file after all. What would be the difference with another file path?

gkoz: It's a path inside the compiled resource bundle not in the file system.

src/builder.rs
gkoz: Grep the generated code for `Error`.

EPashkin:
May be simpler just generate `Builder` and copy paste from it (if not complete generated)?

src/builder.rs
GuillaumeGomez: Ooooooh. Totally misunderstood. Then `Path` isn't appropriate at all!

EPashkin:
From https://developer.gnome.org/gtk3/stable/GtkBuilder.html no errors in it. On any error program just aborted :cry:
Also generating not work, as no `filename` type for `GtkBuilder` functions in .gir.

GuillaumeGomez:
> May be simpler just generate Builder and copy paste from it (if not complete generated)?

Not sure since we provide a very little set of functions.

> From https://developer.gnome.org/gtk3/stable/GtkBuilder.html no errors in it. On any error program just aborted :cry:
> Also generating not work, as no filename type for GtkBuilder functions in .gir.

Arf. That's problematic.

gkoz:
> From https://developer.gnome.org/gtk3/stable/GtkBuilder.html no errors in it. On any error program just aborted :cry:

The way I see it, this applies to constructors (because constructors don't fail) but the `add_*` methods do return errors.

EPashkin:
My bad, I looked only on constructor and missed errors in functions.

src/builder.rs
```diff
@@ -57,4 +61,34 @@ impl Builder {
                 .and_then(|obj| obj.downcast().ok())
         }
     }
+
+    pub fn add_from_file<T: AsRef<Path>>(&self, file_path: T) -> Result<bool, Error> {
```
EPashkin: IMHO `Result<(), Error>` here and all others

src/builder.rs
```diff
@@ -57,4 +61,34 @@ impl Builder {
...
+        unsafe {
+            let mut error = ::std::ptr::null_mut();
+            let ret = ffi::gtk_builder_add_from_file(self.to_glib_none().0,
```
EPashkin: `let _ = ` without `> 0`

gir
```diff
@@ -1 +1 @@
-Subproject commit e6cb5d03d8c165d41e93063ef4f59a3559b76169
+Subproject commit a3f05e3ff6b6917e9ee9e7e0ca7ae3d4da1999fc
```
gkoz: This seems unrelated.

GuillaumeGomez: Oh come on! This is driving me crazy! :rage:

src/builder.rs
GuillaumeGomez: As I understood it, it could fail without returning an error.

src/builder.rs
gkoz: Or simply ignore the return value, there won't be a warning.

src/builder.rs
EPashkin: I was wrong about failing, if ffi function accept error we must return it.
But we return unit `()` in good case instead `bool`.

GuillaumeGomez:
Updated.

EPashkin:
AppVeyor build succeeded, thanks @GuillaumeGomez

gkoz:
Thanks!

PR #334, opened 2016-05-09
Add manual function for gtk_css_provider_load_from_data

### LINKED ISSUE #332: Support gtk_sys::gtk_css_provider_load_from_data in gtk::CssProvider
I'd like to specify styles as a string and apply them, which isn't currently possible in the gtk module. A partial implementation is already commented out: https://github.com/gtk-rs/gtk/blob/4a602797abc45adc12d453e14e4a9fa966cceb07/src/auto/css_provider.rs#L27

EPashkin:
Closes #332


EPashkin:
@Susurrus, please,  check that it works for you

Susurrus:
Indeed it works; thanks.

Does this need to be manually-specified? Shouldn't this be auto-generated as well or is that more work? I'm not really certain how this project works completely.

EPashkin:
It stays manual until we find solution about automatic converting pair arguments `char*data, int len` in ffi declaration to single `data: &str` in rust code.
For this function also problem that `char*data` masked as `CArray[1]`

Susurrus:
Sounds reasonable. Let me know if you need more testing.

EPashkin:
We generate most parts with https://github.com/gtk-rs/gir but it can't handle all edge cases

EPashkin:
@gkoz @GuillaumeGomez What you think about this partial manual implementation?
It can stay as partial or better updated with copy of auto-generated code?

GuillaumeGomez:
Either solution is fine for me.

src/css_provider.rs
```diff
@@ -0,0 +1,22 @@
...
+use glib::translate::*;
+use std::ptr;
+use libc::{c_uchar, ssize_t};
+
+//Other functions autogenerated in src\auto\css_provider.rs
+impl CssProvider {
+    pub fn load_from_data(&self, data: &str) -> Result<(), Error> {
+        unsafe {
+            let mut error = ptr::null_mut();
+            let _ = ffi::gtk_css_provider_load_from_data(self.to_glib_none().0, data.as_ptr() as *mut c_uchar,
```
gkoz: I'd prefer to stick to `mut_override` for switching pointer mutability. E.g. `mut_override(data.as_bytes().as_ptr())`.

gkoz:
Looks good. Looks like we should enable `Gtk.CssSection` so the `parsing-error` signal can be handled.

EPashkin:
Updated

EPashkin:
AppVeyor build succeeded

GuillaumeGomez:
:+1:

src/auto/css_section.rs
```diff
@@ -0,0 +1,58 @@
...
+        }
+    }
+
+    pub fn get_end_position(&self) -> u32 {
+        unsafe {
+            ffi::gtk_css_section_get_end_position(self.to_glib_none().0)
+        }
+    }
+
+    //pub fn get_file(&self) -> /*Ignored*/Option<gio::File> {
```
gkoz: I guess we'll have to generate `Gio.File` soon because there're no alternatives sometimes.

gkoz:
Thanks!

PR #336, opened 2016-05-09
Add tests repository into test scripts

GuillaumeGomez:
cc @gkoz 
cc @EPashkin


gkoz:
Let's wait with these changes a while. I'm trying to make a tool similar to bors/homu, that would allow us to centralize all integration tests reduce CI scripts duplication and test PRs to several repos together.

GuillaumeGomez:
That'd be wonderful! Having a tool like bors to allow r+ commands could be very helpful as well.

PR #339, opened 2016-05-12
Widget signals

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/244

WIP


EPashkin:
First unsolved problem:
after enabling `size-allocate` I get error
```
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 error: the trait bound `cairo_ffi::cairo_rectangle_int_t: glib::translate::Ptr` is not satisfied [E0277]
D:/eap/rust/0/gtk\src\auto\widget.rs:2773     f(&Widget::from_glib_none(this).downcast_unchecked(), &from_glib_none(allocation))
                                                                                                     ^~~~~~~~~~~~~~
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 help: run `rustc --explain E0277` to see a detailed explanation
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 note: required by `glib::translate::from_glib_none`
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 error: the trait bound `rectangle::Rectangle: glib::translate::FromGlibPtr<cairo_ffi::cairo_rectangle_int_t>` is not satisfied [E0277]
D:/eap/rust/0/gtk\src\auto\widget.rs:2773     f(&Widget::from_glib_none(this).downcast_unchecked(), &from_glib_none(allocation))
                                                                                                     ^~~~~~~~~~~~~~
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 help: run `rustc --explain E0277` to see a detailed explanation
D:/eap/rust/0/gtk\src\auto\widget.rs:2773:60: 2773:74 note: required by `glib::translate::from_glib_none`
```

EPashkin:
Second: I separated widget functions by 2 traits: `WidgetExt` (generated), `WidgetExt2` (manual) to remove many need proxying signals.
Tried to use make `WidgetExt2` child of `WidgetExt` it work in `gtk` but `examples` don't compiles as don't seeing it.
So currently publish both traits with different names.
It be wrong with documentation.

EPashkin:
Third: signals with `Event` not generated, but seems it can be fixed in `gir`

EPashkin:
Fourth: `map-event` and `unmap-event` uses `EventAny` and it currently not defined.

EPashkin:
Updated with `Event`s

EPashkin:
For second problem:
IMHO we can publish both manual and generated in this way (not tried):
```
pub use auto::traits::WidgetExt as WidgetExtGen;
pub use widget::WidgetExt;
```
I add currently deleted functions to manual WidgetExe if we don't find solution, may be also add new mode to `Gir` to generate templates.

gkoz:
I'd very much like to avoid splitting traits... But the current way of forwarding from manual to automatic traits appears to be impossible to maintain. So I guess we'll have to do it. We'll need some more descriptive name than just `WidgetExt2` though.

EPashkin:
Renamed exported `WidgetExt`

src/prelude.rs
```diff
@@ -51,7 +51,7 @@ pub use auto::traits::ToolItemExt;
 pub use auto::traits::ToolShellExt;
 pub use auto::traits::TreeModelExt;
 //pub use auto::traits::TreeSortableExt;
-//pub use auto::traits::WidgetExt;
+pub use auto::traits::WidgetExt as WidgetExtGenerated;
```
gkoz: Ah, not quite what I was hoping for. If both traits are exported here, we can go back to just reexporting `auto::traits::*` and only reexport manual traits manually. So it would the manual one that would get renamed (I guess `WidgetExtManual`) but in its module rather than here.

src/widget.rs
```diff
@@ -4,33 +4,20 @@
...
-    StyleContext,
-    TextDirection,
     Widget,
-    Window,
 };
 
 use auto::traits::WidgetExt as Auto;
 
-pub trait WidgetExt {
+pub trait WidgetExt: Auto {
```
gkoz: ... and then the `Auto` workaround will be unnecessary too.

src/auto/entry.rs
```diff
@@ -136,7 +136,7 @@ pub trait EntryExt {
     #[cfg(feature = "v3_16")]
     fn grab_focus_without_selecting(&self);
 
-    //fn im_context_filter_keypress(&self, event: /*Ignored*/&mut gdk::EventKey) -> bool;
+    fn im_context_filter_keypress(&self, event: &mut gdk::EventKey) -> bool;
```
gkoz: Likely needs const override.

src/auto/search_entry.rs
```diff
@@ -34,10 +36,12 @@ impl SearchEntry {
         }
     }
 
-    //#[cfg(feature = "v3_16")]
-    //pub fn handle_event(&self, event: /*Unknown conversion*//*Unimplemented*/Event) -> bool {
-    //    unsafe { TODO: call ffi::gtk_search_entry_handle_event() }
-    //}
+    #[cfg(feature = "v3_16")]
+    pub fn handle_event(&self, event: &mut gdk::Event) -> bool {
```
gkoz: const override

src/auto/text_view.rs
```diff
@@ -291,9 +291,11 @@ impl TextView {
         }
     }
 
-    //pub fn im_context_filter_keypress(&self, event: /*Ignored*/&mut gdk::EventKey) -> bool {
-    //    unsafe { TODO: call ffi::gtk_text_view_im_context_filter_keypress() }
-    //}
+    pub fn im_context_filter_keypress(&self, event: &mut gdk::EventKey) -> bool {
```
gkoz: You get the idea :)

src/prelude.rs
EPashkin: Without exporting `auto::traits::WidgetExt` build fails.
Currently we can't export `auto::traits::*` due we still hide `auto::traits::TreeSortableExt`.
Or not?

gkoz: Then do the same with `DialogExt` and `TreeSortableExt`.

EPashkin: K

EPashkin:
Fixed `&mut Event`s and removed `Auto`s and added last 2 signal to `WidgetExtManual`

EPashkin:
@gkoz, help, please, with stable travis error.

EPashkin:
The Travis CI build passed. I ready for squash commits when you say

src/dialog.rs
```diff
@@ -31,85 +29,14 @@ impl Dialog {
...
-    fn set_default_response(&self, response_id: i32);
-    fn set_response_sensitive(&self, response_id: i32, setting: bool);
 }
 
-impl<O: IsA<Dialog> + IsA<Object>> DialogExt for O {
+impl<O: DialogExt> DialogExtManual for O {
     fn add_buttons(&self, buttons: &[(&str, i32)]) {
         for &(text, id) in buttons {
-            Auto::add_button(self, text, id);
+            O::add_button(self, text, id);
```
gkoz: That's an interesting bug. Does `DialogExt` instead of `O` not work either? I guess there should be a 'FIXME' mentioning that `self.add_button` doesn't work on 1.8.

src/tree_sortable.rs
```diff
@@ -51,18 +49,13 @@ impl FromGlib<i32> for SortColumn {
...
 
-pub trait TreeSortableExt {
-    // -- manual --
+pub trait TreeSortableExtManual {
     fn set_default_sort_func<F>(&self, sort_func: F)
         where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static;
     fn set_sort_func<F>(&self, sort_column_id: SortColumn, sort_func: F)
         where F: Fn(&Self, &TreeIter, &TreeIter) -> Ordering + 'static;
     fn get_sort_column_id(&self) -> Option<(SortColumn, SortType)>;
-    fn set_sort_column_id(&self, sort_column_id: SortColumn, order: SortType);
```
gkoz: This manual function was intended to replace the original one changing its signature, so the way to go here is to ignore the generated one instead of removing it here.

src/dialog.rs
EPashkin: I add it.

EPashkin:
Fixed `TreeSortable`, added comment in `Dialog`

gkoz:
Looks good! :+1:

GuillaumeGomez:
Looks good for me as well. I'm waiting for appveyor.

EPashkin:
Squashed

EPashkin:
Gratz with 700th Travis build :wink:

GuillaumeGomez:
And congrats for this PR! :wink:

PR #351, opened 2016-06-04
Insert text

### LINKED ISSUE #349: Remove 4th callback argument for TextBuffer::connect_insert_text as its redundant
Gtk provides both a character array and a length argument for the `insert_text` callback because it's necessary in C. This isn't necessary in Rust and the length argument is redundant with the 3rd argument, which already has a length. It should be removed to make it more Rusty.

Relevant function is here: https://github.com/gtk-rs/gtk/blob/d9295b9c776c1b15ec4db0a4025838cb2f92595a/src/auto/text_buffer.rs#L482

GuillaumeGomez:
Fixes #349.

cc @gkoz
cc @EPashkin

### DIFF
--- a/src/text_buffer.rs
+++ b/src/text_buffer.rs
@@ -3,8 +3,13 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use libc::{c_char, c_int};
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use std::{slice, str};
 use ffi;
 use glib::translate::*;
+use glib::signal::connect;
+use glib_ffi;
 use TextBuffer;
 use TextIter;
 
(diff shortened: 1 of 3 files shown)


src/text_buffer.rs
```diff
@@ -55,5 +59,21 @@ impl TextBuffer {
...
+}
 
+unsafe extern "C" fn insert_text_trampoline(this: *mut ffi::GtkTextBuffer,
+                                            location: *mut ffi::GtkTextIter,
+                                            text: *mut c_char,
+                                            _len: c_int,
+                                            f: glib_ffi::gpointer) {
+    callback_guard!();
+    let f: &Box_<Fn(&TextBuffer, &TextIter, &str) + 'static> = transmute(f);
+    f(&from_glib_none(this), &from_glib_none(location), &String::from_glib_none(text))
```
EPashkin: Not safer do as in `signals.rs:insert_trampoline` with `slice::from_raw_parts`?

GuillaumeGomez: I don't know, I took the generated code and c/c'ed it here.

EPashkin: Or as in `glib\src\variant.rs`:  `str::from_utf8_unchecked(slice::from_raw_parts(ptr as *const u8, len as usize))`

GuillaumeGomez: I really don't know which solution is the best one. :-/

gkoz: > `str::from_utf8_unchecked(slice::from_raw_parts(ptr as *const u8, len as usize))`

I wouldn't go with the unchecked variant but `str::from_utf8(...).unwrap()` should be okay.

GuillaumeGomez:
Updated.

gkoz:
There's a slight issue with gir version -- the one used to regenerate doesn't seem to match the submodule. IMO updating gir is not necessary yet because the output remains the same.

EPashkin:
:+1: after gir version fix

GuillaumeGomez:
I don't know, I just took the version on the repo. I'll update.

GuillaumeGomez:
Updated.

EPashkin:
Thanks

PR #354, opened 2016-06-06
Add popup method for Menu

### LINKED ISSUE #353: Menu popup
Hello, 

It seems like [gtk_menu_popup](https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-popup) is missing, how to emulate it? Basically I'd like create a right click pop up menu.

GuillaumeGomez:
Fixes #353.

cc @gkoz
cc @EPashkin

### DIFF
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -1,0 +1,45 @@
+// Copyright 2016, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use ffi;
+use glib_ffi;
+use glib::object::Downcast;
+use glib::translate::*;
+use Menu;
+use IsA;
+use Widget;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use libc::c_int;
+use std::ptr;
+
+impl Menu {
+    pub fn popup<T: IsA<Widget>, U: IsA<Widget>,
+                 F: Fn(&Menu, &mut i32, &mut i32) -> bool + 'static>(
+        &self, parent_menu_shell: Option<&T>, parent_menu_item: Option<&U>, f: F,
+        button: u32, activate_time: u32) {
+        unsafe {
+            let f: Box_<Box_<Fn(&Menu, &mut i32, &mut i32) -> bool + 'static>> = Box_::new(Box_::new(f));
+            ffi::gtk_menu_popup(self.to_glib_none().0, parent_menu_shell.to_glib_none().0,
+                                parent_menu_item.to_glib_none().0,
+                                Some(position_callback),
+                                Box_::into_raw(f) as *mut _, button, activate_time)
+        }
+    }
+
+    pub fn popup_easy(&self, button: u32, activate_time: u32) {
+        unsafe {
+            ffi::gtk_menu_popup(self.to_glib_none().0, ptr::null_mut(),
+                                ptr::null_mut(), None, ptr::null_mut(),
+                                button, activate_time)
+        }
+    }
+}
+
+unsafe extern "C" fn position_callback(this: *mut ffi::GtkMenu, x: *mut c_int, y: *mut c_int,
+                                       push_in: *mut glib_ffi::gboolean, f: glib_ffi::gpointer) {
+    let f: &Box_<Fn(&Menu, &mut i32, &mut i32) -> bool + 'static> = transmute(f);
+    *push_in = f(&Menu::from_glib_none(this).downcast_unchecked(), x.as_mut().unwrap(),
+                 y.as_mut().unwrap()).to_glib();
+}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -200,6 +200,7 @@ mod entry_buffer;
 mod enums;
 mod file_chooser_dialog;
 mod list_store;
+mod menu;
 mod message_dialog;
 mod notebook;
 mod recent_chooser_dialog;


src/menu.rs
```diff
@@ -0,0 +1,39 @@
...
+use Menu;
+use IsA;
+use Widget;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use libc::c_int;
+
+impl Menu {
+    pub fn popup<T: IsA<Widget>, F: Fn(&Self, &mut i32, &mut i32, &mut bool) + 'static>(
+        &self, parent_menu_shell: &T, parent_menu_item: &T, f: F, button: u32, activate_time: u32) {
```
GuillaumeGomez: About the button argument: I think it's supposed to be an enum (an event one as it seems) but cannot find it. One of its member is supposed to be RIGHT_CLICK but still not able to find it. Any idea?

EPashkin:
May be this: In gdk-sys\lib.rs
```
pub const GDK_BUTTON_MIDDLE: i32 = 2;
pub const GDK_BUTTON_PRIMARY: i32 = 1;
pub const GDK_BUTTON_SECONDARY: i32 = 3;
```

GuillaumeGomez:
Ah nice! Thanks @EPashkin!

GuillaumeGomez:
@EPashkin: We didn't create an enum about it?

EPashkin:
I don't find rusted enum for button.
May be create it when fixing https://github.com/gtk-rs/gdk/issues/121 as it related?

gkoz:
The number of buttons doesn't have a clear bound, you could have a mouse with 40 buttons so an enum is unlikely to be a good match. It'd be nice to put `Option` there but then would should we do with `Some(0)`?

src/menu.rs
gkoz: The `parent_*` parameters need to have independent types and they're optional. Position func's parameters appear to be output only so they should be in the closure's return value.

gkoz:
I'm having a hard time understanding why we're required to pass a callback there at all. It doesn't seem more powerful than just providing `x`, `y`, `push_in` directly.

src/menu.rs
Eilie: `f` parameter is optional as well.

gkoz:
Maybe the documentation is incorrect about the `y` parameter being an output one.

gkoz:
If so only the boolean needs to be in the return value of the closure.

GuillaumeGomez:
I adressed the comments.

EPashkin:
@gkoz 
1a. As enum will be manual anyway we can add `Int(i32)` case in it.
1b. We can add (Default case for Some(0) and check for `!=0` in `Int` conversion)

EPashkin:
2. Not sure about neediness closure and replacement it with wished values.

Eilie:
I'm trying to use your latest changes but getting following error:
```
src/main.rs:18:10: 18:15 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
src/main.rs:18     menu.popup(&None, &None, None, button, time);
```
How can I fix that, not sure what annotations it expects to see if I'm passing `None`s to it and don't care about result.

Eilie:
Hm, actually I don't have any `_` in my code, could it be https://github.com/gtk-rs/gtk/pull/354/files#diff-602ee907f460c35a3825a0600b352dc2R28 ?

GuillaumeGomez:
I don't think so. Can you show your code please?

Eilie:
@GuillaumeGomez Sure:
```rust
extern crate gtk;

use gtk::prelude::*;
use gtk::{ StatusIcon, Menu, MenuItem };

fn popup_menu(_: &StatusIcon, button: u32, time: u32) {
    println!("Clicked!");

    let menu = Menu::new();
    let exit = MenuItem::new_with_label("Exit");

    menu.attach(&exit, 0, 100, 0, 100);

    exit.show();
    menu.popup(&None, &None, None, button, time);
}

fn main() {
    if gtk::init().is_err() {
        println!("Failed to initialize GTK.");
        return;
    }

    let status_icon = StatusIcon::new_from_file("./icon.jpeg");

    status_icon.connect_popup_menu(popup_menu);

    status_icon.set_visible(true);

    gtk::main();
}
```

GuillaumeGomez:
I don't see any issue in your code. I'll test locally (I finish my servo PR first).

EPashkin:
@Eilie, error about one of `None`(or both), you need add type `None::<Type>`

Eilie:
@EPashkin But what `Type` should I use if it is `None` and doesn't contain anything?

GuillaumeGomez:
@EPashkin: That's what I thought about, but not very beautiful :-/ Isn't there is a workaround for that?

GuillaumeGomez:
@Eilie: Just whatver you want, it doesn't matter Put `None::<Widget>` and it should be good.

Eilie:
@GuillaumeGomez @EPashkin Hm, still getting this error:
```
src/main.rs:18:10: 18:15 error: unable to infer enough type information about `_`; type annotations or generic parameter binding required [E0282]
src/main.rs:18     menu.popup(&None::<gtk::Widget>, &None::<gtk::Widget>, None, button, time);
                        ^~~~~
```

src/menu.rs
```diff
@@ -0,0 +1,46 @@
...
+use IsA;
+use Widget;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use libc::c_int;
+use std::ptr;
+
+impl Menu {
+    pub fn popup<T: IsA<Widget>, U: IsA<Widget>,
+                 F: Fn(&Self) -> (i32, i32, bool) + 'static>(
```
EPashkin: Menu?

src/menu.rs
```diff
@@ -0,0 +1,46 @@
...
+                                    parent_menu_item.to_glib_none().0,
+                                    None, ptr::null_mut(), button, activate_time)
+            }
+        }
+    }
+}
+
+unsafe extern "C" fn position_callback<T>(this: *mut ffi::GtkMenu, x: *mut c_int, y: *mut c_int,
+                                          push_in: *mut glib_ffi::gboolean, f: glib_ffi::gpointer)
+where T: IsA<Menu> {
```
EPashkin: Not needed?

gkoz:
`Option<F>` not working nicely makes a simple variant with just two last arguments necessary rather than good to have.

I think I was wrong about `x` and `y`, they actuall are in/out parameters so it's appropriate to take them as `&mut i32`.

src/menu.rs
```diff
@@ -0,0 +1,46 @@
...
+use Widget;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use libc::c_int;
+use std::ptr;
+
+impl Menu {
+    pub fn popup<T: IsA<Widget>, U: IsA<Widget>,
+                 F: Fn(&Self) -> (i32, i32, bool) + 'static>(
+        &self, parent_menu_shell: &Option<T>, parent_menu_item: &Option<U>, f: Option<F>,
```
gkoz: Shouldn't these be `Option<&T>` and `Option<&U>`?

GuillaumeGomez: Yes, I updated it in this way.

GuillaumeGomez:
@Eilie: I think your problem isn't solve yet and it's all because of the optional `Fn` parameter.

EPashkin:
@Eilie your code compiled with this. I don't understand why.
```
    let f: Option<&(Fn(&Menu, &mut i32, &mut i32, &mut bool) + 'static)> = None;
    menu.popup(None::<&Widget>, None::<&Widget>, f, button, time);
```

Eilie:
@EPashkin Thanks, it works! A bit off topic but not sure where to ask, but how can I connect `activate` signal to `MenuItem`, it has it here https://developer.gnome.org/gtk3/stable/GtkMenuItem.html#GtkMenuItem-activate, basically I'm trying to re-implement https://gist.github.com/pklaus/304963 in Rust, and I'm almost there, just need to connect callback to my `MenuItem` which will be triggered on click. In the gist they're doing it using `menu_item.connect_object("activate", ...)` but it doesn't seem like `gtk-rs` has it.

EPashkin:
Current version with "all" signals not on cargo, so try use gtk from git
https://github.com/gtk-rs/gtk/blob/master/src/auto/menu_item.rs#L80

Eilie:
@EPashkin Got it, thank you!

GuillaumeGomez:
So is it ready to get merged?

src/menu.rs
```diff
@@ -0,0 +1,45 @@
...
+use IsA;
+use Widget;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+use libc::c_int;
+use std::ptr;
+
+impl Menu {
+    pub fn popup<T: IsA<Widget>, U: IsA<Widget>,
+                 F: Fn(&Menu, &mut i32, &mut i32, &mut bool) + 'static>(
```
gkoz: I'm sorry for the runaround with `x` and `y` but the boolean definitely is an outparam making it the return value.

GuillaumeGomez: Haha. Is it your last words?

src/menu.rs
```diff
@@ -0,0 +1,45 @@
...
+        &self, parent_menu_shell: Option<&T>, parent_menu_item: Option<&U>, f: Option<F>,
+        button: u32, activate_time: u32) {
+        unsafe {
+            if let Some(f) = f {
+                let f: Box_<Box_<Fn(&Menu, &mut i32, &mut i32, &mut bool) + 'static>> = Box_::new(Box_::new(f));
+                ffi::gtk_menu_popup(self.to_glib_none().0, parent_menu_shell.to_glib_none().0,
+                                    parent_menu_item.to_glib_none().0,
+                                    transmute(position_callback as usize),
```
gkoz: This should work without transmute.

GuillaumeGomez:
@gkoz: Updated.

gkoz:
> `let f: Option<&(Fn(&Menu, &mut i32, &mut i32, &mut bool) + 'static)> = None;`

In order to avoid this I'd prefer two functions: `popup` with `f: F` (non-optional closure) and `popup_easy(&self, button: u32, activate_time: u32)`.

src/menu.rs
GuillaumeGomez: It doesn't.

GuillaumeGomez:
Updated.

src/menu.rs
gkoz: Why? `GtkMenuPositionFunc = Option<unsafe extern "C" fn(*mut GtkMenu, *mut c_int, *mut c_int, *mut gboolean, gpointer)>`

GuillaumeGomez: You should ask rustc, not me. :smiley_cat:

gkoz:
So I asked rustc and it does work without transmutes:

```patch
diff --git a/src/menu.rs b/src/menu.rs
index b52cef6..90caa35 100644
--- a/src/menu.rs
+++ b/src/menu.rs
@@ -23,7 +23,7 @@ impl Menu {
             let f: Box_<Box_<Fn(&Menu, &mut i32, &mut i32) -> bool + 'static>> = Box_::new(Box_::new(f));
             ffi::gtk_menu_popup(self.to_glib_none().0, parent_menu_shell.to_glib_none().0,
                                 parent_menu_item.to_glib_none().0,
-                                transmute(position_callback as usize),
+                                Some(position_callback),
                                 Box_::into_raw(f) as *mut _, button, activate_time)
         }
     }
```

GuillaumeGomez:
I thought the point of using transmute was to avoid using the `Option` type.

gkoz:
`transmute` always is the last resort when safer ways aren't available.

GuillaumeGomez:
Updated.

EPashkin:
:+1:

gkoz:
Thanks!

PR #357, opened 2016-06-08
Update Cargo.toml

GuillaumeGomez:
I read this very interesting [article](https://lifthrasiir.github.io/rustlog/why-is-a-rust-executable-large.html) which talked about having more little executables. I was wondering if we could bring this in gtk as well (and if it would be useful).

cc @gkoz 
cc @EPashkin

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,9 @@ v3_12 = ["v3_10", "gtk-sys/v3_12", "gdk/v3_12"]
 v3_14 = ["v3_12", "gtk-sys/v3_14", "gdk/v3_14"]
 v3_16 = ["v3_14", "gtk-sys/v3_16", "gdk/v3_16"]
 
+[profile.release]
+lto = true
+
 [build-dependencies.gtk-rs-lgpl-docs]
 git = "https://github.com/gtk-rs/lgpl-docs"
 version = "0.1.0"


gkoz:
So, what kind of effect on the size did you see? Anyway this doesn't seem like something pertaining to GTK and I wouldn't want novices to cargo cult from us settings which have reasons not to be defaults.

GuillaumeGomez:
I didn't test (I'll once I'm back home). Since this is a very little change, if it has a significant impact on the size, I think it's worth it and would allow users' applications to be smaller as well.

EPashkin:
On my Windows7 LTO has no effect even if placed in examples config (As it IMHO option for executable)

EPashkin:
But reduced from examples target from 140Mb to 86Mb if placed in .cargo\config

GuillaumeGomez:
I thought it would impact both lib size and examples size. Interesting.

EPashkin:
It has no effect on `deps` folder.

EPashkin:
But on exe's it has great effect: 63Mb -> 8.5Mb.

GuillaumeGomez:
So, so far it's really interesting!

gkoz:
Oh I didn't notice this PR is against gtk. I believe the option only applies to final targets so it would go into the examples `Cargo.toml`. I'm seeing less than 10% reduction in examples binary sizes on Linux with 1.10 beta.

EPashkin:
For me it don't work if it added to examples.

GuillaumeGomez:
Would be interesting to give another try to this one, no? :)

EPashkin:
Hm, now my result inverses:  50M(9M exes) when placed in `examples\Cargo.toml`, 88M(44M exes) in other cases.
Still no effect when added to gtk.

gkoz:
Okay I guess we could add this to the examples repo. I still don't think it can have any effect in library crates, this should only apply to binary and cdylib crates.

LordZane:
lto only effects binary crates I believe, libs can't use it. Whatever binary uses this lib sets their own policy for lto. And it's probably not worth adding to the examples, just slows down compile time for less space used on a temporary binary. Not to mention lto dosen't affect native libs (gtk.dll etc), and is only considered for rust code

PR #375, opened 2016-08-26
Provide gtk::MessageDialog manual setter for secondary markup and text

hfiguiere:
format_secondary_text() and format_secondary_markup() aren't wrapped due varargs.

Implemented plain setters so that we can set the secondary text/markup. Use the Rust format!() to format it.

### DIFF
--- a/src/message_dialog.rs
+++ b/src/message_dialog.rs
@@ -22,8 +22,41 @@ impl MessageDialog {
             let message: Stash<*const c_char, _> = message.to_glib_none();
             Widget::from_glib_none(
                 ffi::gtk_message_dialog_new(parent.to_glib_none().0, flags.to_glib(), type_.to_glib(), buttons.to_glib(),
-                    b"%s\0".as_ptr() as *const c_char, message.0, ptr::null_mut::<*const c_char>()))
+                    b"%s\0".as_ptr() as *const c_char, message.0, ptr::null::<c_char>()))
                 .downcast_unchecked()
         }
     }
+
+    pub fn set_secondary_markup(&self, message: Option<&str>) {
+        match message {
+            Some(m) => unsafe {
+                let message: Stash<*const c_char, _> = m.to_glib_none();
+                ffi::gtk_message_dialog_format_secondary_markup(
+                    self.to_glib_none().0, b"%s\0".as_ptr() as *const c_char, message.0,
+                    ptr::null::<c_char>())
+            },
+            None => unsafe {
+                ffi::gtk_message_dialog_format_secondary_markup(
+                    self.to_glib_none().0,
+                    ptr::null::<c_char>())
+            },
+        }
+    }
+
+    pub fn set_secondary_text(&self, message: Option<&str>) {
+        match message {
+            Some(m) => unsafe {
+                let message: Stash<*const c_char, _> = m.to_glib_none();
+                ffi::gtk_message_dialog_format_secondary_text(
+                    self.to_glib_none().0,
+                    b"%s\0".as_ptr() as *const c_char, message.0, ptr::null::<c_char>())
+            },
+            None => unsafe {
+                ffi::gtk_message_dialog_format_secondary_text(
+                    self.to_glib_none().0,
+                    ptr::null::<c_char>())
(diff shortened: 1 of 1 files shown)


hfiguiere:
Also replace ptr::null_mut() by ptr::null() and changed its type to be correct (T is turned to *const T). It was working because these are NULL to terminate the varargs

EPashkin:
@hfiguiere, thanks for PR.
Please, add ignoring these function and 'new' for `MessageDialog` to `Gir.toml` and do regen

EPashkin:
On second thought as `gtk_message_dialog_new_with_markup` can be done in same way, seems better do all `GtkMessageDialog` manually and remove it from generation.
Also seems all `message` parameters better be `Option<&str>`
@gkoz, @GuillaumeGomez what you think?

GuillaumeGomez:
`&str` parameters should always be `Option<&str>` because we have to be able to provide `NULL` equivalent. If possible, I'd prefer to keep generation as much as possible.

gkoz:
Seems good as it is to me. With regard to `Option<&str>` it's consistent with whether `message_format` is declared as `nullable`. `new_with_markup` can't be implemented in the same way because we're passing the whole string as a single argument and it will get escaped inappropriately.

hfiguiere:
Only ````gtk_message_dialog_format_secondary_text()```` has ````message_format```` declared as nullable. Hence that's the only one I have declared using Option<&str>, while it is true the doc says "or NULL" for ````gtk_message_dialog_format_secondary_markup()````

gkoz:
@GuillaumeGomez @EPashkin since there were some differences in opinions we need at least one of you to chime in.

GuillaumeGomez:
I maintain my previous position. :)

gkoz:
To expand on this note that removing the secondary text is achievable via `set_secondary_text(None)` and making `set_secondary_markup` also take `Option` just penalizes its users: if someone asks why they have to pass `Some` to it (`markup` is a special case of `text` implying the text is not empty) I don't see how I could justify it.

EPashkin:
Agree with @gkoz, 2 `Option` unneeded as functions paired and change one parameter.
So I'm :+1:  for accept this PR as is.

hfiguiere:
@GuillaumeGomez I now have taken into account your suggestion. Both accept an Option<&str> for the parameter.

GuillaumeGomez:
@hfiguiere: Great, thanks!
@EPashkin & @gkoz: All good for you?

gkoz:
@GuillaumeGomez It'd be nice to know your reasoning for making both arguments optional, considering the majority preference to not do it and inconsistency with metadata we have. To reiterate my unanswered question: since the `set_secondary_markup` method is an extension of `set_secondary_text` why should one have to pass `Some(...)` to `set_secondary_markup`?

GuillaumeGomez:
Like I previously said, all `&str` should be `Option<&str>` to allow to make the difference between `NULL` and an empty string. Now if there is a specific reason in here which demonstrate that a `NULL` pointer cannot be handled by gtk, then you can remove it.

gkoz:
In this context my point is what would be the reason to pass either `None` or `""` to `set_secondary_markup`? Neither has any markup.

GuillaumeGomez:
Then whatever.

hfiguiere:
In both set_secondary_text() and set_secondary_markup(), passing NULL will reset the state as to "no secondary text". This has an impact on the message dialog.

hfiguiere:
As I said before, only set_secondary_text() is annotated as allowing NULL, but according to the documentation, both supports it.

gkoz:
After digging through the history the lack of an `allow-none` annotation in the 'markup' case appears to be unintentional (they [added the annotation with a script](https://github.com/GNOME/gtk/commit/fe852721123fb7dc5749c690613ca6e209363e2f), which failed to handle a multiline docstring or something). This makes my argument against `Option` weaker so I'll just merge.
@hfiguiere sorry for the delay.

GuillaumeGomez:
Victory. :laughing: 

:tada: 

Thanks for your work @hfiguiere!

PR #387, opened 2016-10-16
Adding TreeRowReference

thk1:
Adding TreeRowReference (persistent reference to a node in a TreeModel). new_proxy() and related methods are ignored since reordered() can't be auto-generated. However, proxy references are "not generally needed by most applications." (cit. from docs)


EPashkin:
IMHO `reordered` fact that not generated is no reason to ignore it (without comment we just forget about that function and never implement it 😉 ).
Same as for other: while it normally used together with `reordered`, IMHO there are cases where allowed only insert\remove without reordering and `new_proxy` can be safely used

GuillaumeGomez:
Seems good for me. Waiting for @gkoz's opinion.

thk1:
@EPashkin Maybe you are right. Generation of new_proxy/inserted/deleted is now enabled. I also added a manual implementation of reordered(). TreeRowReference bindings are now complete.

src/tree_row_reference.rs
```diff
@@ -0,0 +1,25 @@
...
+use libc::c_int;
+use TreeIter;
+use TreePath;
+use TreeRowReference;
+
+impl TreeRowReference {
+
+    pub fn reordered<T: IsA<glib::Object>>(proxy: &T, path: &TreePath, iter: Option<&TreeIter>, new_order: &[u32]) {
+        assert_initialized_main_thread!();
+        assert!(iter.is_some() || path.get_depth()==0, "If 'iter' is None, 'path' must point to the root.");
```
gkoz: We also need to assert something (I'm not yet sure what exactly) about the length of `new order`. [This line](https://github.com/GNOME/gtk/blob/ee7ddefce952399f07edbeab465617c98fdccb18/gtk/gtktreemodel.c#L2233) is liable to blow up.

src/auto/tree_row_reference.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    match fn {
+        copy => |ptr| ffi::gtk_tree_row_reference_copy(mut_override(ptr)),
+        free => |ptr| ffi::gtk_tree_row_reference_free(ptr),
+    }
+}
+
+impl TreeRowReference {
+
+    pub fn new<T: IsA<TreeModel>>(model: &T, path: &TreePath) -> TreeRowReference {
```
gkoz: Apparently this class's constructors are fallible i.e. we can get a `NULL` for an invalid path. So we need to make this return `Option<TreeRowReference>`.

src/auto/tree_row_reference.rs
```diff
@@ -0,0 +1,71 @@
...
+        skip_assert_initialized!();
+        unsafe {
+            from_glib_full(ffi::gtk_tree_row_reference_new(model.to_glib_none().0, mut_override(path.to_glib_none().0)))
+        }
+    }
+
+    pub fn new_proxy<T: IsA<glib::Object>, U: IsA<TreeModel>>(proxy: &T, model: &U, path: &TreePath) -> TreeRowReference {
```
gkoz: Ditto.

src/tree_row_reference.rs
thk1: I think we don't have access to the TreeModel (proxy can be anything) which we would need for figuring out the size. But, [according to the GTK docs](https://developer.gnome.org/gtk3/stable/GtkTreeModel.html#gtk-tree-row-reference-new-proxy), the whole proxy thing is only public for rare monitoring reasons and TreeRowReference::reordered is then directly connected to TreeModel's reordered signal. Hence, in this case all arguments are valid.

src/auto/tree_row_reference.rs
thk1: See this [comment](https://github.com/gtk-rs/gtk/pull/387#discussion_r85933474).

src/tree_row_reference.rs
gkoz: Thanks for pointing out that we don't have the means to find out the minimum length of `new_order`. I'm afraid the function has to be marked `unsafe` then for lack of bounds checks.

src/auto/tree_row_reference.rs
gkoz: I was hoping it wasn't the case but we'll have to make the generator respect `nullable` in constructors (https://github.com/gtk-rs/gir/issues/289). The assertions are there behind the scenes already but it'd be more consistent with other treeview-path-related functions to return `Option`. We'll have to fix `TreePath::new_from_string` and other similar cases as well.

The vast majority of constructors are infallible though, that's why the generator doesn't handle this case.

EPashkin:
Added nullable constructor support to gir.
@gkoz Better add nullable for `TreePath.new_from_string` here or we add it separately?

src/auto/tree_row_reference.rs
thk1: Done thanks to [gtk-rs/gir#290](https://github.com/gtk-rs/gir/pull/290). I'm not sure - may I update the gir module for a regen? (last update is six month back)

src/tree_row_reference.rs
thk1: Okay, it's now marked `unsafe`.

src/auto/tree_row_reference.rs
EPashkin: Yes, you need "Update gir submodule" commit to complete PR and then move "Regen" after it and update "Regen" to `gir (3b3585)`.
No changes affecting `gtk` generated code made since last update.

thk1: Done.

thk1:
All requested changes have been carried out. I just rebased the PR and tested the code again. I think it should now be ready to be merged.

EPashkin:
@thk1 There no `TreeRowReference` generated and `src/tree_row_reference.rs` not added as module.
So this PR currently adds nothing.

thk1:
@EPashkin Ok, I wasn't sure if a regen should be part of the PR. The mod is now imported and auto/mod.rs and auto/tree_row_reference.rs are (re)generated.

EPashkin:
Seems all good now, waiting for CI just in case

EPashkin:
Thanks, @thk1 

@GuillaumeGomez 👍 for this.

src/tree_row_reference.rs
```diff
@@ -0,0 +1,24 @@
...
+use glib;
+use glib::translate::*;
+use glib::object::IsA;
+use libc::c_int;
+use TreeIter;
+use TreePath;
+use TreeRowReference;
+
+impl TreeRowReference {
+
```
GuillaumeGomez: Please remove this line.

src/tree_row_reference.rs
```diff
@@ -0,0 +1,24 @@
...
+    // This is unsafe because new_order bounds can't be checked.
+    pub unsafe fn reordered<T: IsA<glib::Object>>(proxy: &T, path: &TreePath, iter: Option<&TreeIter>, new_order: &[u32]) {
+        assert_initialized_main_thread!();
+        assert!(iter.is_some() || path.get_depth()==0, "If 'iter' is None, 'path' must point to the root.");
+        ffi::gtk_tree_row_reference_reordered(proxy.to_glib_none().0, mut_override(path.to_glib_none().0),
+            mut_override(iter.to_glib_none().0), mut_override(new_order.as_ptr() as *const c_int));
+    }
+
```
GuillaumeGomez: Please remove this line.

GuillaumeGomez:
Just two nits and we're good to go.

thk1:
Done.

EPashkin:
@GuillaumeGomez, merge this please, before @sdroege's

sdroege:
Does not interfere with my PRs AFAIU. Should be good to merge before or after without problems.

GuillaumeGomez:
Thanks!

Issue #392, opened 2016-11-03
Child properties


EPashkin:
As mentioned in https://github.com/gtk-rs/gir/issues/287 gir can't generate child properties, but we can do it manually with using `gtk_container_child_g[s]et_property`.
This is tracking issue.

Classes with child properties.
- [x] [GtkActionBar](https://developer.gnome.org/gtk3/stable/GtkActionBar.html)
- [x] [GtkAssistant](https://developer.gnome.org/gtk3/stable/GtkAssistant.html) added without deprecated in 3.2 `header-image` and `sidebar-image`
- [x] [GtkButtonBox](https://developer.gnome.org/gtk3/stable/GtkButtonBox.html) already has functions
- [x] [GtkBox](https://developer.gnome.org/gtk3/stable/GtkBox.html) only need position
- [x] [GtkFixed](https://developer.gnome.org/gtk3/stable/GtkFixed.html)
- [x] [GtkGrid](https://developer.gnome.org/gtk3/stable/GtkGrid.html)
- [x] [GtkHeaderBar](https://developer.gnome.org/gtk3/stable/GtkHeaderBar.html)
- [x] [GtkLayout](https://developer.gnome.org/gtk3/stable/GtkLayout.html)
- [x] [GtkMenu](https://developer.gnome.org/gtk3/stable/GtkMenu.html)
- [x] [GtkNotebook](https://developer.gnome.org/gtk3/stable/GtkNotebook.html) already has functions 
- [x] [GtkOverlay](https://developer.gnome.org/gtk3/stable/GtkOverlay.html) has gtk_overlay_get_overlay_pass_through, only needed index
- [x] [GtkPaned](https://developer.gnome.org/gtk3/stable/GtkPaned.html)
- [x] [GtkPopoverMenu](https://developer.gnome.org/gtk3/stable/GtkPopoverMenu.html)
- [x] [GtkStack](https://developer.gnome.org/gtk3/stable/GtkStack.html)
- [x] [GtkToolbar](https://developer.gnome.org/gtk3/stable/GtkToolbar.html) [PR](https://github.com/gtk-rs/gtk/pull/399)
- [x] [GtkToolItemGroup](https://developer.gnome.org/gtk3/stable/GtkToolItemGroup.html) only has position, need all other [PR](https://github.com/gtk-rs/gtk/pull/394)
- [x] [GtkToolPalette](https://developer.gnome.org/gtk3/stable/GtkToolPalette.html) already has functions

[...]

PR #394, opened 2016-11-03
Add child properties for ToolItemGroup

GuillaumeGomez:
Part of #392.

I think the code isn't very good. I intend to create a macro to avoid code repetition but before that, I need to be sure that the code is the one expected.

cc @gkoz @EPashkin

### DIFF
--- a/src/tool_item_group.rs
+++ b/src/tool_item_group.rs
@@ -1,0 +1,77 @@
+// Copyright 2016, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use ffi;
+use ToolItem;
+use ToolItemGroup;
+use IsA;
+use glib::translate::*;
+use glib::{ToValue, Value};
+use Widget;
+
+impl ToolItemGroup {
+    pub fn set_item_expand<T: IsA<ToolItem> + IsA<Widget>>(&self, item: &T, expand: bool) {
+        unsafe { ffi::gtk_container_child_set_property(self.to_glib_none().0,
+                                                       item.to_glib_none().0,
+                                                       "expand".to_glib_none().0,
+                                                       Value::from(&expand).to_glib_none().0) }
+    }
+
+    pub fn get_item_expand<T: IsA<ToolItem> + IsA<Widget>>(&self, item: &T) -> bool {
+        let mut value = false.to_value();
+        unsafe { ffi::gtk_container_child_get_property(self.to_glib_none().0,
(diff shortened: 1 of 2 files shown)


src/tool_item_group.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn set_item_expand<T: IsA<ToolItem> + IsA<Object> + IsA<Widget>>(&self, item: &T, expand: bool) {
+        let object = self.clone();
+        unsafe { ffi::gtk_container_child_set_property(object.upcast::<Container>().to_glib_none().0,
+                                                       item.to_glib_none().0,
+                                                       "expand".to_glib_none().0,
+                                                       Value::from(&expand).to_glib_none().0) }
+    }
+
+    pub fn get_item_expand<T: IsA<ToolItem> + IsA<Object>+ IsA<Widget>>(&self, item: &T) -> bool {
+        let mut value = unsafe { Value::uninitialized() };
```
EPashkin: For `i32` this not work, IMHO for `bool` too.
I use `let mut value = 0.to_value();`

src/tool_item_group.rs
```diff
@@ -0,0 +1,87 @@
...
+use IsA;
+use glib::translate::*;
+use glib::Value;
+use Container;
+use Widget;
+
+impl ToolItemGroup {
+    pub fn set_item_expand<T: IsA<ToolItem> + IsA<Object> + IsA<Widget>>(&self, item: &T, expand: bool) {
+        let object = self.clone();
+        unsafe { ffi::gtk_container_child_set_property(object.upcast::<Container>().to_glib_none().0,
```
EPashkin: For grid it works fine without `object`

src/tool_item_group.rs
GuillaumeGomez: Great idea!

GuillaumeGomez:
Thanks to @EPashkin, it's way better now!

GuillaumeGomez:
@EPashkin: So, what do we do about this?

EPashkin:
@GuillaumeGomez we don't have @gkoz's opinion, maybe he have better solution.
If we use local macros (as in #395) I like to hear your option: there better current multiproperty macro or one macro usage per property? Now I think that last option better.
Also we don't have tested code for string properties.

GuillaumeGomez:
Like I said, my primary concern about the macro usage is what the generated docs will look like. I need to test first.

GuillaumeGomez:
Yeay! \o/

GuillaumeGomez:
All ok. Do you want to add something or can I merge?

EPashkin:
All ok, thanks

GuillaumeGomez:
Great, thanks for this! :)

PR #395, opened 2016-11-03
Add child properties to Grid

EPashkin:
Part of #392.

WIP
Published as testing place for int child properties.

### DIFF
--- a/src/grid.rs
+++ b/src/grid.rs
@@ -1,0 +1,72 @@
+use ffi;
+use glib::translate::*;
+use glib::Value;
+use IsA;
+use Grid;
+use Widget;
+
+impl Grid {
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Both of our codes are exactly the same now. I guess we're good! :D

EPashkin:
Not exactly: I use `Value::from` for initialize now, it will be changed when we make macros.

EPashkin:
Added macro. As I not found way macroing function constraints it stay in `src/grid.rs`.
IMHO multiplicity of macro arguments not really needed.

GuillaumeGomez:
I'm a bit worry about what the generated doc will look like on this... :-/

PR #399, opened 2016-11-04
Add child properties for Toolbar

GuillaumeGomez:
cc @gkoz @EPashkin

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -213,6 +213,7 @@ mod socket;
 mod text_attributes;
 mod text_buffer;
 mod text_iter;
+mod toolbar;
 mod tree_model_filter;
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
@EPashkin: So, what should we do about all the property PRs? They seem good for me, so except if you have a reason to not merge them, I'm thinking about merging the current ones and implement the missing ones.

EPashkin:
Seems @gkoz don't have time to check it.
So no reason not to merge this and https://github.com/gtk-rs/gtk/pull/394.
I change my PR too: remove multiplicity macros or macros itself as you say.
Also can you add link to https://github.com/gtk-rs/gtk/issues/392 in next PRs for reference.

PR #414, opened 2016-12-07
Add set_translation_domain and get_translation_domain for gtk::Builder

### LINKED ISSUE #413: Missed translation of text from *.glade files using gtk::Builder
Sorry, I'm not very familiar with the GTK, but it seems that there is no support for text translation using gettext and `gtk::Builder`. I think [this method](https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-set-translation-domain) from GTK documentation related to the problem, and it is not implemented yet.
At the moment, I need to manually re-set the text for the elements made in the Glade, wrapping it in `gettext` call (from [gettext-rs](https://github.com/Koka/gettext-rs) crate) in my Rust code. Perhaps there is another way or this feature will appear in future?

fludardes:
Fix #413 
`gtk::Builder` methods `set_translation_domain` and `get_translation_domain` were added.

### DIFF
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -91,4 +91,16 @@ impl Builder {
             if error.is_null() { Ok(()) } else { Err(from_glib_full(error)) }
         }
     }
+    
+    pub fn set_translation_domain(&self, domain: Option<&str>) {
+        unsafe {
+            ffi::gtk_builder_set_translation_domain(self.to_glib_none().0, domain.to_glib_none().0);
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


EPashkin:
👍 for this.
@fludardes, please, change in first comment "Issue #..." to "Fix #...", this way merging PR close issue

GuillaumeGomez:
I just wait for the tests to end and then I merge. Thanks @fludardes!

GuillaumeGomez:
All good, merging.

PR #420, opened 2017-01-02
Generate TextAttributes

EPashkin:
Only visible difference is removing `mut_override` in `copy`


src/auto/text_attributes.rs
```diff
@@ -0,0 +1,36 @@
...
+    pub struct TextAttributes(Shared<ffi::GtkTextAttributes>);
+
+    match fn {
+        ref => |ptr| ffi::gtk_text_attributes_ref(ptr),
+        unref => |ptr| ffi::gtk_text_attributes_unref(ptr),
+    }
+}
+
+impl TextAttributes {
+
```
GuillaumeGomez: Since when do we generate an empty line in here? :-/

EPashkin: Somewhere between 15fe1aa and 34ea1b9 in gir :(

GuillaumeGomez: I'll let you fix it. ;) Meanwhile, I merge this one.

EPashkin: https://github.com/gtk-rs/gir/commit/cac0a0f47b424044c3bd066 ;)

GuillaumeGomez:
Thanks!

src/auto/text_attributes.rs
EPashkin: Ok.

PR #421, opened 2017-01-02
Tree sortable ignore get sort column

EPashkin:
@GuillaumeGomez, I just don't remember why get was not ignored.
May be it was some reason (backward compatibility etc.)?

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -1141,7 +1141,7 @@ status = "generate"
 name = "Gtk.TreeSortable"
 status = "generate"
     [[object.function]]
-    name = "set_sort_column_id"
+    pattern = "[gs]et_sort_column_id"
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
I don't remember at all why it was bound in the first place. Whatever, once tests passed, I merge.

EPashkin:
Enum `SortColumn` was added to manual part to separate -1 and other column numbers

EPashkin:
.., so manual part use enum, not i32 https://github.com/gtk-rs/gtk/blob/master/src/tree_sortable.rs#L86

GuillaumeGomez:
Oh I see. Makes sense.

GuillaumeGomez:
Thanks!

PR #435, opened 2017-01-29
Deignoring

EPashkin:
Generate types, needed for currently generated methods.
Also ignore property `Widget::style` that working with deperecated 3.0 [`Style`](https://developer.gnome.org/gtk3/stable/GtkStyle.html)


GuillaumeGomez:
Yeay! \o/

Issue #442, opened 2017-02-08
Enable drag_dest_set()


Susurrus:
Need to add `DestDefaults`, `TargetEntry`, and `gdk::DragAction` to the Gir generation script. I couldn't figure out why adding `DragAction` to `gdk`s `Gir.toml` didn't generate anything, so I'm posting this issue instead of a general PR.

EPashkin:
With `DragAction` is no problem, next diff will generate it.
Problem with `TargetEntry`:
1. It almost only one class that used with `const*` in sys, and your `ToGlibPrt` implementation for `Boxed<ffi::GtkTargetEntry>` defined for `mut*`, so many build error appears.
2. It don't have usable function or get/setters and `gir` currently don't support direct access to record fields.
(May be this record really don't need those and `new` is enough to use)
```diff
--- a/Gir.toml
+++ b/Gir.toml
@@ -9,6 +9,7 @@ generate_safety_asserts = true
 
 generate = [
     "Gdk.CursorType",
+    "Gdk.DragAction",
     "Gdk.GLError",
     "Gdk.Gravity",
     "Gdk.ModifierIntent",
```

Susurrus:
So it sounds like there needs to be a manual implementation of `GtkTargetEntry` for this functionality to be exposed? Anyone interested in tackling that?

GuillaumeGomez:
Why not doing it? Might be nice for you to discover a bit how it works under the hood. :p

Susurrus:
As of gtk-rs/gdk#144 only a manual implementation of `TargetEntry` is necessary to get `drag_source` and `drag_dest` functions working.

EPashkin:
IMHO as with [GdkRGBA](https://github.com/gtk-rs/gdk/blob/master/src/rgba.rs) better almost ignore `free` and `copy` functions and declare normal rust struct with `ToGlibPtr` ..

PR #443, opened 2017-02-08
Add more missing Gtk types

Susurrus:
Partially address #442.

Haven't used any of this code, but didn't see any warnings about generating it. There were some GTK types that can't be generated, so I commented them out. Should I add them to the `ignore` list instead?


Gir.toml
```diff
@@ -175,7 +182,13 @@ generate = [
     "Gtk.StateType",
     "Gtk.Statusbar",
     "Gtk.StatusIcon",
+    "Gtk.Style",
```
EPashkin: Its deprecated in 3.0, so better don't generate.

src/auto/border.rs
```diff
@@ -0,0 +1,23 @@
+// This file was generated by gir (8a0b5e5) from gir-files (71d73f0)
+// DO NOT EDIT
+
+use ffi;
+use glib::translate::*;
+
+glib_wrapper! {
+    pub struct Border(Boxed<ffi::GtkBorder>);
```
EPashkin: Without fields it IMHO unusable

src/auto/target_list.rs
```diff
@@ -0,0 +1,61 @@
...
+    pub struct TargetList(Shared<ffi::GtkTargetList>);
+
+    match fn {
+        ref => |ptr| ffi::gtk_target_list_ref(ptr),
+        unref => |ptr| ffi::gtk_target_list_unref(ptr),
+    }
+}
+
+impl TargetList {
+    //pub fn new(targets: /*Ignored*/&[&TargetEntry], ntargets: u32) -> TargetList {
```
EPashkin: No reason to generate record without constructors

EPashkin:
Strange that it builds, as I remember `WidgetPath` has mixed `const/mut*` usage and fails.

Susurrus:
I think I resolved all of your comments, so here's another batch of changes. Sorry about not checking over the generated code, I looked over the remaining generated types and they all seemed to have enough useful methods to be worthwhile.

GuillaumeGomez:
Seems ok for me. Anything else for you @EPashkin?

EPashkin:
style_context.rs don't builds on gtk/v3_20.
Function `to_string` conflict in `impl fmt::Display for StyleContext`
`Gir` need change to don't generate this impl for nonstandard functions.
@GuillaumeGomez, seems all good.

GuillaumeGomez:
@EPashkin: I merge then.
@Susurrus: Thanks!

Susurrus:
Given the fix in #313 I think there'll need to be another regent to make StylrContext correct, yes?

EPashkin:
I already made [PR](https://github.com/gtk-rs/gtk/pull/444), but forgot link this.

PR #444, opened 2017-02-08
Remove bad impl display

EPashkin:
@GuillaumeGomez, as only one file "style_context.rs" really changed, this PR IMHO can be ignored.


GuillaumeGomez:
Any fix is good to take from my point of view so I don't mind merging it.

EPashkin:
I close this in favor of #446

PR #452, opened 2017-02-20
RadioButton & RadioMenuItem

Susurrus:
I went to add the `RadioMenuItem` and was looking at how to instantiate them, but it didn't make sense given the `group` argument, which is nullable, is not an `Option` in the generated code. I then looked at the `RadioButton` code and it has the same problem. I didn't see any example or reference code that uses `RadioButton`, so I'm not sure that it's correct. I've "corrected" its autogeneration parameters and added `RadioMenuItem`, though adding these settings didn't seem to change anything. It seems right based on my understanding of Gir and I see no errors or warning when generating this code.


Gir.toml
```diff
@@ -895,6 +894,24 @@ status = "generate"
         const = true
 
 [[object]]
+name = "Gtk.RadioButton"
+status = "generate"
+    [[object.function]]
+    name = "new|new_with_label|new_with_mnemonic"
```
EPashkin: This wont work:
- not `name` but `pattern`,
- for patterns always added `^` and `$`, so `|` may not work at all
- due https://github.com/gtk-rs/gir/blob/master/src/analysis/rust_type.rs#L131 lists ignore nullable. I don't remember why @gkoz do that, but maybe empty list works same as null (not checked with radiobuttons)

Susurrus:
@EPashkin I've regenerated then with what should be necessary then, just adding RadioMenuItem.

EPashkin:
👍 from me

GuillaumeGomez:
Thanks!

Susurrus:
One problem with this PR that I forgot to mention is that no docs are generated for `gtk::RadioMenuItem`. This is especially problematic given that it's not clear from the regular GTK+ docs how to create a `RadioMenuItem` widget since the `group` argument isn't an `Option`. I didn't even think to pass an empty list. So is there something I missed in this PR that needed to be done to get the docs from GTK+ added? And additionally is there an easy way to modify those docs such that they'll talk about the empty list for group instead of NULL?

GuillaumeGomez:
Euh... That's actually a good question.

PR #455, opened 2017-02-22
Now use SourceId

hfiguiere:
Now use SourceId since we have it in glib.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -6,6 +6,7 @@ use std::cell::RefCell;
 use std::mem::transmute;
 
 pub use glib::signal::Inhibit;
+use glib::SourceId;
 use glib::translate::*;
 
 use glib_ffi::{self, gboolean, gpointer};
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -41,13 +42,13 @@ fn into_raw<F: FnMut() -> Continue + 'static>(func: F) -> gpointer {
 ///
 /// Similar to `glib::idle_add` but only callable from the main thread and
 /// doesn't require `Send`.
-pub fn idle_add<F>(func: F) -> u32
+pub fn idle_add<F>(func: F) -> SourceId
     where F: FnMut() -> Continue + 'static {
     assert_initialized_main_thread!();
-    unsafe {
+    from_glib(unsafe {
```
EPashkin: Missed that in glib commit: normally `from_glib` moved to `unsafe` block.

src/signal.rs
```diff
@@ -8,6 +8,7 @@ use std::mem::transmute;
 pub use glib::signal::Inhibit;
 use glib::translate::*;
 
+use glib::SourceId;
```
EPashkin: Move 2 row upper?

src/signal.rs
hfiguiere: I thought that since the compiler is happy with it, it is better to have it out of the unsafe {} block. Will move it inside.

EPashkin:
:+1: Thanks.

GuillaumeGomez:
Thanks!

PR #471, opened 2017-03-17
Gdk update

Susurrus:
This enables all types that were added in gtk-rs/gdk#144


GuillaumeGomez:
👍 

cc @EPashkin

EPashkin:
👍  
@Susurrus Thanks

Susurrus:
One thing I meant to note when I first posted that was that I don't know why `gdk::Atom` is still ignored for generating functions. There are still about 25 functions that only depend on `Gdk::Atom` and were not generated. Is there something I'm missing for that?

EPashkin:
I don't remember reason but seems need to add both "Gdk.Atom" and  "Gdk.Atom_" to deignore it
and to gdk too.

PR #472, opened 2017-03-17
Drag targets

### LINKED ISSUE #442: Enable drag_dest_set()

Susurrus:
Automatically generates `TargetFlags` and `TargetList` and manually implements `TargetEntry`. This follows the same pattern as that used by `gdk::RGBA` following @EPashkin's advice.

There are a few issues I see here:

  1. The in-memory layout of `gdk::RGBA` and `gdk_sys::GdkRGBA` is identical, so casting between them was trivial while I don't believe that holds true for `gtk::TargetEntry` and `gtk_sys::GtkTargetEntry`.
  2. I was getting compilation errors when autogenerating functions that tool an array of `TargetEntry`s so I disabled all functions that do that, though I obviously would like to have them as it includes the functions I want to use.
  3. I'm getting compilation errors for `SetValue` and `SetValueOptional` because of the datatype of `to_glib_none()`, which I didn't get with `gdk::RGBA`, but I don't know if I even need those traits implemented.
 
Closes #442.


src/target_entry.rs
```diff
@@ -0,0 +1,104 @@
...
+use glib;
+use glib::translate::*;
+use gobject_ffi;
+use std::mem;
+use TargetFlags;
+
+#[derive(Clone, Debug)]
+#[repr(C)]
+pub struct TargetEntry {
+    pub target: String,
```
GuillaumeGomez: Is is normal that every field is public?

Susurrus:
All of the fields for `gdk::RGBA` were public, but I don't think we want it for this type. After initialization i don't think these structs are really modified unlike with `gdk::RGBA`.

EPashkin:
If it fields not public we need getters

src/target_entry.rs
```diff
@@ -0,0 +1,104 @@
...
+    pub flags: TargetFlags,
+    pub info: u32,
+}
+
+impl TargetEntry {
+    pub fn new(target: &str, flags: TargetFlags, info: u32) -> TargetEntry {
+        assert_initialized_main_thread!();
+        unsafe {
+            from_glib_full(ffi::gtk_target_entry_new(target.to_glib_none().0, flags.bits(), info))
```
EPashkin: IMHO don't need ffi here just create new struct

src/target_entry.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+#[doc(hidden)]
+impl<'a> ToGlibPtr<'a, *const ffi::GtkTargetEntry> for TargetEntry {
+    type Storage = &'a Self;
+
+    #[inline]
+    fn to_glib_none(&'a self) -> Stash<'a, *const ffi::GtkTargetEntry, Self> {
+        let ptr: *const TargetEntry = &*self;
+        Stash(ptr as *const ffi::GtkTargetEntry, self)
```
EPashkin: You need create new GtkTargetEntry and fill it fields and assign it to first stash parameter
As we need hold both Self and this then `Stash<'a, *const ffi::GtkTargetEntry, (Self, ffi::GtkTargetEntry)>`

src/target_entry.rs
```diff
@@ -0,0 +1,104 @@
...
+    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::GtkTargetEntry, Self> {
+        let ptr: *mut TargetEntry = &mut *self;
+        StashMut(ptr as *mut ffi::GtkTargetEntry, self)
+    }
+}
+
+#[doc(hidden)]
+impl FromGlibPtrNone<*const ffi::GtkTargetEntry> for TargetEntry {
+    unsafe fn from_glib_none(ptr: *const ffi::GtkTargetEntry) -> Self {
+        *(ptr as *const TargetEntry)
```
EPashkin: As in constructor just create new rust struct

GuillaumeGomez:
I'd prefer getters then. It'll allow us to keep full control.

src/target_entry.rs
```diff
@@ -0,0 +1,104 @@
...
+}
+
+impl glib::StaticType for TargetEntry {
+    fn static_type() -> glib::types::Type {
+        skip_assert_initialized!();
+        unsafe { from_glib(ffi::gtk_target_entry_get_type()) }
+    }
+}
+
+impl glib::value::FromValueOptional for TargetEntry {
```
EPashkin: This impl and all bottom need only if we have property with TargetEntry type.

src/target_entry.rs
Susurrus: If I understand you correctly this should then be:
```
impl<'a> ToGlibPtrMut<'a, *mut ffi::GtkTargetEntry> for TargetEntry {
    type Storage = &'a mut Self;

    #[inline]
    fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::GtkTargetEntry, (Self, ffi::GtkTargetEntry)> {
        let target_entry = ffi::GtkTargetEntry {
            target: self.as_ptr(),
            flags: self.flags.bits(),
            info: self.info,
        };
        StashMut((self, target_entry))
    }
}
```

But if I do that I get the following error:

```
$ cargo build
   Compiling gtk v0.1.1 (file:///home/susurrus/Projects/gtk)
error[E0277]: the trait bound `(target_entry::TargetEntry, ffi::GtkTargetEntry): glib::translate::ToGlibPtrMut<'a, *mut ffi::GtkTargetEntry>` is not satisfied
  --> src/target_entry.rs:52:5
   |
52 |       fn to_glib_none_mut(&'a mut self) -> StashMut<'a, *mut ffi::GtkTargetEntry, (Self, ffi::GtkTargetEntry)> {
   |  _____^ starting here...
53 | |         let target_entry = ffi::GtkTargetEntry {
54 | |             target: self.as_ptr(),
55 | |             flags: self.flags.bits(),
56 | |             info: self.info,
57 | |         };
58 | |         StashMut((self, target_entry))
59 | |     }
   | |_____^ ...ending here: the trait `glib::translate::ToGlibPtrMut<'a, *mut ffi::GtkTargetEntry>` is not implemented for `(target_entry::TargetEntry, ffi::GtkTargetEntry)`
   |
   = note: required by `glib::translate::StashMut`

error: Could not compile `gtk`.

To learn more, run the command again with --verbose.
```

I should actually be implementing these traits on the type `(gtk::TargetEntry, ffi::GtkTargetEntry)`?

Susurrus:
Should I also add 3 getters for all 3 fields (`get_target`, `get_flags`, and `get_info`)?

GuillaumeGomez:
> Should I also add 3 getters for all 3 fields (get_target, get_flags, and get_info)?

Yes please.

src/target_entry.rs
EPashkin: Sorry was wrong where tuple go. It need go to storage without changing function result.
See https://github.com/gtk-rs/gtk/blob/master/src/recent_data.rs#L21

src/target_entry.rs
Susurrus: There are two other manually-implemented structs in gtk, `RecentData` and `Requisition`, both of which expose all members as public. I'm leaning towards us just doing the same.

Susurrus:
I think we're pretty close to a final solution here. I've removed the unnecessary `glib::value` trait impls, kept the field public in line with the other manually-implemented structs in Gtk, and fixed the `ToGlibPtr*` impls

Susurrus:
Also there are several drag-related methods on `Widget`s that are blocking on the `gdk::Atom` in gtk-rs/gdk#145, so I suggest we wait to merge this until after that's merged.

Susurrus:
The last thing that needs to be done here is to handle the various `drag` functions that take an array of `TargetEntry`s, like `WidgetExt::drag_dest_set` (which is what #442 actually refers to). The hard part here is converting from a `&[TargetEntry]` to a `&[ffi::GtkTargetEntry]` to be passed as a pointer to `ffi::gtk_drag_dest_set`. What's the correct way to do this conversion as the following is not right:
```
    fn drag_dest_set(&self, flags: DestDefaults, targets: &[TargetEntry], actions: DragAction) {
        let glib_targets: Vec<ffi::GtkTargetEntry> = targets.iter().map(|e| e.to_glib_none().0).collect();
        unsafe {
            ffi::gtk_drag_dest_set(self.to_glib_none().0,
                                   flags.to_glib(),
                                   glib_targets.as_ptr() as *mut ffi::GtkTargetEntry,
                                   glib_targets.len() as c_int,
                                   actions.to_glib());
        }
    }
```

src/target_entry.rs
EPashkin: Still public and no getters

Susurrus: Is there a reason these shouldn't be public? All manually implemented structs so far have used public fields.

EPashkin:
See discussion from https://github.com/gtk-rs/gtk/pull/472#issuecomment-287472293 (you first proposed that).

Susurrus:
@EPashkin Indeed, and the limited examples I came across online didn't show any examples doing that. I've added another commit making them private and adding getters.

src/target_entry.rs
```diff
@@ -23,6 +23,18 @@ impl TargetEntry {
             info: info,
         }
     }
+
+    pub fn get_target(&self) -> String {
```
EPashkin: -> &str ?

Susurrus: Yes, that's much nicer. Don't know why I didn't do that originally.

EPashkin:
About `drag_dest_set` you can't collect `e.to_glib_none().0` as this momently dropped.
I see this way:  collect stashes
```rust
let stashes: Vec<_> = targets.iter().map(|e| e.to_glib_none()).collect();
```
create empty `Vec` and use `set_len` to allocate memory, then iterate stashes and copy field values for all `GtkTargetEntry`s.
May be there better way but I don't see it now.

Anyway this function need be tested that it really works.

Also all generated code say that it `targets: &[&TargetEntry]` not sure what really need (it both compiled with same code)

Susurrus:
I've done another pass, adding enough functionality here that we can actually implement a complete drag-and-drop example (I'm going to file a PR for the examples repo shortly with WIP examples) so this work can be tested in its entirety. I still don't have everything working, but I think this is pretty close. I think the `gdk` work with `Atom`s is the weak link right now.

Gir.toml
```diff
@@ -1344,6 +1390,9 @@ status = "generate"
     name = "destroy-event"
     inhibit = true
     [[object.signal]]
+    name = "drag-data-get"
+    ignore = true #in/out parameter
```
EPashkin: Comment to separate line, please

src/widget.rs
```diff
@@ -34,12 +44,41 @@ pub trait WidgetExtManual {
...
+                                        t.as_mut_ptr() as *mut ffi::GtkTargetEntry,
+                                        t.len() as i32,
+                                        actions.to_glib())};
+    }
+
+    fn drag_source_set(&self, start_button_mask: ModifierType, targets: &[TargetEntry], actions: DragAction) {
+        let stashes: Vec<_> = targets.iter().map(|e| e.to_glib_none()).collect();
+        let mut t = Vec::with_capacity(stashes.len());
+        for stash in stashes {
+            t.push(stash.0.clone());
```
EPashkin: Strange that `clone` work: `GtkTargetEntry` IMHO don't implement.
If it works then you can use map-collect with cloning

src/widget.rs
```diff
@@ -34,12 +44,41 @@ pub trait WidgetExtManual {
...
+        }
+        unsafe { ffi::gtk_drag_source_set(self.to_glib_none().0,
+                                          start_button_mask.to_glib(),
+                                          t.as_mut_ptr() as *mut ffi::GtkTargetEntry,
```
EPashkin: `.as_ptr()` IMHO better

Susurrus: I did this because `gtk_drag_source_set` has a `mut` `targets` argument, but this actually is incorrect according to [the gtk docs](https://developer.gnome.org/gtk3/stable/gtk3-Drag-and-Drop.html#gtk-drag-source-set).

EPashkin: As you converting pointer in `unsafe` mutable source IMHO don't needed

src/widget.rs
Susurrus: It's just cloning the pointer I realized, so this is definitely wrong. I've pushed an updated version that copies all the `TargetEntry`s into new `ffi::GdkTargetEntry`s that I think is more correct, though I'm not certain if it's right. But it removes the null pointer problem revealed in gtk-rs/examples#116

Susurrus:
I'm having trouble trying to use this code and I'm not certain where the error is. If I try and use the new `Widget::drag_source_set()` function like: 
```
button.drag_source_set(gdk::MODIFIER_MASK,
                       &[],
                       gdk::ACTION_COPY);
```

I get datatype errors like:
```
error[E0308]: mismatched types
  --> src/drag_and_drop.rs:18:28
   |
18 |     button.drag_source_set(gdk::MODIFIER_MASK,
   |                            ^^^^^^^^^^^^^^^^^^ expected struct `gdk::auto::flags::ModifierType`, found struct `gdk::ModifierType`
   |
   = note: expected type `gdk::auto::flags::ModifierType`
              found type `gdk::ModifierType`
```

Anyone have any idea what this is actually saying? If I replace line 18 with `gdk::auto::flags::ModifierType`, I get `auto` is a private module. So I'm a little lost here.

EPashkin:
It again old manual duplicate https://github.com/gtk-rs/gdk/blob/master/src/enums.rs#L5-L27 that we forgot to remove

EPashkin:
No, it something else

Susurrus:
Regarding your point of enums, it does seem odd to me that there are two declarations of `MODIFIER_MASK`:
```
src/auto/mod.rs:pub use self::flags::MODIFIER_MASK;
src/enums.rs:    pub const ModifierMask: ffi::GdkModifierType = ffi::GDK_MODIFIER_MASK;
```

For my code above should I have used `gdk::enums::modifier_type::ModifierMask`?

EPashkin:
@Susurrus No `gdk::enums::modifier_type` must be not used.
Strange, I checked your code on examples/basic and compiles it both nightly and stable

Susurrus:
@EPashkin Yeah, looks like I had a bad path in my `Cargo.toml`. Alright, I'll get back to it, thanks!

src/widget.rs
```diff
@@ -2,35 +2,97 @@
...
 
     fn connect_unmap_event<F: Fn(&Self, &Event) -> Inhibit + 'static>(&self, f: F) -> u64;
 }
 
 impl<O: IsA<Widget> + IsA<Object>> WidgetExtManual for O {
+
+    fn drag_dest_set(&self, flags: DestDefaults, targets: &[TargetEntry], actions: DragAction) {
+        let stashes: Vec<_> = targets.iter().map(|e| e.to_glib_none()).collect();
+        let mut t = Vec::with_capacity(stashes.len());
+        for stash in stashes {
```
EPashkin: Found problem by accident: without `&stashes` it moved to for and dropped afterward.

src/widget.rs
```diff
@@ -2,35 +2,97 @@
...
+            unsafe {
+                t.push(ffi::GtkTargetEntry {
+                    target: (*stash.0).target.clone(),
```
EPashkin: clone unneeded

EPashkin:
IMHO `SelectionData::set_text` etc. don't need `&mut self`. While it still don't generate `received` signal, text changed. 
```
    [[object.function]]
    name = "set_text"
        [[object.function.parameter]]
        name = "selection_data"
        const = true
```
Also `Widget::drag-data-get` signal don't need be mut and manual

Susurrus:
Looking over the drag signals, I also think `Widget::drag-drop` needs to be set to `inhibit = true`, is that correct? I think it's for signals that return `gboolean` values, yes?

Susurrus:
Also `Widget::drag-motion` I believe

EPashkin:
You right for both

Susurrus:
Alright, I've updated the code as you've suggested and set those 2 signals to `inhibit`. Hopefully this is in line with your comments.

src/auto/selection_data.rs
```diff
@@ -93,9 +93,9 @@ impl SelectionData {
         }
     }
 
-    pub fn set_text(&mut self, str: &str, len: i32) -> bool {
+    pub fn set_text(&self, str: &str, len: i32) -> bool {
```
EPashkin: All others set need const too.

src/target_list.rs
```diff
@@ -0,0 +1,23 @@
...
+use glib::translate::*;
+use {TargetEntry, TargetList};
+
+impl TargetList {
+    pub fn new(targets: &[TargetEntry]) -> Self {
+        let targets_glib: Vec<_> = targets
+                                       .iter()
+                                       .map(|e| e.to_glib_none().0)
+                                       .collect();
+        let t_ptr: *mut ffi::GtkTargetEntry = if targets_glib.len() > 0 {
```
EPashkin: `.map(|e| e.to_glib_none().0)` will drop stash and collect bad pointers. See drag_dest_set

Susurrus:
Here's an updated version that addresses all of your points.

src/widget.rs
```diff
@@ -31,6 +38,55 @@ pub trait WidgetExtManual {
 }
 
 impl<O: IsA<Widget> + IsA<Object>> WidgetExtManual for O {
+
```
EPashkin: Remove this

EPashkin:
@Susurrus, rebase this to last master please
I don't see problem with code but way to use D&D still not found and this in WIP state?

EPashkin:
Seems I found problem: `from_glib_none` for boxed types use `copy` and you don't changed original.
So need use `from_glib_borrow` on `data`
```toml
    [[object.signal]]
    name = "drag-data-get"
        [[object.signal.parameter]]
        name = "data"
        transformation = "borrow"
```
By way, `drag_and_drop_textview.rs` example panic on windows as it produced wrong path
`filename: /D:/eap/rust/0/0.txt`

EPashkin:
From travis error this need `skip_assert_initialized!();`
```
$ ./check_init_asserts
src/target_list.rs: new(targets: &[TargetEntry]) -> Self 
```

Susurrus:
I think you did it @EPashkin! I've pushed the latest version which makes the examples work for me now. I think this does it!

EPashkin:
👍 Thanks, @Susurrus

GuillaumeGomez:
Awesome! Thanks @Susurrus!

PR #462, opened 2017-03-06
Implement StyleContext::{get_color, lookup_color}

Susurrus:
This is done manually because the RGBA struct isn't generally usable by `gir` yet.

Closes #454.

PR #473, opened 2017-03-21
Remove manual duplicate

EPashkin:
Removed 4 function that generated in #462 but manual is missed due wrongly non-optional https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-override-background-color


EPashkin:
Ping

GuillaumeGomez:
Thanks!

PR #476, opened 2017-03-23
Fix the Gitter badge

Bruflot:
The space in the URL breaks the markdown. See [here](http://i.imgur.com/efjLjvl.png).


GuillaumeGomez:
Thanks!

PR #478, opened 2017-03-27
Finish adding gdk::Atom support to GTK

Susurrus:
Still don't know why this type is necessary, but realized I needed to add it after needing it for #472.


src/auto/accel_group.rs
```diff
@@ -1,4 +1,4 @@
-// This file was generated by gir (b7f5189) from gir-files (71d73f0)
+// This file was generated by gir (b7f5189) from gir-files (71d73f0+)
```
EPashkin: Please, repeat Regen without +

Susurrus:
I believe this requires additional changes to `gdk` for the missing `FromGlibPtr*` impls I think, which should be trivial.

EPashkin:
Seems It was bad idea removing `impl FromGlibPtrNone<ffi::GdkAtom> for Atom {` :(

src/auto/accel_group.rs
Susurrus: I don't know what that `+` is from as my working directory is clean. Ran submodule update and the submodule is pointed to the correct git version. Manually checked out master and pulled to the latest version and it's still giving me that plus. You ever see that anywhere?

EPashkin: Remove gir.exe from target. Maybe it detect that no changes and not rebuild `gir_version.rs`

EPashkin: Or do some change and reset it

Susurrus: I deleted it and rechecked it out, tried deleting `gir-files/target` and rebuilt everything, nothing I have is really doing the trick here. I've seen this show `-dirty`, but never a `+` so I don't even know what that means.

src/auto/text_buffer.rs
```diff
@@ -274,9 +284,13 @@ impl TextBuffer {
         }
     }
 
-    //pub fn get_serialize_formats(&self) -> (/*Ignored*/Vec<gdk::Atom>, i32) {
-    //    unsafe { TODO: call ffi::gtk_text_buffer_get_serialize_formats() }
-    //}
+    pub fn get_serialize_formats(&self) -> (Vec<gdk::Atom>, i32) {
```
EPashkin: This function need ignored or implemented manually.
Test shows 1/6 and fails on printing second name.
```
let buffer = gtk::TextBuffer::new(None);
buffer.set_text("test");

let (formats, count) = buffer.get_serialize_formats();
println!("{}/{}", count, formats.len());
for format in formats {
    println!("{}", format.name());
}
```

src/auto/text_buffer.rs
```diff
@@ -165,9 +171,13 @@ impl TextBuffer {
     //    unsafe { TODO: call ffi::gtk_text_buffer_get_copy_target_list() }
     //}
 
-    //pub fn get_deserialize_formats(&self) -> (/*Ignored*/Vec<gdk::Atom>, i32) {
-    //    unsafe { TODO: call ffi::gtk_text_buffer_get_deserialize_formats() }
-    //}
+    pub fn get_deserialize_formats(&self) -> (Vec<gdk::Atom>, i32) {
```
EPashkin: Same as `get_serialize_formats` not sure how to get it return something

src/auto/text_buffer.rs
EPashkin: Updated test. Seems need use `from_glib_container_num` instead `from_glib_container`. Not checked this.

src/auto/accel_group.rs
EPashkin: `+` after hash mean that you `gir` dir not clean for git.

src/auto/text_buffer.rs
EPashkin: It can return something after `register_deserialize_tagset`

src/auto/accel_group.rs
EPashkin: Hash saved to `src/gir_version.rs` but sometimes build silently fails to do that (or think that nothing changed or just don't run `build.rs`)

EPashkin: Hm, I missed that + in gir-files version.
Check `git status` in `gir-files` directory: it must say `nothing to commit, working tree clean`
After this remove `gir/target` and try rebuild/

src/auto/selection_data.rs
```diff
@@ -21,9 +21,11 @@ impl SelectionData {
     //    unsafe { TODO: call ffi::gtk_selection_data_get_data() }
     //}
 
-    //pub fn get_data_type(&self) -> /*Ignored*/Option<gdk::Atom> {
-    //    unsafe { TODO: call ffi::gtk_selection_data_get_data_type() }
-    //}
+    pub fn get_data_type(&self) -> Option<gdk::Atom> {
```
EPashkin: @Susurrus please test any of these functions to be sure that `from_glib_none` works on atom.

EPashkin: Sorry, tested it on results `register_serialize_tagset` all works.

EPashkin:
Also seems [register_serialize_tagset](https://developer.gnome.org/gtk3/stable/GtkTextBuffer.html#gtk-text-buffer-register-serialize-tagset), register_deserialize_tagset, [get_target](https://developer.gnome.org/gtk3/stable/gtk3-Selections.html#gtk-selection-data-get-target), get-selection and get_data_type has non-optional return

src/auto/text_buffer.rs
Susurrus: I don't quite understand this, but it sounds like I should set the function to ignore during generation.

EPashkin: Better both ignore and add it as manual with other conversion

Susurrus:
I've updated things and now I think we're closer. For those functions erronously returning an `Option<_>` type, I just ignored them. Those should be addressed by changes to `gir` or `gir-files`.

EPashkin:
I mean not ignoring non-optional functions but explain it to gir like in https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L325-L326

Susurrus:
Thanks! I figured there had to be a way to set things like this in the `Gir.toml`. Here's the latest, but I didn't deal with `Clipboard:get_(de)serialize_formats()` yet (just set them as ignored) as I'm not positive what they should return. From reading them I think they should just return a `Vec<gdk::Atom>` as the length doesn't need to be returned as a separate object. If that's the case, then I can implement it manually or should this be something done by `gir`?

EPashkin:
Currently `Clipboard:get_(de)serialize_formats()` need manual implementation. Agree that it need return just vector.

Susurrus:
Alright, I think that resolves all issues you've brough up @EPashkin. Note that CI won't succeed until gtk-rs/gdk#148 is merged.

src/text_buffer.rs
```diff
@@ -4,16 +4,31 @@
...
 use glib::signal::connect;
 use glib_ffi;
 use TextBuffer;
 use TextIter;
 
 impl TextBuffer {
+    pub fn get_deserialize_formats(&self) -> Vec<gdk::Atom> {
+        unsafe {
+            let mut n_formats = mem::uninitialized();
+            FromGlibPtrContainer::from_glib_container(ffi::gtk_text_buffer_get_deserialize_formats(self.to_glib_none().0, &mut n_formats))
```
EPashkin: As I says it need be `from_glib_container_num`

Susurrus: Then what should the `num` argument be when calling it? The total number is returned by the type not specified by the caller, which is why I left it as is. But the docs of `from_glib_container_num` are sparse.

EPashkin: It number of elements in array. In this case `n_formats`

Susurrus:
Sorry, @EPashkin, looks like I was a little too tired, I understand your point about `get_(de)serialize_formats()`. This should be ready to roll now!

EPashkin:
@Susurrus, Thanks, It seems good.
I check with https://github.com/gtk-rs/gtk/pull/478#discussion_r108222841 just in case in 3 hour.
Then we need wait until @GuillaumeGomez have free time, merged https://github.com/gtk-rs/gdk/pull/148 and restart CI.

GuillaumeGomez:
@EPashkin: gtk-rs/gdk#148 needs some modifications so we'll have to wait a bit. ;)

EPashkin:
Strange but `get_deserialize_formats` still return too big vector that panic on iterate.

GuillaumeGomez:
Restarting CI.

EPashkin:
Very strange implementation in https://github.com/gtk-rs/glib/blob/master/src/translate.rs#L886-L937
for *_num functions. `Num` used only for capacity.
Seems this need fix too.

EPashkin:
And others too 😭

EPashkin:
All CI finished. glib path don't change signatures so can't fail build.

GuillaumeGomez:
Great, then I merge. Thanks @Susurrus!

PR #480, opened 2017-03-31
Fix place_on_monitor version

EPashkin:
It don't have version in gir https://developer.gnome.org/gtk3/stable/GtkMenu.html#gtk-menu-place-on-monitor


GuillaumeGomez:
Hum, strange... Thanks!

PR #481, opened 2017-03-31
Add missing child properties for Box

antoyo:
The example will follow soon as a pull request of the examples repository.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -352,6 +352,18 @@ status = "generate"
 name = "Gtk.Box"
 status = "generate"
     [[object.child_prop]]
+    name = "expand"
+    type = "gboolean"
+    [[object.child_prop]]
+    name = "fill"
+    type = "gboolean"
+    [[object.child_prop]]
+    name = "pack-type"
+    type = "Gtk.PackType"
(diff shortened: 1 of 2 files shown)


EPashkin:
Box already has [set_child_packing](https://github.com/gtk-rs/gtk/blob/master/src/auto/box_.rs#L60) and [query_child_packing](https://github.com/gtk-rs/gtk/blob/master/src/auto/box_.rs#L50) so IMHO this properties unneeded.

antoyo:
The reason why I added these is that I created a crate called [`relm`](https://github.com/antoyo/relm) and this allows to set child properties individually like [in this example](https://github.com/antoyo/relm/blob/master/examples/buttons-child-attribute.rs#L89-L95).
This code generates the call to `set_child_fill` for instance.

EPashkin:
Ok, then I don't against adding these

EPashkin:
But rebuild it with Gir b7f5189 or add update gir submodule here

antoyo:
I regenerated with this revision.

GuillaumeGomez:
Thanks a lot!

Issue #495, opened 2017-04-28
Add child properties for gtk::Notebook


antoyo:
Hello.
I would like to add the child properties for `gtk::Notebook`:
I tried the following:
```toml
    [[object.child_prop]]
    name = "detachable"
    type = "gboolean"
    [[object.child_prop]]
    name = "menu-label"
    type = "gchar*"
    [[object.child_prop]]
    name = "position"
    type = "gint"
    [[object.child_prop]]
    name = "reorderable"
    type = "gboolean"
    [[object.child_prop]]
    name = "tab-expand"
    type = "gboolean"
    [[object.child_prop]]
    name = "tab-fill"
    type = "gboolean"
    [[object.child_prop]]
    name = "tab-label"
    type = "gchar*"
```
However, it does not for the `gchar*` type since I get the following error:
```
ERROR:gir::analysis::child_properties: Bad type `gchar*` of child property `menu-label` for `Notebook`
ERROR:gir::analysis::child_properties: Bad type `gchar*` of child property `tab-label` for `Notebook`
```

It is for the same reason I added the child properties for `gtk::Box`: it is to ease the use in my `relm` crate.

Thanks to add these properties.

EPashkin:
Instead "char*" use "utf8"

EPashkin:
But as https://github.com/gtk-rs/gtk/issues/392 says Notebook already has getters-setters as function.
So I against duplicates, better do only separating obsolete `set_tab_label_packing`(`tab-expand`, `tab-fill`)
and `position` as https://developer.gnome.org/gtk2/stable/GtkNotebook.html#gtk-notebook-set-tab-pos is different

antoyo:
The thing is, [`relm`](https://github.com/antoyo/relm) allow to set child properties as you can see in [this example](https://github.com/antoyo/relm/blob/master/examples/child-prop-attribute.rs#L59-L65).
This example uses the child properties that I added in the pull request https://github.com/gtk-rs/gtk/pull/481.
That was merged even thought it is a duplicate of [`set_child_packing`](http://gtk-rs.org/docs/gtk/trait.BoxExt.html#tymethod.set_child_packing).

Having child properties for `Notebook` would allow to do this in `relm`:
```rust
view! {
    gtk::Notebook {
        gtk::Button {
            tab: {
                tab_label: "First Button", // This calls set_child_tab_label().
            },
            label: "+",
        }
    }
}
```
[...]

PR #486, opened 2017-04-09
Partial regen

EPashkin:
cc @GuillaumeGomez
Include only changed files as there will be more regen by https://github.com/gtk-rs/gir/pull/321 or/and https://github.com/gtk-rs/gir/pull/328

PR #496, opened 2017-04-29
Add child properties for gtk::Notebook

antoyo:


GuillaumeGomez:
@EPashkin: I have issues with duplicates in here but the fact is that at least we have a more unified API. So from this point of view, I *think* it's acceptable.

EPashkin:
@antoyo There to many unneeded changes, please, regenerate it with Gir bf7bd49
As @GuillaumeGomez accept this and it compiles I accept despite objections in #495

antoyo:
@EPashkin Actually, there are different git hashes in the repository, right now.
Look at these two files:
https://github.com/gtk-rs/gtk/blob/master/src/auto/about_dialog.rs
https://github.com/gtk-rs/gtk/blob/master/src/auto/app_chooser_dialog.rs
If I use the git hash you mentioned, there'll be more modifications.

EPashkin:
If you build Git from commit bf7bd49f50755a8d1204a3d13b0e8a743a79f458
only 2 files changed: Gir.toml and src/auto/notebool.rs

EPashkin:
On windows sometimes build.rs don't run and gir_version.rs don't changed. It can fixed with touching Cargo.toml. IMHO it not issue on linux
But anyway if you build with right Gir then will no other changes (with the exception of first 2 lines)

EPashkin:
Also you can try do `make` in gtk root ;)

antoyo:
I've fixed it.
By the way, if you don't want the users to use these new methods (for the reasons you mentioned), perhaps we could add `#[doc(hidden)]` on top of them.

EPashkin:
Now I understand why there to many files changes: I do partial regen in #486 as we plan do full regen later.
I forgot that, sorry.
Please remove other files from commit

EPashkin:
Currently no way adding #[doc(hidden)] but it was good idea

GuillaumeGomez:
And could be quite easy to add.

EPashkin:
Added gir issue.

GuillaumeGomez:
Let's make the change in gir first and then please update your code in here @antoyo. :)

EPashkin:
As do regen after to next version of gir will do too many changes IMHO better
Just add `doc_hidden=true` here in Gir.toml and merge this PR

GuillaumeGomez:
@EPashkin: Good point.

@antoyo: Can you update please? :)

EPashkin:
Too many hidden docs ;)
@antoyo Please unhide position and  tab-label as actually don't have equal direct function

GuillaumeGomez:
I'll be out for a few hours. See you later!

antoyo:
@EPashkin I believe the `tab-label` can already be set by [`set_tab_label_text`](http://gtk-rs.org/docs/gtk/struct.Notebook.html#method.set_tab_label_text) and `position` by [`reorder_child`](http://gtk-rs.org/docs/gtk/struct.Notebook.html#method.reorder_child).

EPashkin:
@antoyo You right.
@GuillaumeGomez 👍 for merge

GuillaumeGomez:
Thanks to both of you! :)

PR #498, opened 2017-04-29
Use pango objects

### LINKED ISSUE #497: TextTag's style property is wrapped around as if it is a bool
It should be an enum, and means you can't tag text as italic

EPashkin:
Fix #497


EPashkin:
CI finished

GuillaumeGomez:
Thanks!

Issue #501, opened 2017-05-03
Problem with optional parameters


EPashkin:
@GuillaumeGomez There runtime problem with https://github.com/gtk-rs/gir/pull/321: this example shows label with emply or bad text
```rust
fn main() {
    if gtk::init().is_err() {
        println!("Failed to initialize GTK.");
        return;
    }

    let label = gtk::Label::new(Some("test text"));
    let window = gtk::Window::new(gtk::WindowType::Toplevel);
    window.set_title("Example");
    window.add(&label);
    window.show_all();

    // GTK & main window boilerplate
    window.connect_delete_event(|_, _| {
                                    gtk::main_quit();
                                    Inhibit(false)
                                });
    gtk::main();
}
```
IMHO this due don't assigning stash to temporary variable

GuillaumeGomez:
I'll look into this issue as soon as possible.

EPashkin:
@GuillaumeGomez CIs passed for all linked PRs.

GuillaumeGomez:
All merged, thanks! :D

EPashkin:
Seems its time to publish next version.
Only https://github.com/gtk-rs/gtk/issues/488 from my list still unsolved.

GuillaumeGomez:
Do you want to wait for it or not?

EPashkin:
IMHO waiting uneeded.

GuillaumeGomez:
Ok, I'll start the release in the next hours. Can you prepare the merged PRs list in this release please? :)

EPashkin:
For all: Option conversion by you

Sys:
https://github.com/gtk-rs/sys/pull/45 Use FFI types for constants, string ends with '/0', bitflags update.  By sdroege
https://github.com/gtk-rs/sys/pull/45 Make all struct fields public, etc. By sdroege
GLib:
https://github.com/gtk-rs/glib/pull/153 Fix array of string full transfer
https://github.com/gtk-rs/glib/pull/154 Fix FromGlibPtrContainer implementations
https://github.com/gtk-rs/glib/pull/155 Convert string constants from FFI
https://github.com/gtk-rs/glib/pull/156 Directly access the GType in the GValue instead of doing raw pointer casting By sdroege
https://github.com/gtk-rs/glib/pull/162 Add support for getting a &str from a GValue without copying
By sdroege

GuillaumeGomez:
@EPashkin: Ok, all PRs ready for the first round. Once I have your approval, I merge and publish.

[...]

PR #502, opened 2017-05-06
Regen

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/501
Application https://github.com/gtk-rs/gir/pull/361


GuillaumeGomez:
Shouldn't you do this regen with the newly pushed version of gir? (so you update the gir submodule at the same time.)

EPashkin:
It reason why it was WIP

PR #504, opened 2017-05-06
Update release_process.md

GuillaumeGomez:


GuillaumeGomez:
@EPashkin: So, for the master branch, should we be equal with the crate branch's version or ahead of one?

EPashkin:
IMHO it need be equal and changed before merging to crate.
Offtopic, also I don't see reason having dependency version in Cargo.toml of master branch: in docs only noncrate use of `version` in `path` override

GuillaumeGomez:
I just thought: I will write a full release script because right now, it just too long and easy to make mistakes.

EPashkin:
We have version in dependencies https://github.com/gtk-rs/gtk/blob/master/Cargo.toml#L47 but seems cargo only used it to check max version. And I only found use `version` on `path` dependency http://doc.crates.io/specifying-dependencies.html#specifying-path-dependencies.

EPashkin:
Long forgotten ;)
Thanks.

GuillaumeGomez:
Indeed haha.

PR #509, opened 2017-05-15
Regenerate autogenerated bindings with latest GIR

sdroege:
Also remove the GIO prelude re-export as discussed in https://github.com/gtk-rs/gir/pull/364#issuecomment-301147922


EPashkin:
1. Update Gir submodule, better in separate commit before regen.
2. Many files in `src` need traits with reexport it in prelude: ex. https://github.com/gtk-rs/gtk/blob/master/src/assistant.rs -> AssistantExtManual
3. Doctest errors need fix in https://raw.githubusercontent.com/gtk-rs/lgpl-docs/master/gtk/docs.md
with something like `# use LabelExt;`

sdroege:
Ok, thanks!

sdroege:
Is the plan/intention to add *all* traits to the prelude? Should do that in the other pull requests too then for consistency at least.

EPashkin:
As currently all object traits reexported in prelude adding manual trait need for consistency for classes that both manual and auto. Not sure that it need for only manual classes.
Gio don't have manual part for objects (only for record `Resource`)
Pango and Gdk seems also need change.
Pango currently don't have prelude and don't have manual+auto classes, so IMHO only prelude with reexport from auto::trait need be added.
Gdk need manual trait at minimum for DragContext

EPashkin:
*forgot to say: Currently prelude reexport all trait, that user can use. And IMHO it good.

sdroege:
Ok, so if I see this correctly nothing is need about the prelude. GIR submodule is updated.
And for doctests, it seems you fixed lgpl-docs earlier already?

Anything missing here or in the other pull requests? :)

EPashkin:
https://github.com/gtk-rs/gtk/pull/509#issuecomment-301440175

At minimum please add manual trait for all classes that also generated.

EPashkin:
and gdk too.

sdroege:
> At minimum please add manual trait for all classes that also generated.

I don't understand what you mean. You want manual traits to be added for every automatically generated trait? Why?

EPashkin:
Some classes (ex. Assistant) has both generated and manual code.
After you moved all methods to trait in generated code, you need do it to manual part too.
Or for future children will have "access" only to generated traits method as it have implementation for IsA but not for manual part that implemented only for parent (with exception of constructors and class functions - this need implemented on parent directly).

sdroege:
Ah yes of course, thanks for pointing that out :) Will do

sdroege:
Should all be good now

src/application.rs
```diff
@@ -1,12 +1,22 @@
...
+pub trait ApplicationExtManual {
     #[cfg(feature = "v3_6")]
-    pub fn new(application_id: Option<&str>, flags: ApplicationFlags) -> Result<Application, ()> {
+    fn new(application_id: Option<&str>, flags: ApplicationFlags) -> Result<Application, ()>;
+
+    #[cfg(not(feature = "v3_6"))]
+    fn new(application_id: &str, flags: ApplicationFlags) -> Result<Application, ()>;
+}
+
+impl<O: IsA<Application> + IsA<glib::object::Object>> ApplicationExtManual for O {
```
EPashkin: `+ IsA<glib::object::Object>` unneeded here and possible in any others

EPashkin:
assistant.rs, tree_model_filter.rs, tree_store.rs not changed.
Also I see old bug: none of manual trait is reexported on crate level gtk, 
IMHO just `pub use prelude::*;` in `lib.rs` will fix it.

sdroege:
Done. However prelude::* is also not exported from gdk and pango.

EPashkin:
No actual problem in pango as only manual trait `Matrix` reexported explicitly.
But gdk really has same problem.

👍 Thanks, @sdroege

GuillaumeGomez:
Thanks!

sdroege:
>  But gdk really has same problem.

Great, re-exported prelude::* from gdk too in https://github.com/gtk-rs/gdk/pull/163

PR #515, opened 2017-05-26
Add SourceView

GuillaumeGomez:
This PR isn't working, for now the structs aren't generated because the sys functions aren't generated.

@EPashkin: It seems the correponding sys functions are missing. Any idea why?


EPashkin:
Because it don't part of gtk but separate library https://packages.ubuntu.com/zesty/amd64/libdevel/libgtksourceview-3.0-dev 😉

GuillaumeGomez:
Ah damn, thanks!

PR #525, opened 2017-06-06
Change gtk::Application::new parameter to use Into<Option<&str>>

jeandudey:


EPashkin:
Thanks, can you do same for others in `src` (dialog, entry_buffer, etc.) ?

GuillaumeGomez:
@EPashkin: slow down, don't scare newcomers! :p
@jeandudey: you forgot to add the lifetime:

```rust
 fn new<I: Into<Option<&str>>>(application_id: I, flags: ApplicationFlags) -> Result<Application, glib::BoolError> {
// should be:
 fn new<'a, I: Into<Option<&'a str>>>(application_id: I, flags: ApplicationFlags) -> Result<Application, glib::BoolError> {
```

EPashkin:
Also we missed it on previous PR but both function need `pub`

GuillaumeGomez:
Oh damn indeed!

jeandudey:
@GuillaumeGomez i fixed it and made the functions there public.

GuillaumeGomez:
Thanks!

Susurrus:
Was there a reason the `Into<Option<&str>>` was skipped for the pre-3.6 version of `Application::new()`?

GuillaumeGomez:
It's because it wasn't generated automatically.

Susurrus:
No, I mean there are two `Application::new()` functions. Only 1 was "fixed" in this PR to support the `Into<Option<&str>>`. Was there a reason the second one wasn't updated as well?

EPashkin:
May be because `application_id` is non-optional ;)

GuillaumeGomez:
The other one isn't an `Option<&str>` so why would it need the `Into`?

Susurrus:
You're right, sorry about that!

Issue #488, opened 2017-04-16
Radio buttons problem


EPashkin:
Any functions of `RadioButton` that accepting `group: &[RadioButton]` not works in right way.
IMHO this due recreating GSList when passing it to ffi.
Bad cases:
```rust
let radio_size_10 = gtk::RadioButton::new_with_label(&[], "10x10");
let radio_group = radio_size_10.get_group();
let radio_size_20 = gtk::RadioButton::new_with_label(&radio_group, "20x20");

let radio_size_10 = gtk::RadioButton::new_with_label(&[], "10x10");
let radio_size_20 = gtk::RadioButton::new_with_label(&[], "20x20");
let group = vec![radio_size_10.clone(), radio_size_20.clone()];
radio_size_10.set_group(&group);
radio_size_20.set_group(&group);
```
Current good case:
```rust
let radio_size_10 = gtk::RadioButton::new_with_label(&[], "10x10");
let radio_size_20 = gtk::RadioButton::new_with_label_from_widget(&radio_size_10, "20x20");
let radio_size_30 = gtk::RadioButton::new_with_label(&[], "30x30");
radio_size_30.join_group(Some(&radio_size_10));
```

I propose ignore all these functions and create manual functions:
```rust
pub fn new() -> RadioButton;
pub fn new_with_label(label: &str) -> RadioButton;
pub fn new_with_mnemonic(label: &str) -> RadioButton;
```
and change first parameter in *_from_widget to nonoptional

EPashkin:
Thanks @kbsymanz for noting that.

EPashkin:
Other variant:
Ignore bad functions and create manual functions for all `*_from_widget` without `_from_widget` in name
```rust
    pub fn new(radio_group_member: Option<&RadioButton>) -> RadioButton {
        assert_initialized_main_thread!();
        unsafe {
            Widget::from_glib_none(ffi::gtk_radio_button_new_from_widget(radio_group_member.to_glib_none().0)).downcast_unchecked()
        }
    }
```

EPashkin:
Renaming variant will results bad docs.

EPashkin:
Third option that has minimal affect on docs: just ignore bad functions

EPashkin:
@GuillaumeGomez Can you look at this issue again?
@sdroege Can you look about this too?

GuillaumeGomez:
Well, for now the only solution is indeed to ignore these functions and write them ourselves. Except if you have something in mind?

[...]

PR #529, opened 2017-06-10
Replace functions for RadioButton and RadioMenuItem

EPashkin:
Closes https://github.com/gtk-rs/gtk/issues/488
Using tests:
https://github.com/EPashkin/gtk-rs-runtime-tests/blob/radiobutton/tests/radio_button.rs
https://github.com/EPashkin/gtk-rs-runtime-tests/blob/radiobutton/tests/radio_menu_item.rs
cc @GuillaumeGomez, @sdroege

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -312,6 +312,11 @@ status = "generate"
         name = "surface"
         const = true
     [[object.function]]
+    pattern = "tree_[gs]et_row_drag_data"
+        [[object.function.parameter]]
+        name = "selection_data"
+        const = true
+    [[object.function]]
     name = "selection_add_targets"
     #array as pointer
     ignore = true
--- a/Gir.toml
+++ b/Gir.toml
@@ -132,8 +132,6 @@ generate = [
     "Gtk.PrintQuality",
     "Gtk.ProgressBar",
     "Gtk.PropagationPhase",
-    "Gtk.RadioButton",
-    "Gtk.RadioMenuItem",
     "Gtk.RecentChooser",
     "Gtk.RecentChooserDialog",
     "Gtk.RecentChooserError",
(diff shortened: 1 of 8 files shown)


EPashkin:
@GuillaumeGomez, @sdroege please check it.

GuillaumeGomez:
Seems good for me, however I never used the radio groups so I'd prefer to rely on @sdroege on this.

sdroege:
_I never used the radio groups API either, at least not in the last 10 years, and don't really work with GTK_

This looks good to me and should work fine, however it seems potentially confusing to users as the normal constructors now don't provide a way to define a group and you have to use the ones with `_widget` instead. Maybe implement that ones with the GSList group manually, so that the actual GSList is passed there and not a NULL pointer?

EPashkin:
Problem with GSList changed on every `g_slist_prepend` so it almost impossible make it actual when it changed in C side

sdroege:
I was thinking of just getting the GSList* from one of the buttons in the list (the first I guess?), probably making sure that all of the buttons in the list are in the same group (or maybe not?), and then passing it to the constructor.

EPashkin:
Then we need ensure that user don't use list after it, or we again have wrong list (need change to consume group parameter).
I don't see it really more usable
```
let rd1 = new_with_label(None, "a");
let gr = rd1.get_group();
let rd2 = new_with_label(gr, "b");
let gr = rd2.get_group();
let rd3 = new_with_label(gr, "c");
```
that realized in this PR
```
let rd1 = new_with_label("a");
let rd2 = new_with_label("b");
let rd3 = new_with_label("c");
rd2.join_group(&rd1);
rd3.join_group(&rd1);
```

sdroege:
I was thinking of never passing the GSList* to the user, but using it internally in the constructor, e.g.

```Rust
    pub fn new(group: &[RadioMenuItem]) -> RadioMenuItem {
        unsafe {
            let c_group = ffi::gtk_radio_menu_item_get_group(to_glib_none(group[0])); // check if there is actually group[0] or if it's empty
            // possibly: check if all in group have the same group, or make it so
            Widget::from_glib_none(ffi::gtk_radio_menu_item_new(c_group)).downcast_unchecked()
        }
    }
```

But I missed the `join_group` API, I guess it's sufficient like that then :)

EPashkin:
Hm. this way also doable (but need some checks for empty list). I add it to issue.

EPashkin:
@sdroege You agree to merge PR as is?
IMHO arrayed version while doable, confuses the user: usage `&[]` for first member, need collect members to array or use `&[firstmember]` to adding.

EPashkin:
Ping

sdroege:
Fine with me, yes

GuillaumeGomez:
Good for me too. Thanks!

PR #543, opened 2017-07-12
Regenerate with latest gir

sdroege:
See https://github.com/gtk-rs/glib/pull/196


EPashkin:
No changes in gir submodule

PR #552, opened 2017-08-07
Fix clippy warnings

GuillaumeGomez:
Seems strange but didn't have any warnings or anything... So let's see with travis.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -9,7 +9,7 @@
 //!
 //! The library is a work in progress: expect missing bindings and breaking
 //! changes. A steadily increasing share of the code is machine-generated from
-//! GObject introspection metadata. The API docs were converted from the
+//! `GObject` introspection metadata. The API docs were converted from the
 //! upstream ones so until they've all been reviewed there will be incongruities
 //! with actual Rust APIs.
 //!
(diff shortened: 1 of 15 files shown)


EPashkin:
If was for `cargo clippy`

GuillaumeGomez:
Ah we don't run it on travis? I thought we did... Fixing warnings then.

EPashkin:
Clippy still don't shipped with rust (but planned) and still need nightly, so we can't add it to stables.

GuillaumeGomez:
Oh that's not an issue. I added it to some of my projects which test both stable and nightly (I'm thinking about sysinfo and process-viewer). I just test clippy and nightly and don't call on stable. Nothing magical but very convenient.

GuillaumeGomez:
Updated.

src/lib.rs
```diff
@@ -144,6 +144,11 @@
 //! them once. **Omitting them in the following cargo invocations will not undo
 //! their effects!**
 
+#![cfg_attr(feature = "cargo-clippy", allow(new_without_default))]
+#![cfg_attr(feature = "cargo-clippy", allow(type_complexity))]
+#![cfg_attr(feature = "cargo-clippy", allow(borrowed_box))]
```
EPashkin: We fixed this warning for generated code, please fix it in manual too.

EPashkin:
@GuillaumeGomez, interesting idea, maybe really apply it on gtk too?

GuillaumeGomez:
Sure, I'll add it to this PR.

GuillaumeGomez:
And here you go. We now have cargo clippy check on nightly.

EPashkin:
@GuillaumeGomez Thanks, 👍 with changing PR title.

PR #566, opened 2017-08-27
Check fixed

### LINKED ISSUE #565: Segfault on Fixed get_child_x
Hi!

The `gtk::Fixed` `get_child_x` segfaults if the user provides a widget which does not belong to the `Fixed` widget. Here is an example code:

```rust

extern crate gtk;
use gtk::prelude::*;

fn main() {
    if gtk::init().is_err() {
        println!("Failed to initialize GTK.");
        return;
    }

    let window = gtk::Window::new(gtk::WindowType::Toplevel);
    let container = gtk::Fixed::new();

    let button1 = gtk::Button::new_with_label("Button 1");
    let button2 = gtk::Button::new_with_label("Button 2");
    container.add(&button1);

    // Segfault Here
    let _ = container.get_child_y(&button2);

    window.set_title("Test program");

    window.connect_delete_event(|_, _| {
        gtk::main_quit();
        Inhibit(false)
    });

    window.add(&container);
    window.show_all();

    gtk::main();
}
```

Maybe a panic should be done if the native gtk returns null pointer ?

GuillaumeGomez:
Fixes #565.

cc @EPashkin


EPashkin:
I prefer look C code first and fix upstream, but currently don't have time.
Also not sure that need fix ASAP as it case wrong use.

GuillaumeGomez:
Wrong use but it shouldn't segfault, it should panic. Huge difference! But if you have better, please do. :)

Otherwise we're not into a hurry so it can wait.

EPashkin:
check_init_asserts failed
also `src/fixed.rs` don't have `mod fixed` in lib, and no trait reexport in `prelude`

GuillaumeGomez:
Ah damn, I forgot to add the missing files.

EPashkin:
Also child properties for "Gtk.Fixed" need be commented

EPashkin:
*in Gir.toml

GuillaumeGomez:
It's done in the update as well. ;)

src/fixed.rs
```diff
@@ -0,0 +1,90 @@
...
+fn has_widget<O: IsA<Fixed> + IsA<Container>, T: IsA<Widget>>(c: &O, item: &T) -> bool {
+    skip_assert_initialized!();
+    let t: *mut ffi::GtkContainer = c.to_glib_none().0;
+    let t: Container = unsafe { from_glib_none(t) };
+    let widgets = t.get_children();
+    let this: *mut ffi::GtkWidget = item.to_glib_none().0;
+    widgets.iter().find(|w| {
+        let w: *mut ffi::GtkWidget = w.to_glib_none().0;
+        w == this
+    }).is_some()
```
EPashkin: Please change to this to remove unneeded conversions
```
unsafe {
    let stash = c.to_glib_none();
    let t: *mut ffi::GtkContainer = stash.0;
    let glist = ffi::gtk_container_get_children(t);
    let found = !glib_ffi::g_list_find(glist, item.to_glib_none().0 as _).is_null();
    glib_ffi::g_list_free(glist);
    found
}
```

src/fixed.rs
```diff
@@ -0,0 +1,90 @@
...
+    let container = Fixed::new();
+
+    let button1 = Button::new_with_label("Button 1");
+    let button2 = Button::new_with_label("Button 2");
+    container.add(&button1);
+
+    let _ = container.get_child_y(&button1);
+    // Should panic here.
+    let _ = container.get_child_y(&button2);
+}
```
EPashkin: This test will panic on `init` on travis, so better removed it too.
I will add test to https://github.com/EPashkin/gtk-rs-runtime-tests.

src/fixed.rs
GuillaumeGomez: That's the first code I wrote but I thought we might want to avoid C code as much as possible. But if you prefer as is, fine by me! :D

src/fixed.rs
GuillaumeGomez: Oh nice!

src/fixed.rs
EPashkin: Hm, missed that `stash` and `t` unneeded here and can be replaced with direct `c.to_glib_none().0`

EPashkin:
And please add link to https://github.com/gtk-rs/gtk/issues/565 somewhere in `fixed.rs`

GuillaumeGomez:
Updated.

src/fixed.rs
```diff
@@ -0,0 +1,72 @@
...
+use glib_ffi;
+use glib::translate::*;
+
+use Container;
+use Fixed;
+use IsA;
+use Value;
+use Widget;
+
+
```
EPashkin: Maybe unneeded extra line

EPashkin:
👍  after CI passed

GuillaumeGomez:
All updated for good this time!

serpilliere:
Thank you!

PR #581, opened 2017-10-21
Regen

GuillaumeGomez:
cc @EPashkin


GuillaumeGomez:
(there is no difference so I think we should just not merge this though)

EPashkin:
Thanks

EPashkin:
Agree that it can just merge without waiting CI

GuillaumeGomez:
So we merge it even though nothing change?

EPashkin:
If you want, but IMHO it unneeded,
so better just close PR or leave it for dox fixes

GuillaumeGomez:
Then let's close it!

PR #584, opened 2017-10-24
Add RadioToolButton

pizzaiter:
This PR adds RadioToolButton. The implementation is very similar to that of RadioButton.


GuillaumeGomez:
Thanks!

cc @EPashkin @sdroege

src/radio_tool_button.rs
```diff
@@ -0,0 +1,23 @@
...
+use ToolItem;
+use ffi;
+use glib::object::Downcast;
+use glib::translate::*;
+
+impl RadioToolButton {
+    pub fn new() -> RadioToolButton {
+        assert_initialized_main_thread!();
+        unsafe {
+            ToolItem::from_glib_none(ffi::gtk_radio_tool_button_new(ptr::null_mut())).downcast_unchecked()
```
sdroege: Why `ToolItem` and not `RadioToolButton`? And generally, you can just use `from_glib_none` and let the compiler figure out the right implementation

src/radio_tool_button.rs
```diff
@@ -0,0 +1,23 @@
...
+        }
+    }
+
+    pub fn new_from_stock(stock_id: &str) -> RadioToolButton {
+        assert_initialized_main_thread!();
+        unsafe {
+            ToolItem::from_glib_none(ffi::gtk_radio_tool_button_new_from_stock(ptr::null_mut(), stock_id.to_glib_none().0)).downcast_unchecked()
```
sdroege: This one is deprecated since 3.10, maybe let's not add it? Docs say that `gtk_radio_tool_button_new()` should be used instead.

Also it probably makes sense to expose *something* for the weird groups API it provides? IIRC we already have the same problem with other group API elsewhere in GTK ( @GuillaumeGomez you remember? ) so let's take inspiration from there.

src/radio_tool_button.rs
GuillaumeGomez: Oh good catch, didn't see it at all...

src/radio_tool_button.rs
GuillaumeGomez: I don't recall which at all. :-/

sdroege: @EPashkin Maybe? Some other RadioWhatever API (the plain RadioButton? RadioMenuItems?) where you have groups of things. I don't have time for looking closer now and GTK UI stuff is not really my main area of expertise so I would have to look up, sorry :)

EPashkin: You meant https://github.com/gtk-rs/gtk/issues/488?
Problem with this class that it don't have `join_group` function but it has `group` property that maybe works same and better deignored.
I will check this PR more later.

src/radio_tool_button.rs
EPashkin: It because says Gtk-3.0.gir:
```xml
        <return-value transfer-ownership="none">
          <doc xml:space="preserve">The new #GtkRadioToolButton</doc>
          <type name="ToolItem" c:type="GtkToolItem*"/>
        </return-value>

```

EPashkin:
Problem that `get_group` return vector of `RadioButton`: I don't see way get this button from `RadioToolButton` as C function `_gtk_tool_button_get_button` is private, but maybe this not real problem.

Tested that `group` property works same as `join_group`.
Generated code looks like this:
```
    fn set_property_group(&self, group: Option<&RadioToolButton>) {
        unsafe {
            gobject_ffi::g_object_set_property(self.to_glib_none().0, "group".to_glib_none().0, Value::from(group).to_glib_none().0);
        }
    }
```
Only minor problem with it that it need `Option`, and we prefer use `Into<Option<&RadioToolButton>>` to more simply usage.
I propose ignore property `group` (as now) and add manual version of `join_group` with `Into<Option` parameter same as other radio types and use this code in it.

Not sure about `_from_stock` but if it works for @pizzaiter, then it can stay.

src/radio_tool_button.rs
pizzaiter: I thought it would be best to be consistent with what gets auto-generated for similar functions, even if it does not really make a difference how one calls `from_glib_none`.

src/radio_tool_button.rs
pizzaiter: I actually missed that the `from_stock` functions are deprecated. I can remove them if you want. I am fine either way.

For the group API, I will try to implement what @EPashkin proposed below. At the moment one has to use one of the `new_from_widget` functions.

pizzaiter:
I pushed a commit that adds `join_group` as @EPashkin suggested.

EPashkin:
@pizzaiter Thanks.
About `ToolItem::from_glib_none` you can't replace this as ffi function returns `ToolItem*`

@GuillaumeGomez, @sdroege 👍 for merge from me

GuillaumeGomez:
For me as well, so let's wait for @sdroege's confirmation.

sdroege:
I have no further comments, go ahead :)

GuillaumeGomez:
Thanks then!

pizzaiter:
:tada:

PR #588, opened 2017-11-04
Regen

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/480

cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
I think that with this one merged, we're all good. Thanks a lot @EPashkin!

PR #599, opened 2017-11-27
Typed properties

EPashkin:
Part of https://github.com/gtk-rs/gir/pull/496

cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
:+1: 

We should really display the type that aren't listed in the `Gir.toml`.

EPashkin:
There still will be too many types.

PR #602, opened 2017-12-03
Add support for file chooser buttons

demurgos:
This commit adds support for file chooser buttons. They are represented
as a slice of tuples `(text, action)`. The original function uses
varargs to support an arbitrary number of arguments. This implementation
supports up to 3 buttons.

~~**This is a breaking change.**~~

See: https://developer.gnome.org/gtk3/stable/GtkFileChooserDialog.html#gtk-file-chooser-dialog-new

### DIFF
--- a/src/file_chooser_dialog.rs
+++ b/src/file_chooser_dialog.rs
@@ -5,21 +5,83 @@
 use ffi;
 use glib::translate::*;
 use glib::object::{Downcast, IsA};
+use libc::c_char;
 use std::ptr;
 use FileChooserAction;
 use FileChooserDialog;
+use ResponseType;
 use Widget;
 use Window;
 
 impl FileChooserDialog {
-    pub fn new<T: IsA<Window>>(title: Option<&str>, parent: Option<&T>, action: FileChooserAction)
-            -> FileChooserDialog {
+    // TODO: Keep the other constructor with buttons support as the only constructor (this one was
+    //       left for compatibility) and rename it to `new` for consistency.
+    pub fn new<T: IsA<Window>>(title: Option<&str>, parent: Option<&T>, action: FileChooserAction) -> FileChooserDialog {
         assert_initialized_main_thread!();
         unsafe {
-            Widget::from_glib_none(
-                ffi::gtk_file_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0,
-                    action.to_glib(), ptr::null_mut()))
-                .downcast_unchecked()
+            Widget::from_glib_none(ffi::gtk_file_chooser_dialog_new(
+                title.to_glib_none().0,
+                parent.to_glib_none().0,
+                action.to_glib(),
+                ptr::null::<c_char>()
+            )).downcast_unchecked()
+        }
+    }
+
+    pub fn with_buttons<T: IsA<Window>>(title: Option<&str>, parent: Option<&T>, action: FileChooserAction, buttons: &[(&str, ResponseType)]) -> FileChooserDialog {
+        assert_initialized_main_thread!();
+        unsafe {
+            Widget::from_glib_none(match buttons.len() {
+                0 => {
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        ptr::null::<c_char>()
+                    )
+                },
+                1 => {
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        buttons[0].0.to_glib_none().0,
+                        buttons[0].1.to_glib(),
+                        ptr::null::<c_char>(),
+                    )
+                },
+                2 => {
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        buttons[0].0.to_glib_none().0,
+                        buttons[0].1.to_glib(),
+                        (buttons[1].0.to_glib_none() as Stash<*const c_char, str>).0,
+                        buttons[1].1.to_glib(),
+                        ptr::null::<c_char>(),
+                    )
+                },
+                3 => {
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        buttons[0].0.to_glib_none().0,
+                        buttons[0].1.to_glib(),
+                        (buttons[1].0.to_glib_none() as Stash<*const c_char, str>).0,
+                        buttons[1].1.to_glib(),
+                        (buttons[2].0.to_glib_none() as Stash<*const c_char, str>).0,
+                        buttons[2].1.to_glib(),
+                        ptr::null::<c_char>(),
+                    )
+                },
+                _ => {
+                    // TODO: Support arbitrary number of buttons once variadic functions are supported.
+                    //       See: https://github.com/rust-lang/rust/issues/44930
+                    panic!(format!("`FileChooserDialog::with_buttons` does not support 4+ buttons, received {}", buttons.len()))
+                }
+            }).downcast_unchecked()
         }
     }
 }


src/file_chooser_dialog.rs
```diff
@@ -5,20 +5,31 @@
...
         assert_initialized_main_thread!();
         unsafe {
-            Widget::from_glib_none(
-                ffi::gtk_file_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0,
-                    action.to_glib(), ptr::null_mut()))
+            Widget::from_glib_none(match buttons.len() {
+                0 => ffi::gtk_file_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0, action.to_glib(), ptr::null::<c_char>()),
+                1 => ffi::gtk_file_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0, action.to_glib(), buttons[0].0.to_glib_none().0, buttons[0].1.to_glib(), ptr::null::<c_char>()),
+                2 => {
+                    let second_button_text: *const c_char = buttons[1].0.to_glib_none().0;
```
EPashkin: This won't work.
You need save `Stash` to variable and use `.0` in function call.

src/file_chooser_dialog.rs
```diff
@@ -5,20 +5,31 @@
...
+                    ffi::gtk_file_chooser_dialog_new(title.to_glib_none().0, parent.to_glib_none().0, action.to_glib(), buttons[0].0.to_glib_none().0, buttons[0].1.to_glib(), second_button_text, buttons[1].1.to_glib(), ptr::null::<c_char>())
+                },
+                _ => {
```
sdroege: This hard-coding of 4 different settings for the file chooser is also not ideal, and the usage of the enum prevents adding "custom" buttons to there. Not sure how to fix either of those though, the C API is kind of suboptimal in this regard.

EPashkin:
To remove "breaking change" you can just add this as "with_buttons" instead.
IMHO better panic or unimplemented for 4+ buttons.
Also there crate https://crates.io/crates/va_list

src/file_chooser_dialog.rs
demurgos: Thanks, I'll fix it.

demurgos:
Thank you for the helpful reviews. I am not very familiar with these bindings but I'll fix these issues.
- [ ] Save stash
- [ ] Use `with_buttons` to prevent breaking change (ideally, there should only be one function: there is just one function in Gtk)
- [ ] Use `va_list` or panic if the number of buttons is too high.

sdroege:
The `va_list` crate only supports x86, x86-64 and arm, that's not a good idea. We'd need https://github.com/rust-lang/rfcs/pull/2137 for solving this properly.

EPashkin:
@sdroege `ResponseType` still contains `__Unknown(i32)` so IMHO custom buttons is possible.

demurgos:
Ok, so I'll panic for now.

EPashkin:
@demurgos Please add TODO comment with link to https://github.com/rust-lang/rust/issues/44930

src/file_chooser_dialog.rs
demurgos: Fixed in the latest commit

src/file_chooser_dialog.rs
demurgos: Fixed in the latest commit: instead of silently truncating the buttons, it panics. This is a temporary solution until variadic functions are supported.

demurgos:
@EPashkin @sdroege  I updated my PR following your comments. Could you take a look at it?

EPashkin:
Please, remove this comment, I will add it to right place (https://github.com/gtk-rs/lgpl-docs) after PR merged
```
    /// This function creates a `FileChooserDialog` with buttons:
    /// ```
    /// let dialog = FileChooserDialog::with_buttons::<ApplicationWindow>(
    ///   Option::Some("Open File"),
    ///   Option::None,
    ///   FileChooserAction::Open,
    ///   &[("_Cancel", ResponseType::Cancel), ("_Open", ResponseType::Accept)]
    /// );
    /// ```
```

demurgos:
@EPashkin Done, I remove the doc comment.

EPashkin:
@demurgos I will change doc' example to this. It works for you?
```
let dialog = FileChooserDialog::with_buttons::<Window>(
  Some("Open File"),
  None,
  FileChooserAction::Open,
  &[("_Cancel", ResponseType::Cancel), ("_Open", ResponseType::Accept)]
);
```

demurgos:
Yes, this is fine. I was using `ApplicationWindow` and just copy pasted, but you are right that it is better to simply use `Window`.

EPashkin:
cc @sdroege and @GuillaumeGomez for final decision.

src/file_chooser_dialog.rs
```diff
@@ -5,21 +5,91 @@
...
+                        ptr::null::<c_char>()
+                    )
+                },
+                1 => {
+                    let first_button_text: Stash<*const c_char, str> = buttons[0].0.to_glib_none();
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        first_button_text.0,
```
sdroege: Why not just `buttons[0].0.to_glib_none().0` here?

sdroege:
Looks good to me except for the comment I just added. Feel free to ignore me on that though if there's a reason

GuillaumeGomez:
I wrote a [va_list](https://crates.io/crates/va_list-rs) crate which works on every OS/arch. It's used by a VLC demuxer so I suppose it's fine enough. Maybe give it a look?

src/file_chooser_dialog.rs
demurgos: For consistency with the other branches.

To be honest, I also do not really understand how Stash works

demurgos:
@sdroege 
I replied to your comment. Is it fine to leave stashes for consistency or do I have to remove it. If I have to remove it, do I have to remove every `first_button_text` variable or just the one in the "one button" case?

sdroege:
If it's not needed, it seems simpler/shorter to not define new variables for the Stash but just directly use the value

demurgos:
@sdroege 
I removed the variables for the text of the first button. I tested it locally once for 1, 2 and 3 buttons (it works fine).

It is still not clear for me why I can omit the stash for the first button but if I try to do it for the others it breaks.

demurgos:
The CI error seems to be code by the usage example. Travis tries to compile it even if it's not a complete program.

sdroege:
How does it break for the other buttons? It should work in all cases

demurgos:
I get the wrong text, instead of printing the text of the button 2, it reuses the window title as button labels. I'll try to reproduce it, it may have been an error on my side.

demurgos:
I was not able to reproduce my issue (wrong text when not using the Stash) so I assume the error was on my side. I pushed the version without any local variables.

src/file_chooser_dialog.rs
```diff
@@ -5,21 +5,85 @@
...
+                    )
+                },
+                2 => {
+                    ffi::gtk_file_chooser_dialog_new(
+                        title.to_glib_none().0,
+                        parent.to_glib_none().0,
+                        action.to_glib(),
+                        buttons[0].0.to_glib_none().0,
+                        buttons[0].1.to_glib(),
+                        (buttons[1].0.to_glib_none() as Stash<*const c_char, str>).0,
```
EPashkin: `Stash` IMHO don't needed here and in other places

EPashkin: Forgot that it `var_arg` :( so rust can't infer type

demurgos: Exactly, I can omit it for the first text but not for the other ones.

sdroege:
Seems good now?

demurgos:
It's good for me.

demurgos:
@EPashkin  @sdroege 
Is it possible to advance with this PR?

sdroege:
All good from my side

src/file_chooser_dialog.rs
```diff
@@ -5,21 +5,85 @@
...
+                        ptr::null::<c_char>(),
+                    )
+                },
+                _ => {
+                    // TODO: Support arbitrary number of buttons once variadic functions are supported.
+                    //       See: https://github.com/rust-lang/rust/issues/44930
+                    panic!(format!("`FileChooserDialog::with_buttons` does not support 4+ buttons, received {}", buttons.len()))
+                }
+            })
+              .downcast_unchecked()
```
GuillaumeGomez: Please put this call on the previous line, right after `})`.

demurgos: Fixed in latest commit.

demurgos:
@GuillaumeGomez 
Done, I squashed the commits, it should be ready now 😄

GuillaumeGomez:
Just waiting confirmation from @EPashkin.

EPashkin:
@demurgos Thanks.
@GuillaumeGomez LGFM

PR #611, opened 2017-12-28
Add WidgetExt::add_tick_callback binding

fengalin:


EPashkin:
Please add `gtk_widget_remove_tick_callback` too
`FnMut` really needed?

fengalin:
> Please add `gtk_widget_remove_tick_callback` too

`gtk_widget_remove_tick_callback` is auto-generated in: https://github.com/gtk-rs/gtk/blob/342f69e640383afc29e602c2e20b933e5493837a/src/auto/widget.rs#L445

> `FnMut` really needed?

Not sure about this. I followed the example of: https://github.com/gtk-rs/gtk/blob/342f69e640383afc29e602c2e20b933e5493837a/src/signal.rs#L46

EPashkin:
Ok, I missed both.
Then I only see problem that callback added for concrete type `O` but called only for `Widget`, so better make both `trampoline` and `destroy` internal generic function (something like https://github.com/gtk-rs/gio/blob/master/src/buffered_input_stream.rs#L24)
cc @sdroege too.

src/widget.rs
```diff
@@ -98,6 +111,45 @@ impl<O: IsA<Widget> + IsA<Object>> WidgetExtManual for O {
         }
     }
 
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    fn add_tick_callback<F>(&self, func: F) -> u32
+    where
+        F: FnMut(&Self, &FrameClock) -> Continue + 'static,
+    {
+        unsafe extern "C" fn add_tick_callback_trampoline(
```
EPashkin: Both internal function need be generalized by `O: IsA<Widget> + IsA<Object>` (maybe without `Object`)

fengalin: Ah Ok! Sorry about that.

fengalin:
Please see the commit's message. I had to use `Widget` instead of `Self` for the callback. Hope this is acceptable. This is the way is work for e.g.:
https://github.com/sdroege/gstreamer-rs/blob/d44869711a958b2e9fe479e4dc0e1c56c2753cc0/gstreamer/src/pad.rs#L105

src/widget.rs
sdroege: Why? All the other signal connectors use `&Self` too

EPashkin:
@fengalin Solution was it end of `widget.rs`.
This change at minimum compiles.
```diff
@@ -114,37 +114,35 @@ impl<O: IsA<Widget> + IsA<Object>> WidgetExtManual for O {
     #[cfg(any(feature = "v3_8", feature = "dox"))]
     fn add_tick_callback<F>(&self, func: F) -> u32
     where
-        F: FnMut(&Widget, &FrameClock) -> Continue + 'static,
+        F: FnMut(&Self, &FrameClock) -> Continue + 'static,
     {
-        type Callback = FnMut(&Widget, &FrameClock) -> Continue + 'static;
-        
-        unsafe extern "C" fn add_tick_callback_trampoline(
+        unsafe extern "C" fn add_tick_callback_trampoline<O: IsA<Widget>> (
             widget: *mut ffi::GtkWidget,
             frame_clock: *mut gdk_ffi::GdkFrameClock,
             func: gpointer,
         ) -> gboolean {
             callback_guard!();
-            let func: &RefCell<Box<Callback>> = transmute(func);
+            let func: &RefCell<Box<FnMut(&O, &FrameClock) -> Continue + 'static>> = transmute(func);
 
             (&mut *func.borrow_mut())(
-                &from_glib_borrow(widget),
+                &Widget::from_glib_borrow(widget).downcast_unchecked(),
                 &from_glib_borrow(frame_clock)
             ).to_glib()
         }
 
-        unsafe extern "C" fn destroy_closure(ptr: gpointer) {
+        unsafe extern "C" fn destroy_closure<O: IsA<Widget>>(ptr: gpointer) {
             callback_guard!();
-            Box::<RefCell<Box<Callback>>>::from_raw(ptr as *mut _);
+            Box::<RefCell<Box<FnMut(&O, &FrameClock) -> Continue + 'static>>>::from_raw(ptr as *mut _);
         }
 
-        let func: Box<RefCell<Box<Callback>>> = Box::new(RefCell::new(Box::new(func)));
+        let func: Box<RefCell<Box<FnMut(&O, &FrameClock) -> Continue + 'static>>> = Box::new(RefCell::new(Box::new(func)));
                 
         unsafe {
             ffi::gtk_widget_add_tick_callback(
                 self.to_glib_none().0,
-                Some(add_tick_callback_trampoline),
+                Some(add_tick_callback_trampoline::<O>),
                 Box::into_raw(func) as gpointer,
-                Some(destroy_closure),
+                Some(destroy_closure::<O>),
             )
         }
     }
```

fengalin:
@EPashkin , sorry didn't see your message as I was working on it. What was missing was `downcast_unchecked()`.

EPashkin:
@fengalin Now it all good for me.

@sdroege How about you? I think that your comment about `&Self` was from `&Widget` version.

sdroege:
Looks good to me now, yes. But I just noticed that basically all signal handlers in gtk can be FnMut :)

fengalin:
Yes indeed :)

EPashkin:
@GuillaumeGomez Please take a look at this PR.

src/widget.rs
```diff
@@ -2,14 +2,20 @@
 // See the COPYRIGHT file at the top-level directory of this distribution.
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
+#[cfg(any(feature = "v3_8", feature = "dox"))]
```
GuillaumeGomez: We should maybe write a macro for things like this. It's getting a bit ugly when reading the code...

GuillaumeGomez:
Thanks!

PR #613, opened 2018-01-14
Add signal Overlay::get-child-position

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/234

cc @GuillaumeGomez, @sdroege, @vhakulinen

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -9,6 +9,7 @@ pub use glib::signal::Inhibit;
 use glib::SourceId;
 use glib::translate::*;
 use glib::signal::SignalHandlerId;
+use gdk::Rectangle;
 
 use glib_ffi::{self, gboolean, gpointer};
 
@@ -16,6 +17,7 @@ use {
     Continue,
     ScrollType,
     SpinButton,
+    Widget,
 };
 
 // idle_add and timeout_add fixed to the main thread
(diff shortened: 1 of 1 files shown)


src/signal.rs
```diff
@@ -293,3 +295,67 @@ mod spin_button {
...
+        T: IsA<Overlay> + IsA<Object>,
+    {
+        callback_guard!();
+        let f: &&(Fn(&T, &Widget) -> Option<Rectangle> + 'static) = transmute(f);
+        match f(
+            &Overlay::from_glib_borrow(this).downcast_unchecked(),
+            &from_glib_borrow(widget),
+        ) {
+            Some(rect) => {
+                ptr::write(allocation, transmute(rect));
```
EPashkin: Maybe `write_unaligned` need used instead?

sdroege: No, the rectangle destination is required to be properly aligned. It's otherwise a bug in the code that emitted the signal.

But why is the transmute necessary? Don't we have the normal translation traits implemented for Rectangle? Otherwise looks all good

src/signal.rs
```diff
@@ -293,3 +295,67 @@ mod spin_button {
...
+                GTRUE
+            }
+            None => GFALSE,
```
sdroege: Maybe want to let the match return true/false and then call .to_glib() on the overall result?

src/signal.rs
EPashkin: `*allocation = *rect.to_glib_none().0` will fails with `error[E0507]: cannot move out of dereference of raw pointer`,
using dereference instead `transmute` will produce same error.
Currently no direct conversion to `GtkRectangle` other that `to_glib_none`.

src/signal.rs
EPashkin: You right, I will change this later

src/signal.rs
sdroege: `ptr::write(allocation, rect.to_glib_none().0)`?

EPashkin: Thanks, I missed that possibility.

EPashkin:
Updated

sdroege:
👍

GuillaumeGomez:
Good for me as well, thanks! Just waiting for CI before merging.

PR #616, opened 2018-01-19
Regen

GuillaumeGomez:
cc @EPashkin @sdroege (since you didn't seem very confident on the generated code, you can take a look here 😉 )


EPashkin:
LGFM, but please update gir-files too, it not on last version

GuillaumeGomez:
`gir-files` is up-to-date apparently? I merge then!

PR #617, opened 2018-01-23
Expose gtk_clipboard_set_with_data.

tmiasko:
Rust implementation exposes only GtkClipboardGetFunc as customization
point. GtkClipboardClearFunc is used implicitly to clean up the provided
closure.

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -1,0 +1,58 @@
+// Copyright 2018, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use Clipboard;
+use SelectionData;
+use TargetEntry;
+use ffi;
+use glib::object::IsA;
+use glib::translate::*;
+use glib_ffi::gpointer;
+use libc::c_uint;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+
+pub trait ClipboardExtManual {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, u32) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool;
+}
+
+impl<O: IsA<Clipboard>> ClipboardExtManual for O {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, u32) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool {
+        let stashed_targets: Vec<_> = targets.iter().map(|e| e.to_glib_none()).collect();
+        let mut t = Vec::with_capacity(stashed_targets.len());
+        for stash in &stashed_targets {
+            unsafe {
+                t.push(ffi::GtkTargetEntry {
+                    target: (*stash.0).target,
+                    flags: (*stash.0).flags,
+                    info: (*stash.0).info,
+                });
+            }
+        }
+        let t_ptr: *mut ffi::GtkTargetEntry = t.as_mut_ptr();
+        let f: Box_<Box_<Fn(&Clipboard, &SelectionData, u32) + 'static>> = Box_::new(Box_::new(f));
+        let user_data = Box_::into_raw(f) as *mut _;
+        let success : bool = unsafe { from_glib(ffi::gtk_clipboard_set_with_data(self.to_glib_none().0,
+                                             t_ptr, t.len() as c_uint, 
+                                             Some(trampoline), Some(cleanup), user_data))
+        };
+        if !success {
+            // Cleanup function is not called in case of a failure.
+            unsafe { Box_::<Box_<Fn(&Clipboard, &SelectionData, u32) + 'static>>::from_raw(user_data as *mut _); }
+        }
+        success
+    }
+}
+
+unsafe extern "C" fn trampoline(clipboard: *mut ffi::GtkClipboard, selection_data: *mut ffi::GtkSelectionData, info: c_uint, user_data: gpointer) {
+    callback_guard!();
+    let f: &&(Fn(&Clipboard, &SelectionData, u32) + 'static) = transmute(user_data);
+    f(&from_glib_borrow(clipboard), &from_glib_borrow(selection_data), info);
+}
+
+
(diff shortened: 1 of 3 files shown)


src/clipboard.rs
```diff
@@ -0,0 +1,58 @@
...
+use libc::c_uint;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+
+pub trait ClipboardExtManual {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, c_uint) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool;
+}
+
+impl<O: IsA<Clipboard>> ClipboardExtManual for O {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, c_uint) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool {
```
GuillaumeGomez: I think we never directly expose C-like types (I'm referring to `c_uint` in here).

tmiasko: OK, I will replace this with u32, which is the same type as in TargetEntry::info field.
Those two are supposed to go together.

GuillaumeGomez: Use the `from_glib` function to make the conversion.

tmiasko: @GuillaumeGomez you refer to u32 -> c_uint conversion? As far as I can see
there seems to be no suitable FromGlib implementation. Either way those types
are exactly the same on all platforms supported by Rust so far.

GuillaumeGomez: I thought it did exist. It's been a while I didn't write bindings by hand so I'm maybe completely wrong...

cc @sdroege @EPashkin

src/clipboard.rs
```diff
@@ -0,0 +1,58 @@
...
+use libc::c_uint;
+use std::boxed::Box as Box_;
+use std::mem::transmute;
+
+pub trait ClipboardExtManual {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, u32) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool;
+}
+
+impl<O: IsA<Clipboard>> ClipboardExtManual for O {
+    fn set_with_data<F: Fn(&Clipboard, &SelectionData, u32) + 'static>(&self, targets: &[TargetEntry], f: F) -> bool {
```
sdroege: This seems more like a `set`. The with data part is an implementation detail of how the bindings implement it. No data is provided by the caller, the bindings use the data for passing a closure

src/clipboard.rs
```diff
@@ -0,0 +1,58 @@
...
+        }
+        let t_ptr: *mut ffi::GtkTargetEntry = t.as_mut_ptr();
+        let f: Box_<Box_<Fn(&Clipboard, &SelectionData, u32) + 'static>> = Box_::new(Box_::new(f));
+        let user_data = Box_::into_raw(f) as *mut _;
+        let success : bool = unsafe { from_glib(ffi::gtk_clipboard_set_with_data(self.to_glib_none().0,
+                                             t_ptr, t.len() as c_uint, 
+                                             Some(trampoline), Some(cleanup), user_data))
+        };
+        if !success {
+            // Cleanup function is not called in case of a failure.
```
sdroege: What an ugly API :)

src/clipboard.rs
```diff
@@ -0,0 +1,58 @@
...
+                    target: (*stash.0).target,
+                    flags: (*stash.0).flags,
+                    info: (*stash.0).info,
+                });
+            }
+        }
+        let t_ptr: *mut ffi::GtkTargetEntry = t.as_mut_ptr();
+        let f: Box_<Box_<Fn(&Clipboard, &SelectionData, u32) + 'static>> = Box_::new(Box_::new(f));
+        let user_data = Box_::into_raw(f) as *mut _;
+        let success : bool = unsafe { from_glib(ffi::gtk_clipboard_set_with_data(self.to_glib_none().0,
```
sdroege: Is the closure only called from here or might it also be called after the function returned? In case of the former the static bound is not needed, and boxing neither.

src/clipboard.rs
tmiasko: Essentially it configures a callback that provides clipboard data when
requested to do so at some point in the future. 

Regarding the name, maybe set_data_provider, set_provider, or set like you have
suggested. Though, it seems to me that either way you will have to describe its
behaviour in terms of gtk_clipboard_set_with_data, so changing the name doesn't 
seem to be of much help. I would rather remain with current one.

src/clipboard.rs
tmiasko: It won't be called immediately. Only after the function returns.

src/clipboard.rs
sdroege: This seems to be wrong.
>  clear_func | when the clipboard contents are set again, this function will be called, and get_func will not be subsequently called.

It is called whenever the contents are set again. Our function that we set can be called again if I don't misunderstand.

I think the correct way of doing this is by using the `set_with_owner` function. Then you can have a GObject subclass that has reference counting and could store the closure.

You would however have to create a new GObject for that from Rust, which is rather boring and verbose. See the first chapters [here](https://coaxion.net/blog/2017/09/exporting-a-gobject-c-api-from-rust-code-and-using-it-from-c-python-javascript-and-others/), the others are not required for this simple case.

src/clipboard.rs
sdroege: I guess `set_data` is fine, or `set_with_closure`

src/clipboard.rs
sdroege: The docs of `GtkClipboardClearFunc` are more useful here
> A function that will be called when the contents of the clipboard are changed or cleared. Once this has called, the user_data_or_owner argument will not be used again.

So this should be fine after all

EPashkin:
Oh, I forgot about this PR.
LGFM but seems @sdroege want function rename (docs will be lost in this case: IMHO not big problem).

sdroege:
I'm fine with keeping it as-is, it's all bad options so let's keep it as in C :)

EPashkin:
@GuillaumeGomez Then this PR ready for merge it

GuillaumeGomez:
Ok! Thanks @tmiasko!

PR #618, opened 2018-01-24
Unignore gtk_container_child_set_property and gtk_container_child_get_property.

tmiasko:


EPashkin:
Normally we generate child properties with Gir. see https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L401-L409
Any other reason to generate function?

tmiasko:
Thanks, that is even better. I completely missed those.

PR #622, opened 2018-02-08
Fix TargetList::new implementation

### LINKED ISSUE #621: Incorrect conversion of &[TargetEntry] in TargetList::new.
Currently each entry is converted to glib independently, and then address of
one of them is passed to gtk_target_list_new. See https://github.com/gtk-rs/gtk/blob/d3fb35605254f794c4cf669f49d0cdfe09dfc9dc/src/target_list.rs#L10 for details. It will crash when passed more than one TargetEntry.

On the other hand, implementation in https://github.com/gtk-rs/gtk/blob/1efdfd8d9a9bafcd0c4f5c87816bbecc3930a1eb/src/widget.rs#L53
appears to be doing the correct thing.

EPashkin:
Fix #621 

cc @GuillaumeGomez , @sdroege , @tmiasko

### DIFF
--- a/src/target_list.rs
+++ b/src/target_list.rs
@@ -7,14 +7,8 @@ use {TargetEntry, TargetList};
 impl TargetList {
     pub fn new(targets: &[TargetEntry]) -> Self {
         skip_assert_initialized!();
-        let stashes: Vec<_> = targets
-                                  .iter()
-                                  .map(|e| e.to_glib_none())
-                                  .collect();
-        let mut t = Vec::with_capacity(stashes.len());
-        for stash in &stashes {
-            t.push(stash.0);
-        }
+        let stashes: Vec<_> = targets.iter().map(|e| e.to_glib_none()).collect();
+        let t: Vec<_> = stashes.iter().map(|stash| unsafe { *stash.0 }).collect();
         let t_ptr: *mut ffi::GtkTargetEntry = if !t.is_empty() {
             t.as_ptr() as *mut _
         } else {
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
Less code, seems like a good thing to me. :)

sdroege:
:+1:

EPashkin:
@sdroege There strange gio-related error on travis added in https://github.com/gtk-rs/gio/pull/78
```
error[E0433]: failed to resolve. Could not find `windows` in `os`
  --> /home/travis/.cargo/git/checkouts/gio-a4b825b310a073b9/37c2a08/src/socket.rs:18:14
   |
18 | use std::os::windows::io::{IntoRawSocket, FromRawSocket};
   |              ^^^^^^^ Could not find `windows` in `os`
```

EPashkin:
Oh, it on doc building, gio don't have it so we missed.
Seems we need declare fake traits `IntoRawSocket, FromRawSocket` in case `dox` and not `windows`?

GuillaumeGomez:
Maybe there's a way to make gir directly generate this as well?

EPashkin:
cc @GuillaumeGomez as it doc related

EPashkin:
Gir generated functions with wrong array type
```rust
impl TargetList {
    pub fn new(targets: &[&TargetEntry]) -> TargetList {
        let ntargets = targets.len() as u32;
        unsafe {
            from_glib_full(ffi::gtk_target_list_new(targets.to_glib_none().0, ntargets))
        }
    }
```

sdroege:
> Oh, it on doc building, gio don't have it so we missed.
Seems we need declare fake traits IntoRawSocket, FromRawSocket in case dox and not windows?

That seems annoying. I thought the point of dox was that it can build things for any platform

GuillaumeGomez:
It should actually.

EPashkin:
@GuillaumeGomez please restart travis here

GuillaumeGomez:
Yes sir!

EPashkin:
@GuillaumeGomez This ready for merge

GuillaumeGomez:
Thanks!

PR #623, opened 2018-02-10
Remove gir-files from crate

EPashkin:
Part of https://github.com/gtk-rs/glib/issues/289

cc @GuillaumeGomez


GuillaumeGomez:
Oh good idea!

Issue #637, opened 2018-03-17
GLArea segfaults when dropped


anniryynanen:
GLAreas don't seem to be getting dropped cleanly, this example segfaults when the window is closed:

```
extern crate gtk;

use gtk::prelude::*;
use gtk::{GLArea, Window, WindowType};

fn main() {
    if gtk::init().is_err() {
        println!("Failed to initialize GTK.");
        return;
    }

    let window = Window::new(WindowType::Toplevel);
    let gl_area = GLArea::new();
    window.add(&gl_area);
    window.show_all();

    window.connect_delete_event(|_, _| {
        gtk::main_quit();
        Inhibit(false)
    });

    gtk::main();
}
```

EPashkin:
For me it not segfaults but writes:
```
GLib-GObject-CRITICAL **: g_object_unref: assertion 'G_IS_OBJECT (object)' failed
```
can be solved `std::mem::forget(gl_area);` after `gtk::main();` but this sure wrong way to fix.

tmiasko:
We don't assume ownership of GLArea returned gtk_gl_area_new, because gir says
that we are supposed to own it, which is not true as it returns floating
reference.

GuillaumeGomez:
Fixed so closing it.

PR #643, opened 2018-04-04
Use correct transfer annotation for gtk_gl_area_new.

tmiasko:
This regen that applies changes from https://github.com/gtk-rs/gir-files/pull/19 and fixes issue #637.
The actual change is single word diff in src/auto/gl_area.rs.


EPashkin:
@tmiasko Thanks.
@GuillaumeGomez 👍 for this too

tmiasko:
Single word vs Files changed 208. It seems like we should start using single version file.

EPashkin:
@tmiasko One month passed since https://github.com/gtk-rs/gio/pull/92, maybe it's time to add gtk as second test subject.
Want do it in this PR?

tmiasko:
Yeah, lets do it here.

EPashkin:
@tmiasko Thanks, now we have much more changes here.
@GuillaumeGomez 👍

GuillaumeGomez:
Thanks!

Issue #646, opened 2018-04-11
TextBuffer: connect_insert_text closure should take `&mut Iter`


jpathy:
From the docs:
> Note that if your handler runs before the default handler it must not invalidate the location iter (or has to revalidate it). The default signal handler revalidates it to point to the end of the inserted text.

So we may want to revalidate the iter, if we do some mutation on the buffer. Currently not possible without `unsafe`.

GuillaumeGomez:
I suppose we can close this issue now.

PR #647, opened 2018-04-18
Make location in TextBuffer insert-text callback mutable.

tmiasko:
Make it possible to revalidate the iterator position with `iter.assign(&new_iter)`
in the case the callback has to make additional changes to buffer.

Fixes issue #646.

Warning: It is also possible to write `*iter = new_iter`,
but that does something completely different than assign.

### DIFF
--- a/src/text_buffer.rs
+++ b/src/text_buffer.rs
@@ -16,13 +16,13 @@ use TextBuffer;
 use TextIter;
 
 pub trait TextBufferExtManual {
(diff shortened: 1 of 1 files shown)


EPashkin:
👍
Travis error on nightly very strange, but even if it real, only can be fixed in glib.

tmiasko:
The errors about void* casting is somewhat useless (and there is already a
clippy issue about that).  Other cases of under aligned pointers seems to be
related to how interfaces are generated right now, for example:
`pub struct GtkActionable(c_void);`.

Of course neither is actually related to changes done in this PR.

GuillaumeGomez:
Thanks!

PR #648, opened 2018-04-21
Regenerate

sdroege:
With this all regens done


Cargo.toml
```diff
@@ -47,6 +48,8 @@ git = "https://github.com/gtk-rs/lgpl-docs"
 libc = "0.2"
 bitflags = "1.0"
 lazy_static = "1.0"
+send-cell = "0.1"
```
EPashkin: `send-cell` optional too ?

sdroege: Indeed, fixed

EPashkin:
IMHO better always add to futures minimal version current module too:
glib\gio 2_36, gdk\gtk 3_10, something like v2_28 for gdk_pixbuf and v1_32_4 for pango.

sdroege:
> IMHO better always add to futures minimal version current module too:
glib\gio 2_36, gdk\gtk 3_10, something like v2_28 for gdk_pixbuf and v1_32_4 for pango.

I don't understand what you mean

EPashkin:
And also maybe better not always repeat `"glib/futures", "gio/futures"` but use minimal list of direct childs
(Ex. IMHO `gio\futures` includes `glib\futures`),
but that required add `futures` to gdk too as it works currently main version limiter for gio, pango, cairo and gdk-pixbuf,
and gtk depends of gdk version only

EPashkin:
Example for gio "futures = ["futures-core", "futures-channel", "futures-util", "glib/futures"]"
internally it set `glib/v2_36` for glib, but don't set same version for gio itself, so functional from v2_36 disabled.
For gio it seems no problem with this actually, so maybe I just overthinking

sdroege:
OK, I'll update later. The version should not matter for anything, it's only required in glib

sdroege:
@GuillaumeGomez This should be good to go too now. The travis failure on nightly looks problematic though, we should investigate that. Must be a problem in sys though

GuillaumeGomez:
Wo, that's quite freaky! I think we'll need to add another step to make our conversions work without having rustc yelling about it...

GuillaumeGomez:
Let's go for it then.

sdroege:
The errors are caused by clippy, and false warnings. The alignment ones at least.

PR #651, opened 2018-04-29
Fix present method on macos

GuillaumeGomez:
cc @EPashkin @sdroege


EPashkin:
@sdroege Examples failed build second time. Any idea?
```
error: custom attribute panicked
  --> src/bin/gio_futures_await.rs:23:1
   |
23 | #[async]
   | ^^^^^^^^
   |
   = help: message: unsupported character _
```

sdroege:
Bug in nightly of today, see https://github.com/rust-lang-nursery/futures-rs/issues/992

sdroege:
Can you explain why this is necessary and correct? And please link to the relevant GTK bug in Bugzilla, or create one if none exists. This looks very ugly and needs at least an explanation in the code

EPashkin:
@sdroege So we can ignore it now, Thanks.

GuillaumeGomez:
> Can you explain why this is necessary and correct? And please link to the relevant GTK bug in Bugzilla, or create one if none exists. This looks very ugly and needs at least an explanation in the code

Because on OSX, this function doesn't bring the window to the foreground whereas it should. As simple as that. I found the bug in bugzilla but can't it again...

sdroege:
please find it again so we can properly link it here :) It's important so that we can remove this ugly workaround once we depend on a new enough GTK, and so our futures selfes can later understand how this disaster of code happened

GuillaumeGomez:
For reference, bug is referenced [here](https://bugzilla.gnome.org/show_bug.cgi?id=782029).

Issue #653, opened 2018-05-02
Cross Compilation issues


foo-jin:
I'm in a similar predicament to #600, trying to cross compile from Ubuntu to Windows. I followed the same steps as the author of #600 did, made sure that my glade file used the same version of gtk as specified in the `gtk` crate's `version` feature flag (I'm using `v3.10`). The error I end up with is `gtk-0.4.1/src/auto/enums.rs:5432: undefined reference to 'gtk_text_view_layer_get_type'`. It could imagine that this error is fixable in a similar way as the issues in #600, but I don't know where to start looking. Could someone help me out with this? It feels like I've gotten to the last step of working through the cross-compilation issues, so I'd love to get it working completely.

EPashkin:
Seems GtkTextViewLayer only added in 3.14 as `GtkTextViewClass.draw_layer` is only function that use it.
So you only need do same as in #603 (also look at https://github.com/gtk-rs/gtk/issues/600#issuecomment-348775765)

foo-jin:
Although compilation succeeds now, I get the following error when running it on windows:
```
Error: "(packt-gui.exe:372) Gtk-ERROR **: failed to add UI:
<input>:21:1 <object> requires attribute "id""'
```

EPashkin:
Check your glade file

sdroege:
The issue with the `get_type()` function for enums should be solved since a while.

PR #654, opened 2018-05-03
Change version for Gtk.TextViewLayer 

foo-jin:
The fix as suggested in #653.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -201,7 +201,6 @@ generate = [
     "Gtk.TextMark",
     "Gtk.TextSearchFlags",
     "Gtk.TextTagTable",
-    "Gtk.TextViewLayer",
     "Gtk.TextWindowType",
     "Gtk.ToggleButton",
     "Gtk.ToggleToolButton",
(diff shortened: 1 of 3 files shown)


Gir.toml
```diff
@@ -1962,3 +1961,8 @@ status = "generate"
     #value glib.Variant
     name = "action-target"
     ignore = true
+
+[[object]]
+name = "Gtk.TextViewLayer"
+status = "generate"
+version = "3.14"
```
EPashkin: Please, move this between Gtk.TextView and Gtk.Toolbar.

GuillaumeGomez: We should add a test to enforce that it's alpha sorted in the toml file. I'm going to add when I have a bit of time. :)

EPashkin:
@GuillaumeGomez Yes no version in docs, but enum not present in 3.10 libs and used only in internal `GtkTextViewClass.draw_layer`, so it safe to say that it added with it.

GuillaumeGomez:
That's weird... But whatever, let's merge it then! Thanks!

EPashkin:
Hm, missed that it not against master.
I will repeat  on master

GuillaumeGomez:
Wait what?! Oh damn! I remove the commit from history.

GuillaumeGomez:
Ok done. Please reopen it against the master branch (or maybe not needed if @EPashkin does it?).

EPashkin:
@foo-jin I replaced PR. Thanks for reporting problem.

PR #655, opened 2018-05-03
Text view layer version

EPashkin:
Replacement of #654 
Part of #653


GuillaumeGomez:
Thanks for reopening it on the correct branch! I'll just go ahead and merge it. It passed crate CIs already after all...

PR #660, opened 2018-05-08
Regen

EPashkin:
cc @GuillaumeGomez, @sdroege


GuillaumeGomez:
Emoji are coming in as well. 😆

GuillaumeGomez:
Just the error:

```
error: casting from `*mut ffi::GtkAppChooser` to a more-strictly-aligned pointer (`*mut gobject_ffi::GObject`)
```

Otherwise all good. Thanks @EPashkin!

EPashkin:
While clippy formally right in this case IMHO it need be ignored in glib, as pointers come from C and actually have right alignment.
```
#[repr(C)]
pub struct GtkTreeSortable(c_void);

#[repr(C)]
#[derive(Copy, Clone)]
pub struct GObject {
    pub g_type_instance: GTypeInstance,
    pub ref_count: /*volatile*/c_uint,
    pub qdata: *mut glib::GData,
}

//! assert_eq!(std::mem::align_of::<ffi::GtkTreeSortable>(), 1);
//! assert_eq!(std::mem::align_of::<*mut ffi::GtkTreeSortable>(), 8);
//! assert_eq!(std::mem::align_of::<gobject_ffi::GObject>(), 8);
//! assert_eq!(std::mem::align_of::<*mut gobject_ffi::GObject>(), 8);
```

PR #662, opened 2018-05-20
Add ListBoxExt::set_sort_func()

Ereski:


EPashkin:
@Ereski, Thansk, but seems you make manual change in file that can't editing manually (see comment in header) and thus get https://travis-ci.org/gtk-rs/gtk/jobs/381303988#L876.
Right way is add `src/list_box.rs` with trait `ListBoxExtManual` with this function (Ex. https://github.com/gtk-rs/gtk/blob/master/src/list_store.rs#L27).
Also better make function ignored (https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L353-L356) and regenerate files with 'make gir'.

EPashkin:
Functions itself looks good for me.

Ereski:
Didn't realise the build system was that complicated. So, the right way would be to add a new trait `ListBoxExtManual` and implement `set_sort_func()` there, right? Will the new fn be merged into `ListBoxExt`?

EPashkin:
Function will be not me merged in `ListBoxExt`, but it can simply be called on any `ListBox` and its descendant.

sdroege:
@GuillaumeGomez Please close this one, we autogenerate the function now

PR #667, opened 2018-06-23
Remove callback guard

EPashkin:
Part of https://github.com/gtk-rs/glib/pull/250

cc @GuillaumeGomez, @sdroege


check_init_asserts
```diff
@@ -10,8 +10,7 @@ while (/^(\N*)\V*fn\s+(\w+)\s*(<[^(]+>)?\s*(\([^{;]+)\{\N*\n^(\N*)$/gms) {
 		$first_line =~ /^\s*(
 			assert_initialized_main_thread |
 			assert_not_initialized |
-			skip_assert_initialized |
-			callback_guard
+			skip_assert_initialized
```
EPashkin: unsafe functions just not checked, so this line not worked anyway

src/rt.rs
```diff
@@ -47,15 +47,6 @@ macro_rules! assert_not_initialized {
     )
 }
 
-macro_rules! callback_guard {
-    () => (
-        let _guard = ::glib::CallbackGuard::new();
-        if cfg!(debug_assertions) {
-            assert_initialized_main_thread!();
-        }
```
EPashkin: I slightly worry about this check, maybe it really needed

sdroege:
Yes but it should not be in the callback guard macro. I thought there was another macro for that?

EPashkin:
This macro named `assert_initialized_main_thread` ;)
I just not sure that we need check that all callback called from main thread in gtk.

EPashkin:
Ok, then I think that PR is ready

GuillaumeGomez:
I just hope gtk doesn't have another bug. :-/

Thanks @EPashkin and @sdroege!

PR #669, opened 2018-06-28
Use futures-preview crates instead of futures

sdroege:
The futures crates were janked from crates.io and renamed to
futures-preview for whatever reason.


GuillaumeGomez:
Thanks!

PR #672, opened 2018-07-02
Add ListBox::bind_model() and other functions taking closures

sdroege:
This is currently WIP, I'd like to add the other commented out functions too but it's becoming repetitous. Should probably do https://github.com/gtk-rs/gir/issues/618 first.

Any comments so far?


GuillaumeGomez:
Nope, seems good to me. Or at least didn't see any outstanding issue.

EPashkin:
Looks good, I little worried about the lack of boxing, but I trust you

sdroege:
> Looks good, I little worried about the lack of boxing, but I trust you

It's the same as boxing twice, just that we store it on the stack instead. The first reference taken creates a trait object (two pointers, the struct pointer and the vtable), the second reference is a single pointer to that trait object (and if it was more than a single pointer, the cast to `gpointer` would fail to compile).

sdroege:
Should we get this in for now until the gir changes are done (probably going to take a few weeks :) )?

GuillaumeGomez:
Sounds fine to me. Thanks! :)

PR #678, opened 2018-07-21
New types

GuillaumeGomez:
cc @EPashkin @sdroege


src/auto/accel_group.rs
```diff
@@ -35,9 +36,12 @@ impl AccelGroup {
         }
     }
 
-    //pub fn from_accel_closure(closure: /*Ignored*/&glib::Closure) -> Option<AccelGroup> {
-    //    unsafe { TODO: call ffi::gtk_accel_group_from_accel_closure() }
-    //}
+    pub fn from_accel_closure(closure: &glib::Closure) -> Option<AccelGroup> {
```
sdroege: Instead of having a plain `GClosure` here it would be nicer to implement this manually around the actual type that is expected from the closure. `Fn(&AccelGroup, &Object, u32, gdk::ModifierType) -> bool` (`GtkAccelGroupActivate`) from what I can see

src/auto/accel_group.rs
```diff
@@ -49,11 +53,11 @@ impl Default for AccelGroup {
 pub trait AccelGroupExt {
     //fn activate<P: IsA<glib::Object>>(&self, accel_quark: /*Ignored*/glib::Quark, acceleratable: &P, accel_key: u32, accel_mods: gdk::ModifierType) -> bool;
 
-    //fn connect(&self, accel_key: u32, accel_mods: gdk::ModifierType, accel_flags: AccelFlags, closure: /*Ignored*/&glib::Closure);
+    fn connect(&self, accel_key: u32, accel_mods: gdk::ModifierType, accel_flags: AccelFlags, closure: &glib::Closure);
```
sdroege: Same here

src/auto/accel_group.rs
```diff
@@ -49,11 +53,11 @@ impl Default for AccelGroup {
 pub trait AccelGroupExt {
     //fn activate<P: IsA<glib::Object>>(&self, accel_quark: /*Ignored*/glib::Quark, acceleratable: &P, accel_key: u32, accel_mods: gdk::ModifierType) -> bool;
```
sdroege: We should probably really add `GQuark` bindings at some point. Easy enough and appearing in so many places of APIs.

src/auto/accel_group.rs
sdroege: And basically everywhere where a `GClosure` appears but the API actually expects a specific signature

src/auto/im_context_simple.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+impl Default for IMContextSimple {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+pub trait IMContextSimpleExt {
+    fn add_compose_file(&self, compose_file: &str);
```
sdroege: This should probably take a `&Path`?

src/auto/im_context_simple.rs
```diff
@@ -0,0 +1,57 @@
...
+    fn add_compose_file(&self, compose_file: &str) {
+        unsafe {
+            ffi::gtk_im_context_simple_add_compose_file(self.to_glib_none().0, compose_file.to_glib_none().0);
+        }
+    }
+
+    fn add_table(&self, data: &[u16], n_seqs: i32) {
+        let max_seq_len = data.len() as i32;
+        unsafe {
+            ffi::gtk_im_context_simple_add_table(self.to_glib_none().0, data.to_glib_none().0, max_seq_len, n_seqs);
```
sdroege: This is wrong: `Adds an additional table to search to the input context. Each row of the table consists of max_seq_len key symbols followed by two guint16 interpreted as the high and low words of a gunicode value. Tables are searched starting from the last added.` and ` n_seqs | number of sequences in the table`

We need to do something more clever here for the length of the array

src/pad_controller.rs
```diff
@@ -0,0 +1,28 @@
...
+
+pub trait PadControllerExtManual {
+    #[cfg(any(feature = "v3_22", feature = "dox"))]
+    fn set_action_entries(&self, entries: &[&PadActionEntry]);
+}
+
+impl<O: IsA<PadController>> PadControllerExtManual for O {
+    #[cfg(any(feature = "v3_22", feature = "dox"))]
+    fn set_action_entries(&self, entries: &[&PadActionEntry]) {
+        let mut entries = entries.iter().map(|p| p.to_glib()).collect::<Vec<_>>();
```
sdroege: If you add a `[repr(C)]` to `PadActionEntry`, you can simply to `entries.as_ptr()` here instead of the copy

src/page_range.rs
```diff
@@ -0,0 +1,42 @@
+use ffi;
+use glib::translate::*;
+
+pub struct PageRange(ffi::GtkPageRange);
+
+impl PageRange {
```
sdroege: This is not used anywhere right now. But would be used in `gtk_print_settings_get_page_ranges()` and the setter.

Please add those at the same time (and same thing with `[repr(C)]` and directly converting slices via casts here)

src/auto/accel_group.rs
GuillaumeGomez: My next glib PR I suppose. ;)

src/auto/im_context_simple.rs
GuillaumeGomez: It should indeed!

src/pad_controller.rs
GuillaumeGomez: Seems weird to convert a `&[&T]` into a `*mut T`. There is an extra pointer level no?

GuillaumeGomez: Or should I just transform `&[&T]` parameter to `&[T]`? Sounds simpler.

src/auto/accel_group.rs
GuillaumeGomez: Since this one is supposed to return an `AccelGroup` for the matching `Closure`, I wonder it would work then...

src/im_context_simple.rs
```diff
@@ -0,0 +1,35 @@
...
+
+impl<O: IsA<IMContextSimple>> IMContextSimpleExtManual for O {
+    fn add_compose_file<P: AsRef<Path>>(&self, compose_file: P) {
+        unsafe {
+            let compose_file = compose_file.as_ref();
+            ffi::gtk_im_context_simple_add_compose_file(self.to_glib_none().0, compose_file.to_glib_none().0);
+        }
+    }
+
+    fn add_table(&self, data: &[u16], max_seq_len: u32, n_seqs: u32) {
```
GuillaumeGomez: @sdroege: So I did it like this. Should we instead compute `n_seqs` from the `data` length or just keep it like this?

sdroege: I'll check in detail on Monday, also your question about the closure/accelgroup

sdroege: I don't actually understand how that API works. This adds a whole table, so the data is a table. The docs say that `Each row of the table consists of max_seq_len key symbols followed by two guint16 interpreted as the high and low words of a gunicode value.`. Now how do we know how long each row in the data is?

Once we know how that API works, it might be better to instead pass a (repr packed!)  slice of `struct TableRow { symbols: XX, value: [u16; 2] }`.

src/auto/accel_group.rs
sdroege: It wouldn't indeed, unless we let `gtk_accel_group_connect()` return the closure that we internally created. This API is really nice :man_facepalming:

Issue #690, opened 2018-08-25
Calling class functions


Exalm:
It seems it's impossible to call class functions like `gtk_widget_class_set_name()` or `gtk_container_class_install_child_properties()` in gtk-rs currently.

GuillaumeGomez:
I didn't find those functions anywhere. Do you have a link to docs for those?

Exalm:
argh, I've managed to make a typo in the first one :(
https://developer.gnome.org/gtk3/stable/GtkWidget.html#gtk-widget-class-set-css-name
https://developer.gnome.org/gtk3/stable/GtkContainer.html#gtk-container-class-install-child-properties

GuillaumeGomez:
Ok, now I found it. The problem is that `WidgetClass` isn't (and can't be) generated. It needs to be bound manually. If you're motivated to do it, we can provide you help.

Exalm:
It's not urgent in any way :D

Exalm:
The thing I had in mind was making a subclass of a widget and using set_css_name() to change its css node name during initialization, yes.

Exalm:
Well, as I said, it's not urgent in any way and was discovered pretty much accidentally. :)

sdroege:
I also wrote a bit about subclassing here if you want to do the work manually: https://coaxion.net/blog/2017/09/exporting-a-gobject-c-api-from-rust-code-and-using-it-from-c-python-javascript-and-others/

And if you want to help with gobject-subclass, let me know :)

PR #691, opened 2018-08-25
Widget class

GuillaumeGomez:
Part of #690.

cc @sdroege @EPashkin


EPashkin:
@GuillaumeGomez, Thanks, LGFM
But seems we need add `#![cfg_attr(feature = "cargo-clippy", allow(derive_hash_xor_eq))]` to `lib.rs` too.
Currently 3 clippy error like this:
```
error: you are deriving `Hash` but have implemented `PartialEq` explicitly
  --> src/auto/paper_size.rs:16:38
   |
16 |     #[derive(Debug, PartialOrd, Ord, Hash)]
   |                                      ^^^^
   |
   = note: #[deny(derive_hash_xor_eq)] on by default
note: `PartialEq` implemented here
  --> src/auto/paper_size.rs:182:1
   |
182| / impl PartialEq for PaperSize {
183| |     #[inline]
184| |     fn eq(&self, other: &Self) -> bool {
185| |         self.is_equal(other)
186| |     }
187| | }
   | |_^
   = help: for further information visit https://rust-lang-nursery.github.io/rust-clippy/v0.0.212/index.html#derive_hash_xor_eq
```

GuillaumeGomez:
@EPashkin Weird... I'll make another PR to do it.

@sdroege: Ok! Can't wait to see what you'll come with. :)

sdroege:
This is not really fixed btw :) You would now add the functions to `gtk::WidgetClass`, which exists now (and there's a way to get it from an instance). See for example here https://gitlab.freedesktop.org/gstreamer/gstreamer-rs/blob/7a69a1137c190927e351205074b641d525e6ce25/gstreamer/src/subclass/element.rs#L245-274

PR #692, opened 2018-08-26
remove more clippy warnings

GuillaumeGomez:
cc @EPashkin


EPashkin:
To many changes on check regen and undoc check :(
I don't understand, why.
 IMHO we can just merge as is and then rebase #691 to master

GuillaumeGomez:
Let's do it like this then!

PR #708, opened 2018-09-28
Remove property `events` getter and setter as we already have normal …

EPashkin:
…variant

Fix https://github.com/gtk-rs/gir/issues/639

cc @GuillaumeGomez, @sdroege


sdroege:
:+1:

GuillaumeGomez:
Why did it remove two methods? Is it because before it was all automatically generated and now only one is?

sdroege:
The whole point of the commit is to remove these two functions... because we have getters/setters based on functions already, no need to go through the property.

GuillaumeGomez:
Ok. Makes sense. Let's merge it then!

Thanks!

PR #720, opened 2018-10-06
Allow the create-context signal for GLArea to return null

### LINKED ISSUE #703: Change GLArea create_context handler to return Option<GLContext> instead of GLContext
Right now, the signal handler for `create_context` on the `GLArea` widget must return a `GLContext` object. However, creating a context is not guaranteed to work, as `window.create_gl_context` returns a `Result`. This essentially requires the code implementing the signal to panic if context creation fails, instead of possibly implementing a more graceful error mechanism.

The documentation and the source code of GLArea suggest that the `create_context` signal is allowed to return a null pointer, so the `connect_create_context` function should take a function that returns `Option<GLContext>` instead of `GLContext`.

ColonelThirtyTwo:
The signal should return None if creating the context fails. Optionally set
an error message with GLAreaExt::set_error.

Fixes #703

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -813,6 +813,10 @@ module_name = "gl_area"
     [[object.signal]]
     pattern = "render"
     inhibit = true
+    [[object.signal]]
+    name = "create-context"
+        [object.signal.return]
+        nullable = true
 
 [[object]]
 name = "Gtk.Grid"
--- a/src/auto/gl_area.rs
+++ b/src/auto/gl_area.rs
@@ -113,7 +113,7 @@ pub trait GLAreaExt {
     fn set_use_es(&self, use_es: bool);
 
     #[cfg(any(feature = "v3_16", feature = "dox"))]
-    fn connect_create_context<F: Fn(&Self) -> gdk::GLContext + 'static>(&self, f: F) -> SignalHandlerId;
+    fn connect_create_context<F: Fn(&Self) -> Option<gdk::GLContext> + 'static>(&self, f: F) -> SignalHandlerId;
 
     #[cfg(any(feature = "v3_16", feature = "dox"))]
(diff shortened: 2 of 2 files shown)


ColonelThirtyTwo:
Example program:

```rust
extern crate gtk;
extern crate gio;
extern crate gdk;

use std::env::args;
use gio::ApplicationExt;
use gio::ApplicationExtManual;
use gtk::prelude::*;

fn build_ui(application: &gtk::Application) {
	let window = gtk::ApplicationWindow::new(application);
	window.set_title("Test Window");
	window.set_border_width(10);
	window.set_position(gtk::WindowPosition::Center);
	window.set_default_size(350, 70);
	window.connect_delete_event(move |win, _| {
		win.destroy();
		Inhibit(false)
	});
	
	let glarea = gtk::GLArea::new();
	glarea.connect_create_context(|this| {
		this.set_error(Some(&gtk::Error::new(gdk::GLError::NotAvailable, "error happened")));
		None
	});
	window.add(&glarea);
	window.show_all();
}

fn main() {
	let application = gtk::Application::new("com.github.basic",
		gio::ApplicationFlags::empty())
		.expect("Initialization failed...");
	
	application.connect_startup(|app| {
		build_ui(app);
	});
	application.connect_activate(|_| {});
	application.run(&args().collect::<Vec<_>>());
}
```

EPashkin:
Don't forget do `make` and add generated changes :wink:

EPashkin:
@GuillaumeGomez 2 travis jobs failed due network, please. restart them.
👍 for merge when CI passed from me.

GuillaumeGomez:
Looks good to me as well, thanks! (Waiting for restarted CIs.)

Also, can you add your example (with whitespaces instead of tabs) into the `gtk-rs/examples` repository please?

EPashkin:
@GuillaumeGomez IMHO this example is useless, but if some drawing added instead, then that it be much better.

GuillaumeGomez:
Well, we can always draw a triangle. :)

PR #726, opened 2018-10-11
Update ResponseType

### LINKED ISSUE #725: Add other variant into ResponseType
Shouldn't we add an `Other(i32)` variant into the [`ResponseType`](http://gtk-rs.org/docs/gtk/enum.ResponseType.html) enum? Docs says that gtk response types are below 0 but that users can provide custom ones. Shouldn't it be wrapped as well?

GuillaumeGomez:
Fixes #725.

cc @sdroege @EPashkin


src/response_type.rs
```diff
@@ -0,0 +1,92 @@
...
+            -3 => ResponseType::Accept,
+            -4 => ResponseType::DeleteEvent,
+            -5 => ResponseType::Ok,
+            -6 => ResponseType::Cancel,
+            -7 => ResponseType::Close,
+            -8 => ResponseType::Yes,
+            -9 => ResponseType::No,
+            -10 => ResponseType::Apply,
+            -11 => ResponseType::Help,
+            value if value >= 0 => ResponseType::Other(value as u16),
```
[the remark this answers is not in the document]
sdroege: Yes

sdroege:
Looks good except for the one comment @EPashkin wrote

GuillaumeGomez:
Updated.

src/response_type.rs
```diff
@@ -0,0 +1,92 @@
...
+    Ok,
+    Cancel,
+    Close,
+    Yes,
+    No,
+    Apply,
+    Help,
+    Other(u16),
+    #[doc(hidden)]
+    Unknown(i32),
```
sdroege: Why did it lose its two underscores in the beginning?

GuillaumeGomez: Want me to put them back? At first I just replaced it with `Other` and then put it back.

sdroege: I'd prefer that with underscores, yes

GuillaumeGomez:
Updated `Unknown`.

sdroege:
:+1:

Issue #730, opened 2018-10-24
Update doc examples


GuillaumeGomez:
The documentation examples should be updated (using old style gtk).

sdroege:
What does that mean exactly?

sdroege:
Ah yes, using `gtk::Application` for everything. I thought we had an issue for that already.

GuillaumeGomez:
Fixed by #731.

PR #731, opened 2018-10-24
Update doc example

GuillaumeGomez:
Part of #730.

cc @sdroege @EPashkin


src/lib.rs
```diff
@@ -53,26 +53,29 @@
...
 //! fn main() {
-//!     gtk::init().unwrap();
-//!     // Create the main window.
-//!     let window = Window::new(WindowType::Toplevel);
-//!     // UI initialization.
-//!     // ...
-//!     // Don't forget to make all widgets visible.
-//!     window.show_all();
-//!     // Handle closing of the window.
-//!     window.connect_delete_event(|_, _| {
```
EPashkin: All your examples contains
```
    window.connect_delete_event(|win, _| {
        win.destroy();
        Inhibit(false)
    });
```

EPashkin: so IMHO it still needed here too

GuillaumeGomez: Yes sir!

GuillaumeGomez:
Updated.

EPashkin:
@GuillaumeGomez Thanks

src/lib.rs
```diff
@@ -53,26 +53,37 @@
...
-//!         // Stop the main loop.
-//!         gtk::main_quit();
-//!         // Let the default handler destroy the window.
-//!         Inhibit(false)
+//!     let uiapp = gtk::Application::new("org.gtkrsnotes.demo",
+//!                                       gio::ApplicationFlags::FLAGS_NONE)
+//!                                  .expect("Application::new failed");
+//!     uiapp.connect_activate(|app| {
```
EPashkin: Strange, UI builds in `connect_startup` and `connect_activate` empty in examples,
but gtk examples https://gitlab.gnome.org/GNOME/gtk/blob/master/examples/bp/bloatpad.c shows that UI needed be created in `connect_activate`,
seems we need change examples.

src/lib.rs
EPashkin: @GuillaumeGomez, @sdroege Seems main windows destroyed normally even without `connect_delete_event`,
any reason to use it?

src/lib.rs
GuillaumeGomez: Ah indeed!

GuillaumeGomez:
Removed the unneeded `connect_delete`.

EPashkin:
For me this PR ready for merge.
@GuillaumeGomez, thanks

sdroege:
Yes, go ahead :)

PR #741, opened 2018-11-15
Add check for 1.28

GuillaumeGomez:


EPashkin:
@GuillaumeGomez Seems I missed something. Why you to do this?

GuillaumeGomez:
If these files are useless, shouldn't we remove them?

EPashkin:
This question about replace stable with 1.28  (while PR title says "add")

GuillaumeGomez:
It's a discussion we had with @sdroege: gstreamer is supporting `1.28` (giving you context about all those PRs). But yes, it should be called "replace" instead. Replacing all titles (:sob:).

PR #748, opened 2018-12-16
use &self for SelectionData::set(), fix #747

### LINKED ISSUE #747: SelectionData::set() uses &mut self where all other functions use &
SelectionData::set() is the only function which takes self as &mut. It should be consistent with other functions.
https://gtk-rs.org/docs/gtk/struct.SelectionData.html#method.set
@sdroege

jsparber:


jsparber:
I have no idea why the CI fails, the log has only warnings

EPashkin:
This was error that was fixed today, restarted some jobs
```
error[E0605]: non-primitive cast: `i32` as `socket::RawSocket
38;5;    ffi::g_socket_get_fd(self.to_glib_none().0) as _
```

Gir.toml
```diff
@@ -1390,6 +1390,11 @@ status = "generate"
         name = "selection_data"
         const = true
     [[object.function]]
+    name = "set"
+        [[object.function.parameter]]
+        name = "selection_data"
+        const = true
```
sdroege: While you're here, you could also ignore the getter and implement it with its correct name :)

EPashkin:
@osa1 Thanks for reminder.
We waited for https://github.com/gtk-rs/gtk/pull/748#discussion_r242232424,
but seems @jsparber don't have time and we forgot about this issue.
Changing connect_drag_data_get maybe tough IMHO as it inner function type.
Also there was inconsistency with all other setters which  already changed to `&self`.
Agree change boxed value without mutable reference is strange from rust point of view,
but can seen same as using `RefCell`

I restarted travis CI just in case, if it still pass 👍 from me.

Gir.toml
sdroege: I.e. `fn get_data()` should call `gtk_selection_data_get_data_with_length()`

sdroege:
I'd prefer if someone also fixed the getter while we're at it. It's a trivial change and otherwise it will never happen. @osa1 do you want to submit another PR for that?

> Also there was inconsistency with all other setters which already changed to `&self`.
> Agree change boxed value without mutable reference is strange from rust point of view,
> but can seen same as using `RefCell`

There's nothing strange about that really. It works like this for all shared boxed types and GObjects in the bindings, it's simply interior mutability.

osa1:
> I'd prefer if someone also fixed the getter while we're at it. It's a trivial change and otherwise it will never happen. @osa1 do you want to submit another PR for that?

I don't understand what's wrong currently so not sure how can I help with that.

osa1:
But in the meantime it'd be really helpful if we could merge this (for some reason I can't use my fork of `gtk-rs` in my crate, cargo doesn't like it even though it's just `gtk-rs` master + one commit).

sdroege:
> I don't understand what's wrong currently so not sure how can I help with that.

```rust
    pub fn get_data_with_length(&self) -> Vec<u8>
```

should become

```rust
    pub fn get_data(&self) -> Vec<u8>
```

> for some reason I can't use my fork of `gtk-rs` in my crate, cargo doesn't like it even though it's just `gtk-rs` master + one commit

You need to replace it for all of your dependencies that depend themselves on gtk too. The easiest is via the [patch feature](https://doc.rust-lang.org/cargo/reference/manifest.html#the-patch-section) of cargo.

osa1:
How do I do that? I don't see `get_data_with_length()` defined in the `Gir.toml` file, and the `Gtk-3.0.gir` says that I the file is generated from C and I should be editing that instead, even though there isn't a line of C in this repo.

EPashkin:
@sdroege I seen `new_name=` for parameters, but it seems not works with function (not tested).
IMHO we better try using rename annotation
In both cases may problem with docs.

jsparber:
I actually forgot about this MR myself. And i didn't have time to implemented the getter. @sdroege Did i understand it correctly that we should just extant gir witht the rename functionality?

EPashkin:
@jsparber It was done in #762

sdroege:
We still need to add support to gir for doing this automatically though :)

jsparber:
awesome!

PR #750, opened 2018-12-16
Generate ComboBox::set_active by hand

### LINKED ISSUE #749: Combobox::set_active should take an option
The docs say that [set_active](https://gtk-rs.org/docs/gtk/trait.ComboBoxExt.html#tymethod.set_active) will deactivate the current one if the given value is `-1`. `Option` seems better in rust case.

GuillaumeGomez:
Fixes #749 

cc @EPashkin @sdroege

### DIFF
--- a/src/auto/combo_box.rs
+++ b/src/auto/combo_box.rs
@@ -136,8 +136,6 @@ pub trait ComboBoxExt {
 
     fn popup_for_device<P: IsA<gdk::Device>>(&self, device: &P);
 
-    fn set_active(&self, index_: i32);
-
     fn set_active_id<'a, P: Into<Option<&'a str>>>(&self, active_id: P) -> bool;
 
     fn set_active_iter<'a, P: Into<Option<&'a TreeIter>>>(&self, iter: P);
@@ -169,6 +167,8 @@ pub trait ComboBoxExt {
     fn set_title(&self, title: &str);
 
     fn set_wrap_width(&self, width: i32);
+
+    fn set_property_active(&self, active: i32);
 
     fn get_property_cell_area(&self) -> Option<CellArea>;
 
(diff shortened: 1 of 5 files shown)


src/combo_box.rs
```diff
@@ -0,0 +1,24 @@
+// Copyright 2018, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use ComboBox;
+use ffi;
+use glib::object::IsA;
+
+pub trait ComboBoxExtManual {
+    fn set_active<'a, P: Into<Option<u32>>>(&self, index_: P);
```
EPashkin: `'a` is really needed ?

EPashkin:
`ComboBoxExtManual` need be in prelude, and old function ignored

src/combo_box.rs
sdroege: No, and I'm surprised it's not a compiler warning :)

sdroege:
looks good to me apart from what @EPashkin said

GuillaumeGomez:
I completely forgot to commit all files, sorry...

EPashkin:
Why don't ignore property getter and setter?
```
    [[object.property]]
    name = "active"
    #manual Option
    generate = ["notify"]
```

src/auto/combo_box.rs
```diff
@@ -447,6 +435,20 @@ impl<O: IsA<ComboBox> + IsA<glib::object::Object> + glib::object::ObjectExt> Com
         }
     }
 
+    fn get_property_active(&self) -> i32 {
```
sdroege: The property functions should disappear (but notify should stay), otherwise looks good to me

EPashkin:
@GuillaumeGomez thanks, 👍

PR #751, opened 2019-01-04
Update BoolError constructions

fengalin:
See https://github.com/gtk-rs/glib/pull/419


GuillaumeGomez:
Seems good to me. Thanks!

EPashkin:
@fengalin Thanks again.
If no changes in generated code, then update sub-modules not needed and only cause nightly CI failure on "make regen_check"

fengalin:
@EPashkin, yes of course! Sorry about that.

PR #752, opened 2019-01-06
Ignore some more problematic functions

GuillaumeGomez:
They all need to be generated by hands.

cc @EPashkin @sdroege

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -678,6 +678,22 @@ status = "generate"
     name = "set_can_store"
     #array with size
     ignore = true
+    [[object.function]]
+    name = "request_targets"
+    #array with size
+    ignore = true
+    [[object.function]]
+    name = "request_uris"
+    #array with size
+    ignore = true
+    [[object.function]]
+    name = "set_with_data"
+    #array with size
+    ignore = true
+    [[object.function]]
+    name = "set_with_owner"
+    #array with size
+    ignore = true
 
 [[object]]
 name = "Gtk.ColorChooser"
@@ -1505,6 +1521,9 @@ status = "generate"
         const = true
     [[object.signal]]
     name = "insert-text"
+    ignore = true
+    [[object.function]]
+    name = "register_serialize_format"
     ignore = true
 
 [[object]]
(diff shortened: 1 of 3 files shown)


EPashkin:
@GuillaumeGomez Even after your gtk-rs/gir#667 ?
You will add regen or/and manual implementations in other PR?

GuillaumeGomez:
The problem for those ones is that it has pointers of pointers. The problem is even before my PR. :)

EPashkin:
Now reason is more clear, its preparing for time when closure implemented (as currently functions commented anyway).
Still comments `#array with size` is misleading for `request_targets` as no arrays in function definition, (only in its subtype),
also in `request_uris` type `ClipboardURIReceivedFunc` don't contains any size,

Still not clear for me if PR is ready for merge or not as it not pass "regen_check"

GuillaumeGomez:
That's interesting: setting functions as ignored in this case make them generated (as commented but still). Surprising, no?

EPashkin:
No, regen_check shows inverse diff, so these commented lines just removed

GuillaumeGomez:
Is it expected?

EPashkin:
Currently it shows changes that "wrongly added" to generated version.
Not fully sure is right or we better show just changes after "right done regen"

GuillaumeGomez:
In any case it's a bit surprising since I just ignored some functions that aren't (yet) generated.

EPashkin:
👍 , Travis again can't install clippy 😢

GuillaumeGomez:
That's... interesting? XD Thanks for the review, I'll just merge then.

Gir.toml
```diff
@@ -678,6 +678,22 @@ status = "generate"
     name = "set_can_store"
     #array with size
     ignore = true
+    [[object.function]]
+    name = "request_targets"
+    #array with size
```
sdroege: What's the problem with this? We support that elsewhere just fine and we have conversion traits for this specific case :) `from_glib_none_num()`

Same for everything else below

GuillaumeGomez: Then that means that my gir PR is incomplete. :'(

sdroege: Indeed :(

Issue #702, opened 2018-09-12
gtk-rs on Windows with MSVC should be possible with zero attrition 


fungos:
note: I'm not sure if the best place is here or at sys repository, but as any user looking for gtk-rs on windows will tend to land here, I'm opening this issue here.

This is more of a distribution problem, but may be extremely beneficial for Rust GUI situation.

As it stands, gtk-rs is buildable on Windows using MSVC toolchain, but this does not mean it is usable without additional steps, as an application trying to link with GTK will still need all the GTK libraries previously downloaded (or worse, separately built) and well configured.

The ideal would have to be any application using gtk-rs and being compiled on windows using the MSVC toolchain, to have these libraries (downloaded or compiled) and configuration done by cargo itself (via plugin or build.rs in *-sys crates).

One way to achieve this, would be by abusing or mimicking the "cc" and "cmake" crates. This could be best achieved by using [vcpkg-rs](https://github.com/mcgoo/vcpkg-rs), although vcpkg itself has some issues[1] with GTK port, it builds flawlessly, so this can be used as a mean of distribution for dependencies when using MSVC toolchain.

Further [improvement](https://github.com/mcgoo/vcpkg-rs/issues/9) can be collaborated with vcpkg-rs

[1] The first issue is that vcpkg GTK port outputs "gtk-3.0.lib" and "gdk-3.0.lib" while gtk-rs expects "gtk-3.lib" and "gtk-3.lib". The second issue is that gtk-rs expects some debug features to be enabled on the libraries, for this I've opened [this issue/PR](https://github.com/Microsoft/vcpkg/issues/4212) there.

GuillaumeGomez:
> note: I'm not sure if the best place is here or at sys repository, but as any user looking for gtk-rs on windows will tend to land here, I'm opening this issue here.

Here is fine.

> ...

I agree that windows support isn't great. Any improvement on this side would be very appreciated!

sdroege:
I wouldn't suggest using the vcpkg packages as they have a custom build system different from upstream and last I checked it compiled but didn't work. Writing a custom CMake build system around these projects is not trivial and requires considerable maintenance effort.

GTK and all (?) its dependencies now have support for [meson](https://mesonbuild.com), which can either build via ninja with the MSVC toolchain or output VS solution files that can be built with msbuild.exe.

As that is supported and maintained by the GTK project, I would suggest going down that road instead.

----

[...]

PR #753, opened 2019-01-10
Implement MSVC compatibility.

daxpedda:
See #702 for more discussions about this.
I got the inspiration from here: https://github.com/kykc/gtk-rs-msvc.

There is very little necessary to make it compile and run:
1. install [vcpkg](https://github.com/Microsoft/vcpkg)
2. `vcpkg install gtk:x64-windows` (or `vcpkg install gtk:x86-windows` respectively)
3. define `VCPKG_ROOT` for [`vcpkg::find_package()`](https://docs.rs/vcpkg/0.2.6/vcpkg/) to work
4. run `vcpkg::find_package("gtk")` to feed the linker with libs
5. copy `gtk-3.0.lib` to `gtk-3.lib` and `gdk-3.0.lib` to `gdk-3.lib`

This PR currently implements step 4. and 5. into `build.rs`.

**Question**
- Should we automatize step 1. and 2. too? I have already tried it and it works splendidly, but may be out of scope for this.

**Pending Tasks if it's decided to implement this**
- [ ] Update https://gtk-rs.org/docs/requirements.html.
- [ ] Document the quirks of the vcpkg version of GTK. One difference that I noticed is that it uses a non-default theme by default.
- [ ] Document some help for dll distribution.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -42,6 +42,9 @@ v3_14 = ["v3_12", "gtk-sys/v3_14", "gdk/v3_14"]
 dox = ["gdk/dox", "gtk-sys/dox"]
 futures = ["futures-core-preview", "fragile", "gio/futures"]
 
+[target.'cfg(all(target_os = "windows", target_env = "msvc"))'.build-dependencies]
+vcpkg = "0.2.6"
+
 [target.'cfg(target_os = "macos")'.build-dependencies]
 cc = "^1.0"
 
--- a/build.rs
+++ b/build.rs
@@ -1,10 +1,20 @@
 #[cfg(target_os = "macos")]
 extern crate cc;
 
+#[cfg(all(target_os = "windows", target_env = "msvc"))]
+extern crate vcpkg;
+
+use std::{
+    env,
+    fs::{copy, metadata},
+};
+
 fn main() {
     manage_docs();
     #[cfg(target_os = "macos")]
     build_foreground();
+    #[cfg(all(target_os = "windows", target_env = "msvc"))]
+    build_with_vcpkg();
 }
 
 #[cfg(any(feature = "embed-lgpl-docs", feature = "purge-lgpl-docs"))]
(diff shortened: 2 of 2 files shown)


EPashkin:
@daxpedda Not simpler just change names in sys build script https://github.com/gtk-rs/sys/blob/master/gtk-sys/build.rs#L22 ?
Problem that its generated, so solution need be more general.

sdroege:
See also https://github.com/gtk-rs/gtk/issues/702

This should indeed be done in the -sys crates, and more importantly it needs be optional. It would break at least my current workflow on Windows/MSVC so would likely also break others workflows. I'm using `pkg-config` and GTK/etc binaries not from vcpkg.

sdroege:
One could first check with `pkg-config`, if that fails `vcpkg` can be checked. And the automatic download/etc should probably only happen on Windows if at all :)

EPashkin:
Problem that we usually check GTK_LIB_DIR first and then just leave, at minimum in case that
https://github.com/kykc/gtk-rs-msvc suggest.
Also I don't like add new dependence that will be used only in two target to build script,
so prefer just use environment.

sdroege:
> Problem that we usually check GTK_LIB_DIR first

That hack should really disappear anyway. `pkg-config` is supposed to be used for GLib-based libraries, and that's what also works fine in general except for problems on the CI here :)

> Also I don't like add new dependence

What new dependency?

> that will be used only in two target to build script, so prefer just use environment.

What do you mean with "only in two target"? Windows and macOS? It can be added to the target specific build dependencies, there's not really a problem then :)

EPashkin:
> 
> 
> > Problem that we usually check GTK_LIB_DIR first
> 
> That hack should really disappear anyway. `pkg-config` is supposed to be used for GLib-based libraries, and that's what also works fine in general except for problems on the CI here :)

So your prefer add vcpkg fallback in case no pkg-config or its failed https://github.com/gtk-rs/sys/blob/master/gtk-sys/build.rs#L89?

> > Also I don't like add new dependence
> 
> What new dependency?

https://crates.io/crates/vcpkg

> > that will be used only in two target to build script, so prefer just use environment.
> 
> What do you mean with "only in two target"? Windows and macOS? It can be added to the target specific build dependencies, there's not really a problem then :)

I thought about x64_86\i686 on windows\msvc only, but by description in can be used in Linux and macOS too

sdroege:
> So your prefer add vcpkg fallback in case no pkg-config or its failed https://github.com/gtk-rs/sys/blob/master/gtk-sys/build.rs#L89?

I would completely remove the `GTK_LIB_DIR` hack. First check `pkg-config`, fall back to `vcpkg` and otherwise error out and ask people to set up their environment correctly :)



> > What new dependency?
> 
> https://crates.io/crates/vcpkg

This doesn't really have any big dependencies itself (we already depend on those anyway or not?) so the overhead of having this as a build dependency does not seem very big. I wouldn't worry about that.

EPashkin:
I added temporary fix in sys, that used vcpkg

EPashkin:
Sorry, we decide that `vcpkg` incomplete "build system" and we don't need support it.
So no vcpkg direct support, use `GTK_LIB_DIR` instead

PR #760, opened 2019-01-22
Regen

GuillaumeGomez:
I had to remove functions from manual code since they're now generated. We can say that's a great achievement! :D

cc @sdroege @EPashkin


src/auto/cell_layout.rs
```diff
@@ -94,9 +97,31 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
     //    unsafe { TODO: call ffi::gtk_cell_layout_set_attributes() }
     //}
 
-    //fn set_cell_data_func<'a, P: IsA<CellRenderer>, Q: Into<Option<&'a /*Unimplemented*/CellLayoutDataFunc>>, R: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(&self, cell: &P, func: Q, func_data: R, destroy: /*Unknown conversion*//*Unimplemented*/DestroyNotify) {
-    //    unsafe { TODO: call ffi::gtk_cell_layout_set_cell_data_func() }
-    //}
+    fn set_cell_data_func<P: IsA<CellRenderer>, Q: Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + Send + Sync + 'static, R: Into<Option<Q>>>(&self, cell: &P, func: R) {
```
EPashkin: Send+Sync?

src/auto/clipboard.rs
```diff
@@ -49,21 +50,70 @@ impl Clipboard {
...
+        let super_callback0: Box_<Option<P>> = callback_data;
+        unsafe {
+            ffi::gtk_clipboard_request_image(self.to_glib_none().0, callback, Box::into_raw(super_callback0) as *mut _);
+        }
+    }
 
-    //pub fn request_rich_text<P: IsA<TextBuffer>>(&self, buffer: &P, callback: /*Unknown conversion*//*Unimplemented*/ClipboardRichTextReceivedFunc) {
-    //    unsafe { TODO: call ffi::gtk_clipboard_request_rich_text() }
-    //}
+    pub fn request_rich_text<P: IsA<TextBuffer>, Q: FnOnce(&Clipboard, &gdk::Atom, GString, usize) + 'static>(&self, buffer: &P, callback: Q) {
```
EPashkin: GString again, will need regen.

src/auto/list_box.rs
```diff
@@ -302,10 +301,21 @@ impl<O: IsA<ListBox>> ListBoxExt for O {
         }
     }
 
-    //#[cfg(any(feature = "v3_14", feature = "dox"))]
-    //fn selected_foreach<P: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(&self, func: /*Unknown conversion*//*Unimplemented*/ListBoxForeachFunc, data: P) {
-    //    unsafe { TODO: call ffi::gtk_list_box_selected_foreach() }
-    //}
+    #[cfg(any(feature = "v3_14", feature = "dox"))]
+    fn selected_foreach<P: FnMut(&ListBox, &ListBoxRow)>(&self, func: P) {
```
EPashkin: We lost `IsA<ListBox>` here.

src/auto/widget.rs
```diff
@@ -901,10 +901,30 @@ impl<O: IsA<Widget>> WidgetExt for O {
         }
     }
 
-    //#[cfg(any(feature = "v3_8", feature = "dox"))]
-    //fn add_tick_callback(&self, callback: /*Unknown conversion*//*Unimplemented*/TickCallback, notify: /*Unknown conversion*//*Unimplemented*/DestroyNotify) -> u32 {
-    //    unsafe { TODO: call ffi::gtk_widget_add_tick_callback() }
-    //}
+    #[cfg(any(feature = "v3_8", feature = "dox"))]
+    fn add_tick_callback<P: Fn(&Widget, &gdk::FrameClock) -> bool + 'static>(&self, callback: P) -> u32 {
```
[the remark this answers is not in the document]
EPashkin: Added issue #761 For not forget do this someday (not in this PR).

src/auto/cell_layout.rs
sdroege: Yes that's wrong. Where does that come from? :)

src/auto/widget.rs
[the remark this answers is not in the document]
sdroege: Yes, ignore it for now please. Or fix it by manually implementing

src/auto/calendar.rs
```diff
@@ -200,9 +201,28 @@ impl<O: IsA<Calendar>> CalendarExt for O {
         }
     }
 
-    //fn set_detail_func<P: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(&self, func: /*Unknown conversion*//*Unimplemented*/CalendarDetailFunc, data: P, destroy: /*Unknown conversion*//*Unimplemented*/DestroyNotify) {
-    //    unsafe { TODO: call ffi::gtk_calendar_set_detail_func() }
-    //}
+    fn set_detail_func<P: Fn(&Calendar, u32, u32, u32) -> String + 'static>(&self, func: P) {
```
sdroege: Why does this not return a `GString`, but for parameters we use `GString`? :)

src/auto/cell_layout.rs
```diff
@@ -94,9 +97,31 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
...
+            let tree_model = from_glib_borrow(tree_model);
+            let iter = from_glib_borrow(iter);
+            let callback: &Box_<Option<Q>> = &*(data as *mut _);
+            if let Some(ref callback) = **callback {
+                callback(&cell_layout, &cell, &tree_model, &iter)
+            } else {
+                panic!("cannot get closure...")
+            };
+        }
+        let func = if func_data.is_some() { Some(func_func::<P, Q> as _) } else { None };
```
sdroege: Why does this have a `as _` btw? That case should not have any effect

src/auto/cell_layout.rs
```diff
@@ -94,9 +97,31 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
...
-    //}
+    fn set_cell_data_func<P: IsA<CellRenderer>, Q: Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + Send + Sync + 'static, R: Into<Option<Q>>>(&self, cell: &P, func: R) {
+        let func = func.into();
+        let func_data: Box_<Option<Q>> = Box::new(func.into());
+        unsafe extern "C" fn func_func<P: IsA<CellRenderer>, Q: Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + Send + Sync + 'static>(cell_layout: *mut ffi::GtkCellLayout, cell: *mut ffi::GtkCellRenderer, tree_model: *mut ffi::GtkTreeModel, iter: *mut ffi::GtkTreeIter, data: glib_ffi::gpointer) {
+            let cell_layout = from_glib_borrow(cell_layout);
+            let cell = from_glib_borrow(cell);
+            let tree_model = from_glib_borrow(tree_model);
+            let iter = from_glib_borrow(iter);
+            let callback: &Box_<Option<Q>> = &*(data as *mut _);
```
sdroege: This is wrong. You have a `Box<Option<Q>>` converted to a pointer, which then would be `*mut Option<Q>`. And here you convert that to a `&Box<Option<Q>>`. That's one pointer indirection too much, it should only be `&Option<Q>`.

src/auto/cell_area.rs
```diff
@@ -237,13 +237,37 @@ impl<O: IsA<CellArea>> CellAreaExt for O {
...
     }
 
-    //fn foreach<P: Into<Option</*Unimplemented*/Fundamental: Pointer>>>(&self, callback: /*Unknown conversion*//*Unimplemented*/CellCallback, callback_data: P) {
-    //    unsafe { TODO: call ffi::gtk_cell_area_foreach() }
-    //}
+    fn foreach<P: FnMut(&CellRenderer) -> bool>(&self, callback: P) {
+        let callback_data: Option<P> = callback.into();
+        unsafe extern "C" fn callback_func<P: FnMut(&CellRenderer) -> bool>(renderer: *mut ffi::GtkCellRenderer, data: glib_ffi::gpointer) -> glib_ffi::gboolean {
+            let renderer = from_glib_borrow(renderer);
+            let callback: *mut P = data as *const _ as usize as *mut P;
```
sdroege: This is wrong. The `user_data` is a `&Option<P>`, i.e. `*mut Option<P>` or whatever. You dropped the `Option<_>` part here.

sdroege: Also unclear why it's an `Option<_>` at all as the function is not optional :)

src/auto/flow_box.rs
```diff
@@ -215,7 +215,7 @@ pub trait FlowBoxExt: 'static {
 
 impl<O: IsA<FlowBox>> FlowBoxExt for O {
     //#[cfg(any(feature = "v3_18", feature = "dox"))]
-    //fn bind_model<'a, P: IsA</*Ignored*/gio::ListModel> + 'a, Q: Into<Option<&'a P>>>(&self, model: Q, create_widget_func: /*Unknown conversion*//*Unimplemented*/FlowBoxCreateWidgetFunc, user_data_free_func: /*Unknown conversion*//*Unimplemented*/DestroyNotify) {
+    //fn bind_model<P: Fn(&glib::Object) -> Widget + 'static>(&self, model: /*Ignored*/Option<&gio::ListModel>, create_widget_func: P) {
```
sdroege: Why can't this be generated? Looks like it only need `gio.ListModel` to be added to the manual section of the toml

src/auto/calendar.rs
EPashkin: Because there for trampolines we decode that returning String instead GString better, don't remember reasoning.
So hack added in https://github.com/gtk-rs/gir/blob/master/src/codegen/return_value.rs#L19.

EPashkin: .. and as you say in https://github.com/gtk-rs/gio/pull/183#discussion_r249974199 we need use `&str` as parameter instead GString.

src/auto/cell_area.rs
GuillaumeGomez: It's not anymore. :)

src/auto/cell_layout.rs
GuillaumeGomez: If I don't put it:

```
error[E0308]: mismatched types
   --> src/auto/cell_layout.rs:122:117
    |
122 |             ffi::gtk_cell_layout_set_cell_data_func(self.as_ref().to_glib_none().0, cell.as_ref().to_glib_none().0, func, Box::into_raw(super_callback0) as *mut _, destroy_call4);
    |                                                                                                                     ^^^^ expected fn pointer, found fn item
    |
    = note: expected type `std::option::Option<unsafe extern "C" fn(*mut ffi::GtkCellLayout, *mut ffi::GtkCellRenderer, *mut ffi::GtkTreeModel, *mut ffi::GtkTreeIter, *mut libc::c_void)>`
               found type `std::option::Option<unsafe extern "C" fn(*mut ffi::GtkCellLayout, *mut ffi::GtkCellRenderer, *mut ffi::GtkTreeModel, *mut ffi::GtkTreeIter, *mut libc::c_void) {<O as auto::cell_layout::CellLayoutExt>::set_cell_data_func::func_func::<P, Q>}>`

```

sdroege: Wat? This probably works if you directly put `Some(the_actual_function::<P, Q>)` here instead of first going via a variable, can you confirm?

src/auto/cell_layout.rs
```diff
@@ -94,9 +97,31 @@ impl<O: IsA<CellLayout>> CellLayoutExt for O {
...
+            } else {
+                panic!("cannot get closure...")
+            };
+        }
+        let func = if func_data.is_some() { Some(func_func::<P, Q> as _) } else { None };
+        unsafe extern "C" fn destroy_func<P: IsA<CellRenderer>, Q: Fn(&CellLayout, &CellRenderer, &TreeModel, &TreeIter) + 'static>(data: glib_ffi::gpointer) {
+            let _callback: Box_<Q> = Box_::from_raw(data as *mut _);
+        }
+        let destroy_call4 = Some(destroy_func::<P, Q> as _);
+        let super_callback0: Box_<Option<Q>> = func_data;
```
EPashkin: Conflict with `let _callback: Box_<Q> = Box_::from_raw(data as *mut _);` in destroy_func

src/auto/combo_box.rs
```diff
@@ -409,9 +409,25 @@ impl<O: IsA<ComboBox>> ComboBoxExt for O {
...
+            let callback: &P = &*(data as *mut _);
+            let res = (*callback)(&model, &iter);
+            res.to_glib()
+        }
+        let func = Some(func_func::<P> as _);
+        unsafe extern "C" fn destroy_func<P: Fn(&TreeModel, &TreeIter) -> bool + 'static>(data: glib_ffi::gpointer) {
+            let _callback: Box_<Option<P>> = Box_::from_raw(data as *mut _);
+        }
+        let destroy_call3 = Some(destroy_func::<P> as _);
+        let super_callback0: Box_<P> = func_data;
```
EPashkin: Inverse confilct with destroy_func, maybe bad flag using?

src/auto/cell_layout.rs
sdroege: Well, it does the same if the value is not `None` in this case. If it's `None` you might free a `NULL` pointer instead :)

GuillaumeGomez: I still don't get it...

sdroege: It's `Box<Q>` in the destroy notify but you create a `Box<Option<Q>>` and pass that to it. The types don't match. In the closure trampoline the types match

GuillaumeGomez: Ooooooh!

GuillaumeGomez: Fixed.

EPashkin: Big thanks, we near final 🚀

sdroege:
:+1:

PR #762, opened 2019-01-23
Rename SelectionData get_data_with_length method to get_data

osa1:


osa1:
Ping @sdroege

src/selection_data.rs
```diff
@@ -0,0 +1,19 @@
+// Copyright 2013-2018, The Gtk-rs Project Developers.
```
EPashkin: 2013-2019

EPashkin:
@osa1 Thanks
👍

src/selection_data.rs
```diff
@@ -0,0 +1,19 @@
...
+
+use SelectionData;
+use ffi;
+use glib::translate::*;
+
+impl SelectionData {
+    pub fn get_data(&self) -> Vec<u8> {
+        unsafe {
+            let mut length = mem::uninitialized();
+            let ret = FromGlibContainer::from_glib_none_num(ffi::gtk_selection_data_get_data_with_length(self.to_glib_none().0, &mut length), length as usize);
```
GuillaumeGomez: Will `length` be used **after** its value change from the C call or before? (I'm very curious about this!)

osa1: That's a good question. This code is actually auto-generated, I just copied it from the auto/ directory to here and renamed it. Perhaps it's a good idea to make it explicit by adding one more line that `length` will be used after being initialized.

GuillaumeGomez: What do you think about it @EPashkin @sdroege ?

sdroege: It will only be written to by the C function. See that it's marked as `out` via the annotations, not as `inout`.

sdroege:
:+1:

src/selection_data.rs
GuillaumeGomez: It wasn't my point. The value received by `from_glib_none_mut` will be the one of `length` before its modification in the C function or after?

sdroege: Oh... I see. Evaluation order is defined/guaranteed, yes. Inside-out, left-to-right. In this specific case that means
1. `self.to_glib_none().0`
2. `&mut length`
3. `gtk_selection_data_get_data_with_length()`
4. `length as usize`
5. `from_glib_none_num()`
6. `ret = ...`

GuillaumeGomez: Thanks for the clarification!

EPashkin: As I know, first get_data called, then its result pushed to stack, length pushed and next `from_glib_none_num` called.

sdroege:
@GuillaumeGomez all green apart from the examples being generally broken now :)

GuillaumeGomez:
Oh indeed. Merging then!

PR #767, opened 2019-01-28
Don't box signal callbacks twice

sdroege:
And also change various callbacks to take a generic &Self as first
argument just because we can.

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -16,24 +16,22 @@ use glib_ffi::{self, gboolean, gpointer};
 use {
     Continue,
     ScrollType,
-    SpinButton,
     Widget,
 };
 
 // idle_add and timeout_add fixed to the main thread
 
-unsafe extern "C" fn trampoline(func: gpointer) -> gboolean {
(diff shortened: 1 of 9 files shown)


src/signal.rs
```diff
@@ -206,66 +202,64 @@ mod spin_button {
...
     }
 
     unsafe extern "C" fn input_trampoline(this: *mut GtkSpinButton,
                                           new_value: *mut c_double,
-                                          f: &&(Fn(&SpinButton) -> Option<Result<f64, ()>> + 'static))
-                                          -> c_int {
-        match f(&from_glib_borrow(this)) {
+                                          f: &F)
+                                          -> c_int
+    T: IsA<SpinButton> {
```
EPashkin: No `where` ?

sdroege: Yeah I can't compile things currently until @GuillaumeGomez is done with his regen PRs :)

EPashkin: *3 and many other errors,

sdroege: Yeah we can fix and merge this one once the other things are done :)

EPashkin:
Canceled CI

PR #768, opened 2019-01-28
Add TreeSelectionExtManual to set selection func

SamWhited:
I noticed after writing this that if I regenerated the bindings from the gir files that this now exists directly in `TreeSelectionExt`. However, I got segfaults every time I tried to use it.

I wasn't sure what the policy was, or if its inclusion was deliberate or something I broke (because I thought it was ignored before), so I figured I'd submit this (which is working fine for me in a project using a gtk-rs from source) in case you wanted the code instead of the auto-generated version.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -269,6 +269,7 @@ mod text_buffer;
 mod text_iter;
 mod tree_model_filter;
 mod tree_row_reference;
+mod tree_selection;
 mod tree_sortable;
 mod tree_path;
 mod tree_store;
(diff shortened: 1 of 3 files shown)


sdroege:
The code generator was producing broken code until yesterday evening for that function, that's why you got crashes. And until very recently it wasn't able to generate code for functions taking callbacks at all, that's why it was not generated before :)

Thanks for working on this, but this is most likely covered by https://github.com/gtk-rs/gtk/pull/760 already. Once that is merged (and all the gtk-rs crates are in a consistent state again), can you confirm that this also works correctly for what you're doing?

sdroege:
It's merged now :)

sdroege:
@GuillaumeGomez Please close this one, we autogenerate the function now

PR #769, opened 2019-01-29
Fix CI

GuillaumeGomez:
cc @EPashkin @sdroege


sdroege:
Please just merge all PRs that do only this change, no need to wait for review :)

PR #771, opened 2019-01-29
Support non-GNU versions of make

SamWhited:
I have bmake (NetBSD's make, which is mostly just POSIX like and doesn't support many GNU-isms) installed on most of my machines. The makefile for this project mostly works, but `GIR_SRC` wasn't being set correctly because the `shell` expansion is a GNU-ism.

This sets those files in a more standard way that other versions of Make (including GNU's) will also pick up on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,6 @@
 GIR = gir/target/bin/gir
-GIR_SRC = gir/Cargo.toml gir/Cargo.lock gir/build.rs $(shell find gir/src -name '*.rs')
+GIR_SRC != find gir/src -name '*.rs'
+GIR_SRC += gir/Cargo.toml gir/Cargo.lock gir/build.rs
 GIR_FILES = gir-files/Gtk-3.0.gir
 
 # Run `gir` generating the bindings


GuillaumeGomez:
`GIR_SRC != find gir/src -name '*.rs'` doesn't work for me.

SamWhited:
@GuillaumeGomez what sort of system are you on or what version of Make? That appears to work for me on both GNUMake and NetBSD make; I *think* (though I'd have to double check) that that's part of POSIX, so it should work on all versions

GuillaumeGomez:
Linux mint so the problem will be the same on debian and ubuntu as well.

For the Make command version: `GNU Make 4.1`

GuillaumeGomez:
Ah nevermind. Seems to be working. Don't know what I failed earlier...

SamWhited:
*whew* I was pretty sure != has been around in GNUMake for a lot longer than 4.1, so I was really getting confused digging through the change log and commits :)

EPashkin:
@SamWhited Thanks, Works on windows+msys64 too
👍

EPashkin:
@GuillaumeGomez, @sdroege IMHO this PR can be merged

GuillaumeGomez:
For me as well, just waiting for @sdroege's opinion on this, just in case.

sdroege:
Fine with me

GuillaumeGomez:
Then I merge. Thanks @SamWhited !

PR #779, opened 2019-02-10
Bring Travis and AppVeyor configurations in sync with the GLib one

sdroege:


.travis.yml
```diff
@@ -53,15 +54,9 @@ script:
...
   - ./check_init_asserts
-  - if [ "$TRAVIS_RUST_VERSION" == "nightly" ] && [ "$GTK" == "3.4" ]; then
-    rustup component add clippy-preview;
-    cargo clippy;
-    fi
-  - cargo test --features embed-lgpl-docs --jobs 1
   - cargo doc --features "dox"
-  # catch any sneaked in lgpl docs
-  - cargo build --features purge-lgpl-docs --jobs 1
-  - git diff -R --exit-code
```
EPashkin: All test at up need stay

sdroege: You should've said that before all the other ones are merged. I asked if that's fine in the other PRs because the GLib one didn't have any of this and nobody replied to that and then it was merged.

At this point I'd prefer to have all this merged and then we can as a separate step look at what can be improved in the CI configuration. That way it does not go out of sync between all the different repos again.

EPashkin: Sorry for missing this.
All crates with embed-lgpl-docs must have tests for it.
Good that cargo allow --features ",v3_16," without any question
`cargo test --features "embed-lgpl-docs,$FEATURES" --jobs 1`
Please just start write right with this PR.

sdroege: You redo the other ones then? I do gdk and the ones I didn't touch yet still. My patience with CI things is getting to an end :)

For doing things via Docker btw:
https://docs.travis-ci.com/user/docker/
https://hub.docker.com/_/fedora/?tab=tags

Seems a bit more complicated than with gitlab-ci but still manageable. Then we could at least test the latest versions.

sdroege: > I do gdk and the ones I didn't touch yet still

I meant gdk-pixbuf, not gdk.

sdroege:
@EPashkin @GuillaumeGomez that appveyor failure looks transient, can you retry it or just merge?

GuillaumeGomez:
Weird error. Let's merge for now. Thanks!

EPashkin:
Seems we broke test on win32 again 😢

GuillaumeGomez:
So it wasn't transient? :-/ Urg...

EPashkin:
We just not tested gtk on apveyor before, maybe because win32 test always fails 😉

GuillaumeGomez:
That might explain it! Haha

sdroege:
I think the problem is DLL hell. There seem to be cygwin and mingw DLLs mixed in the PATH here, some of them at least coming from git.

How can this be cleaned up?

sdroege:
@EPashkin any ideas? :)

PR #789, opened 2019-02-22
Improve Dialog API

GuillaumeGomez:
cc @sdroege @EPashkin @philn

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -758,6 +758,12 @@ manual_traits = ["DialogExtManual"]
     name = "get_content_area"
         [object.function.return]
         nullable = false
+    [[object.function]]
+    name = "run"
+    ignore = true
+    [[object.function]]
+    name = "get_response_for_widget"
+    ignore = true
 
 [[object]]
 name = "Gtk.Editable"
--- a/src/tree_sortable.rs
+++ b/src/tree_sortable.rs
@@ -8,11 +8,13 @@ use glib::object::{Cast, IsA};
 use glib::translate::*;
 use std::mem::{self, transmute};
 use std::cmp::Ordering;
+use std::fmt;
 
 use glib_ffi::gpointer;
 use {TreeIter, TreeModel, TreeSortable};
 use ffi::{GtkTreeIter, GtkTreeModel};
 
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Ord, PartialOrd, Hash)]
 pub enum SortColumn {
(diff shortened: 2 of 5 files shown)


Gir.toml
```diff
@@ -758,6 +758,12 @@ manual_traits = ["DialogExtManual"]
     name = "get_content_area"
         [object.function.return]
         nullable = false
+    [[object.function]]
+    name = "run"
+    ignore = true
+    [[object.function]]
+    name = "get_response_for_widget"
+    ignore = true
```
sdroege: Don't we have a way to override the return type?

sdroege:
Good otherwise

Gir.toml
GuillaumeGomez: Doesn't seem like we do. :-/

EPashkin:
👍

GuillaumeGomez:
I need to update gtk examples as well.

src/tree_sortable.rs
```diff
@@ -49,6 +51,15 @@ impl FromGlib<i32> for SortColumn {
     }
 }
 
+impl fmt::Display for SortColumn {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        write!(f, "SortColumn::{}", match *self {
+            SortColumn::Default => "Default",
+            SortColumn::Index(_) => "Index",
```
EPashkin: Don't print column number?

GuillaumeGomez: `Debug` does it. :)

sdroege: This is not a `Display` impl anyway but you know my opinion about that ;)

In this case it should print things like "Default column", "Column at index {}", etc. and not basically the same as the `Debug` impl.

GuillaumeGomez: Yes I know. But the current way of implementing `Display` is this one. However I agree this is not very nice... I'm open to suggestions for next release!

sdroege: My suggestion is that you can't automatically implement `Display` as it requires context specific knowledge about the type that we don't have.

For this specific type a correct `Display` implementation would be the one I mentioned.

GuillaumeGomez:
CI passed so I merge.

About the `Display` implementations issue: we need to make something official. I don't fully agree with @sdroege's views on it and I think that @EPashkin might have some ideas about it. I'll open the issue on gir directly.

PR #791, opened 2019-02-22
[regen] No more into

GuillaumeGomez:
Not to be merged.


GuillaumeGomez:
Updated.

GuillaumeGomez:
cc @sdroege @EPashkin

EPashkin:
Don't want change manual parts too?
Thanks and 👍 for this batch anyway

GuillaumeGomez:
That's a good point! I'll go over the manual part as well.

src/auto/accel_label.rs
```diff
@@ -53,9 +53,9 @@ pub trait AccelLabelExt: 'static {
 
     fn set_accel(&self, accelerator_key: u32, accelerator_mods: gdk::ModifierType);
 
-    fn set_accel_closure<'a, P: Into<Option<&'a glib::Closure>>>(&self, accel_closure: P);
+    fn set_accel_closure(&self, accel_closure: Option<&glib::Closure>);
 
-    fn set_accel_widget<'a, P: IsA<Widget> + 'a, Q: Into<Option<&'a P>>>(&self, accel_widget: Q);
+    fn set_accel_widget<P: IsA<Widget>>(&self, accel_widget: &P);
```
GuillaumeGomez: That should be `Option<&P>`, not `&P`. The PR is invalid. :)

GuillaumeGomez:
Updated!

src/auto/accel_label.rs
EPashkin: Hm, this still not fixed, seems update go wrong as it have right in my regen.

GuillaumeGomez: I think my rebase went crazy. I'll restart the whole process.

EPashkin:
Thanks, 👍

src/auto/about_dialog.rs
```diff
@@ -472,97 +463,97 @@ impl<O: IsA<AboutDialog>> AboutDialogExt for O {
...
 where P: IsA<AboutDialog> {
-    let f: &F = transmute(f);
+    let f: &F = &*(f as *const F);
     f(&AboutDialog::from_glib_borrow(this).unsafe_cast())
 }
 
 unsafe extern "C" fn notify_comments_trampoline<P, F: Fn(&P) + 'static>(this: *mut ffi::GtkAboutDialog, _param_spec: glib_ffi::gpointer, f: glib_ffi::gpointer)
 where P: IsA<AboutDialog> {
-    let f: &F = transmute(f);
+    let f: &F = &*(f as *const F);
```
sdroege: You probably want to remove the `transmute` in the manual code too. There are 5 cases like this in the manual code that can be changed, leaving only the cases where the trampolines are transmuted (I'm not sure why exactly this is needed)

sdroege:
> +1 except for the failing doctests

... and the remaining transmutes

EPashkin:
👍

PR #792, opened 2019-03-01
Add a special type for add_tick_callback's return value

### LINKED ISSUE #761: gtk_widget_add_tick_callback need special return type
Same as idle_add return SourceId, this will need special type, that consumed by tk_widget_remove_tick_callback for safety reason

sfanxiang:
For safety reasons, the value consumed by remove_tick_callback needs to be created by add_tick_callback. FromGlib and ToGlib are not implemented for the same reason.

First contribution, so I may have missed things.

Fixes #761, cc @EPashkin, @sdroege.

### DIFF
--- a/src/widget.rs
+++ b/src/widget.rs
@@ -5,9 +5,10 @@
 use std::mem::transmute;
 use std::ptr;
 
-use glib::object::{Cast, IsA};
+use glib::object::{Cast, IsA, WeakRef};
 use glib::signal::{SignalHandlerId, connect_raw};
 use glib::translate::*;
+use glib::ObjectExt;
 use glib_ffi::gboolean;
 use gdk::{DragAction, Event, ModifierType};
 use gdk;
@@ -16,12 +17,28 @@ use pango;
 use ffi;
 
 use {
+    Continue,
     DestDefaults,
     Inhibit,
     Rectangle,
     TargetEntry,
     Widget,
 };
+
+pub struct TickCallbackId {
+    id: u32,
+    widget: WeakRef<Widget>,
+}
+
+impl TickCallbackId {
+    pub fn remove(self) {
+        if let Some(widget) = self.widget.upgrade() {
+            unsafe {
+                ffi::gtk_widget_remove_tick_callback(widget.to_glib_none().0, self.id);
+            }
+        }
+    }
+}
 
 pub trait WidgetExtManual: 'static {
     fn drag_dest_set(&self, flags: DestDefaults, targets: &[TargetEntry], actions: DragAction);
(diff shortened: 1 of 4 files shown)


GuillaumeGomez:
Nope, seems good globally after a quick look. I'll go back to it later on. cc @EPashkin @sdroege

src/widget.rs
```diff
@@ -40,6 +44,13 @@ pub trait WidgetExtManual: 'static {
 
     fn connect_unmap_event<F: Fn(&Self, &Event) -> Inhibit + 'static>(&self, f: F) -> SignalHandlerId;
 
+    fn add_tick_callback<P: Fn(&Widget, &gdk::FrameClock) -> bool + 'static>(
```
sdroege: Should return a `glib::Continue` instead of a `bool`

sdroege:
Looks good otherwise

src/widget.rs
```diff
@@ -108,6 +119,48 @@ impl<O: IsA<Widget>> WidgetExtManual for O {
...
+            ffi::gtk_widget_add_tick_callback(
+                self.as_ref().to_glib_none().0,
+                callback,
+                Box::into_raw(super_callback0) as *mut _,
+                destroy_call3,
+            )
+        })
+    }
+
+    fn remove_tick_callback(&self, id: TickCallbackId) {
```
EPashkin: Not better add this function to `TickCallbackId` ?

EPashkin: And please add `#[allow(clippy::needless_pass_by_value)]` if clippy nags.

EPashkin: @GuillaumeGomez, @sdroege You agree to moving remove function to type itself?

sdroege: Works for me

src/widget.rs
EPashkin: Hm, body better added to `impl<O: IsA<Widget>> WidgetExtManual for O {`
Also in current case IMHO first callback's parameter better be `&Self`.

EPashkin: Sorry about "moving body" part, its already there, I mistakes it with multi-line definition

sfanxiang: Done both

src/widget.rs
sfanxiang: Done, but `TickCallbackId` no longer implements Debug, Eq, PartialEq, and Send because WeakRef<Widget> doesn't.

src/widget.rs
```diff
@@ -16,13 +17,30 @@ use pango;
...
 use {
+    Continue,
     DestDefaults,
     Inhibit,
     Rectangle,
     TargetEntry,
     Widget,
 };
 
+/// The id of tick callback returned by `add_tick_callback`.
```
EPashkin: @sfanxiang Sorry, we store comment in other crate https://github.com/gtk-rs/lgpl-docs,
so please, remove it after https://github.com/gtk-rs/lgpl-docs/pull/52 merged, it fix CI doctest

sfanxiang: Done

EPashkin:
@sfanxiang Thanks.
👍 after travis passed.

sfanxiang:
@EPashkin 
> after travis passed.

Travis passed, but idk why appveyor BITS=32 failed.

GuillaumeGomez:
Thanks!

PR #800, opened 2019-03-23
add doc rule

GuillaumeGomez:
cc @EPashkin @sdroege 

Since I'm updating the docs, I add some rules to make my life easier (automation, here we come!).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,9 @@ GIR_FILES = gir-files/Gtk-3.0.gir
 # Run `gir` generating the bindings
 gir : src/auto/mod.rs
 
+doc: $(GIR) $(GIR_FILES)
+	$(GIR) -m doc -c Gir.toml
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -6,6 +6,9 @@ GIR_FILES = gir-files/Gtk-3.0.gir
 # Run `gir` generating the bindings
 gir : src/auto/mod.rs
 
+doc: $(GIR) $(GIR_FILES)
+	$(GIR) -m doc -c Gir.toml
```
EPashkin: Don't want add "-o ../lgpl-docs\gtk\" ?

EPashkin:
👍

Makefile
GuillaumeGomez: I'd prefer not: we generate a `vendor.md`, then we can just grab it. We don't know for sure if the one starting the command has the `lgpl-docs` repository locally as well. Let's keep it simple. ;)

PR #810, opened 2019-05-06
Remove manually implemented Gtk.Socket

moggiesir:
The manual implementation isn't needed since the auto-generated one has more functionality.


GuillaumeGomez:
cc @EPashkin @sdroege

sdroege:
:+1:

EPashkin:
Old version was from beginning, we just forgot to remove it
Only difference is `not(windows)` for generated and `linux` for manual.
@moggiesir Thanks
👍

PR #812, opened 2019-05-11
Remove traits for disguised types without children

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/760

cc @GuillaumeGomez, @sdroege


sdroege:
Seems good

EPashkin:
Updated, added builders

cc @GuillaumeGomez, @sdroege, @antoyo

Gir.toml
```diff
@@ -214,6 +214,146 @@ generate = [
     "Gtk.WrapMode",
 ]
 
+generate_builders = [
```
GuillaumeGomez: I still think we should call this list "builders". :p

antoyo:
Looks good to me.
I was under the impression that @sdroege wanted to enable builders on a case-by-case basis or by need. What do you think @sdroege?

sdroege:
Case-by-case basis, yes. And I want that to be opt-in.

I'd be fine with enabling it for all widgets though, as long as it results in a useful API. So please carefully review the API of each builder you enable if the properties they have make any sense like that :)

antoyo:
The API of each builder in this PR makes sense.

EPashkin:
Seems builders need more working:
```
    let window = gtk::ApplicationWindowBuilder::new()
.application(application.clone())
.build();
```
Firstly it need `clone` as setter want `Application` but not `&Application` or `&IsA<Application>`.
Secondly it produce many colored criticals:
```
(builders.exe:2172): GLib-GObject-CRITICAL **: 21:04:58.413: g_object_set_qdata: assertion 'G_IS_OBJECT (object)' failed

(builders.exe:2172): GLib-GObject-WARNING **: 21:04:58.420: instance with invalid (NULL) class pointer

(builders.exe:2172): GLib-GObject-CRITICAL **: 21:04:58.425: g_signal_handlers_destroy: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed

(builders.exe:2172): GLib-GObject-CRITICAL **: 21:04:58.429: g_object_unref: assertion 'G_IS_OBJECT (object)' failed

(builders.exe:2172): GLib-GObject-WARNING **: 21:04:58.435: instance with invalid (NULL) class pointer

(builders.exe:2172): GLib-GObject-CRITICAL **: 21:04:58.439: g_signal_handlers_destroy: assertion 'G_TYPE_CHECK_INSTANCE (instance)' failed
error: process didn't exit successfully: `D:\eap\rust\0\./target\debug\builders.exe` (exit code: 0xc0000005, STATUS_ACCESS_VIOLATION)
```

sdroege:
@EPashkin Do you use the latest version of the glib crate with my `g_object_new()` fix? That looks very similar to that one.

EPashkin:
@sdroege Thanks, I really not updated glib.
Now even full version works fine:
```
    let window = gtk::ApplicationWindowBuilder::new()
.application(application.clone())
.title("First GTK+ Program")
.border_width(10)
.window_position(gtk::WindowPosition::Center)
.default_width(350)
.default_height(70)
.build();
```

EPashkin:
Updated, builders still configured with array, except for "Gtk.TreeView"

EPashkin:
*still but intentional.
PR IMHO is ready if CI agrees.

EPashkin:
Updated with gtk-rs/gir#775

GuillaumeGomez:
CI is breaking on `1.31`. You need to rename the import into something else. I vote for "_whateveeeeeeeer". XD

EPashkin:
My bad, I checked on stable only.
How about "GlibObjectType" ?

sdroege:
@EPashkin see https://github.com/gtk-rs/cairo/pull/260/files#r285871201

sdroege:
>  How about "GlibObjectType" ?

Works for me though

Gir.toml
```diff
@@ -214,6 +214,144 @@ generate = [
     "Gtk.WrapMode",
 ]
 
+builders = [
+    "Gtk.AboutDialogBuilder",
```
sdroege: Hmm why do we not refer to the actual type name here but the Builder type? Will we also get `Gtk.BuilderBuilder`? ;)

It seems nicer to specify the actual types instead of the Builder type, that's what we do everywhere else in the configuration too.

EPashkin:
All current variants about `use glib::object::ObjectType as _` failure in rust 1.31.0:
rename to "GlibObjectType",
rename to "ObjectType_" (same as we do with Box),
make minimum supported version 1.34.2 (or 1.34.0) and change .travis.yml (from https://github.com/gtk-rs/cairo/pull/260/files#r285871201)

Gir.toml
EPashkin: Gir support both way, for `builders` IMHO having suffix is more reasonable.
Although current version will have problem with `Gtk.Builder` instead `Gtk.BuilderBuilder`,
but it have only one `translation_domain: Option<String>`, so I don't generate it anyway.

EPashkin:
Updated with `use glib::object::ObjectType as ObjectType_;`

sdroege:
@GuillaumeGomez Please merge

GuillaumeGomez:
Thanks @EPashkin !

PR #813, opened 2019-05-11
Remove main_quit() assert

antoyo:


EPashkin:
Something is broken with old code?

antoyo:
Well, this assert does not seem to make sense.
For instance, if you manually run the loop like this:
```rust
while gtk::main_iteration() {
}
```
Calling `gtk::main_quit()` will panic in this case.

EPashkin:
Thanks for explanation, 👍

EPashkin:
`gtk_main_quit` really works in this cycle?
By https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkmain.c#L1370 it just exit without main loop

antoyo:
I'm not sure what you mean, but it will make `gtk_main_iteration()` returns `false`: https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkmain.c#L1431

antoyo:
Ah, I think I understand.
I'll test.

antoyo:
So, you were right @EPashkin, `gtk_main_quit()` will output a CRITICAL error.
So, I'm closing this PR.

PR #814, opened 2019-05-12
Allow for GtkApplication to be subclasses

alatiera:
This expands further upon the gio MR that does the same
for GApplictiona.

https://github.com/gtk-rs/gio/pull/209

### DIFF
--- a/src/subclass/application.rs
+++ b/src/subclass/application.rs
@@ -1,0 +1,81 @@
+use gtk_sys;
+
+use glib::translate::*;
+
+use glib::subclass::prelude::*;
+
+use Application;
+use ApplicationClass;
+use Window;
+
+pub trait GtkApplicationImpl: GtkApplicationImplExt + gio::subclass::prelude::ApplicationImpl + 'static {
+    fn window_added(&self, application: &Application, window: &Window) {
+        self.parent_window_added(application, window)
+    }
+
+    fn window_removed(&self, application: &Application, window: &Window) {
+        self.parent_window_removed(application, window)
+    }
+}
+
+pub trait GtkApplicationImplExt {
+    fn parent_window_added(&self, application: &Application, window: &Window);
+    fn parent_window_removed(&self, application: &Application, window: &Window);
+}
+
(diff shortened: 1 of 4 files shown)


src/subclass/mod.rs
```diff
@@ -0,0 +1,5 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+pub mod application;
```
alatiera: this nees a prelude

src/subclass/application.rs
```diff
@@ -0,0 +1,81 @@
...
+    fn window_added(&self, application: &Application, window: &Window) {
+        self.parent_window_added(application, window)
+    }
+
+    fn window_removed(&self, application: &Application, window: &Window) {
+        self.parent_window_removed(application, window)
+    }
+}
+
+pub trait ApplicationImplExt {
```
sdroege: All the traits should probably get `Gtk` as prefix

sdroege:
Looks good otherwise :)

src/subclass/application.rs
alatiera: Done :)

Gir.toml
```diff
@@ -426,6 +426,7 @@ status = "generate"
 name = "Gtk.Application"
 status = "generate"
 trait_name = "GtkApplicationExt"
+subclassing = true
```
EPashkin: @sdroege This options now not used, can we delete it later?

sdroege: I guess we can delete it now already :)

EPashkin: Ok, I will do it.
@alatiera Please remove this line

alatiera: done :)

sdroege:
@GuillaumeGomez good to go then :)

GuillaumeGomez:
Thanks @alatiera!

PR #815, opened 2019-05-20
Reorder "::insert_text()" params according to autogenerated functions

apiraino:
Hi!

During the latest Gnome Hackfest :^) I've noticed what I believe is a tiny inconsistency in the parameters ordering for the same `insert_text` function in different widgets, e.g.:

- (autogenerated) [gtk::EditableExt::insert_text](http://gtk-rs.org/docs/gtk/trait.EditableExt.html#tymethod.insert_text)
- [gtk::ComboBoxTextExt::insert_text](http://gtk-rs.org/docs/gtk/trait.ComboBoxTextExt.html#tymethod.insert_text)
- [gtk::EntryBuffer::insert_text](http://gtk-rs.org/docs/gtk/struct.EntryBuffer.html#method.insert_text)

This is just nitpicking, but I thought it could make sense to respect the parameters ordering from the autogenerated functions (since we have less control on them).

One thing I'm asking for advice: after this change tests don't fail, which I kind of expected.

There may be other functions that could benefit from some uniform signature (but didn't investigate deep enough).

Your thoughts on this?

thanks!

### DIFF
--- a/src/auto/combo_box_text.rs
+++ b/src/auto/combo_box_text.rs
@@ -57,7 +57,7 @@ pub trait ComboBoxTextExt: 'static {
 
     fn insert(&self, position: i32, id: Option<&str>, text: &str);
 
-    fn insert_text(&self, position: i32, text: &str);
+    fn insert_text(&self, text: &str, position: i32);
 
     fn prepend(&self, id: Option<&str>, text: &str);
 
@@ -93,7 +93,7 @@ impl<O: IsA<ComboBoxText>> ComboBoxTextExt for O {
         }
     }
 
-    fn insert_text(&self, position: i32, text: &str) {
+    fn insert_text(&self, text: &str, position: i32) {
         unsafe {
             gtk_sys::gtk_combo_box_text_insert_text(self.as_ref().to_glib_none().0, position, text.to_glib_none().0);
         }
--- a/src/entry_buffer.rs
+++ b/src/entry_buffer.rs
@@ -64,7 +64,7 @@ impl EntryBuffer {
         unsafe { from_glib_none(gtk_sys::gtk_entry_buffer_get_text(self.to_glib_none().0)) }
     }
 
-    pub fn insert_text(&self, position: u16, chars: &str) -> u16 {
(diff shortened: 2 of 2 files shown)


src/auto/combo_box_text.rs
```diff
@@ -57,7 +57,7 @@ pub trait ComboBoxTextExt: 'static {
 
     fn insert(&self, position: i32, id: Option<&str>, text: &str);
```
GuillaumeGomez: In here, the position comes first too. :)

GuillaumeGomez:
Hi @apiraino and thanks for the PR!

This is a good point, API should be constant in the order of its parameters. Unfortunately, you can't just switch parameters like this in the generated code (inside the `src/auto` folder) because it'll be removed as soon as it's regenerated.

Actually, that'd be a nice add to set parameters position in gir. What do you think about this @EPashkin and @sdroege ?

EPashkin:
`Editable` only one has in-out `position`, IMHO this reason why it have different order.
About reordering: I don't think that is a good idea.

apiraino:
uhm .. yes, we don't want to touch autogenerated sources (although that means we will live with inconsistencies from GTK itself), such as:
```
src/auto/editable.rs
39:    fn insert_text(&self, new_text: &str, position: &mut i32);

src/auto/combo_box_text.rs
60:    fn insert_text(&self, position: i32, text: &str);
```

Probably it's not worth the effort, then?

GuillaumeGomez:
For me it is but it's up to debate.

sdroege:
> src/auto/editable.rs
> 39:    fn insert_text(&self, new_text: &str, position: &mut i32);

This is an `inout` parameter while the other is an `in` parameter, so there are even more difference here. And the order given in both is from the C API. I don't think we want to change the order compared to the C API.

sdroege:
Even more, reordering the arguments would make things a bit more tricky for the documentation too.

GuillaumeGomez:
@sdroege True, completely forgot about that...

apiraino:
I think there are enough arguments to safely close the discussion and mark this issue as invalid.

Thanks for the explanations!

PR #819, opened 2019-05-26
Get entry

### LINKED ISSUE #772: EntryCompletionExt::get_entry() should return Option<Entry>, not Option<Widget>

GuillaumeGomez:
Fixes #772.

I was sure that we could handle cast directly inside the "from_glib_*" functions but I was surprised that we couldn't. But I'm wondering with my manual code is the best that we could write. I'm pretty sure I missed something...

cc @sdroege @EPashkin


src/entry_completion.rs
```diff
@@ -0,0 +1,27 @@
...
+    fn get_entry(&self) -> Option<Entry>;
+}
+
+impl<O: IsA<EntryCompletion>> EntryCompletionExtManual for O {
+    fn get_entry(&self) -> Option<Entry> {
+        let widget: Option<Widget> = unsafe {
+            from_glib_none(gtk_sys::gtk_entry_completion_get_entry(self.as_ref().to_glib_none().0))
+        };
+        match widget {
+            Some(w) => w.downcast().ok(),
```
sdroege: This should probably panic. If it's not an entry then things are broken.

sdroege:
> I was sure that we could handle cast directly inside the "from_glib_*" functions

How would that look like?

src/entry_completion.rs
GuillaumeGomez: I don't like panicking in library calls. :-/

sdroege: `from_glib_none()` also panics if the thing there is of the wrong type. A panic is the only sensible thing to do here if basic assumptions are wrong, it's nothing a user can handle in a meaningful way.

GuillaumeGomez: Yes but in here, it comes from us in a "normal" way supposed to be fully working.

sdroege: I don't understand what you mean. If the widget there is not an entry, then there's a bug in GTK somewhere and the basic assumptions of this API are broken. That's nothing the user can handle in a meaningful way.

GuillaumeGomez: Isn't it exactly why we have the `Result` type in rust?

GuillaumeGomez: Not from my point of view but I'll replace with a call to `expect`.

src/entry_completion.rs
```diff
@@ -0,0 +1,27 @@
...
+pub trait EntryCompletionExtManual: 'static {
+    fn get_entry(&self) -> Option<Entry>;
+}
+
+impl<O: IsA<EntryCompletion>> EntryCompletionExtManual for O {
+    fn get_entry(&self) -> Option<Entry> {
+        let widget: Option<Widget> = unsafe {
+            from_glib_none(gtk_sys::gtk_entry_completion_get_entry(self.as_ref().to_glib_none().0))
+        };
+        match widget {
```
sdroege: Could become
```rust
Option::<Widget>::from_glib_none(...).map(|widget| widget.downcast().expect("...")
```
which is shorter and easier to read IMHO :)

GuillaumeGomez:
Updated.

sdroege:
:+1:

EPashkin:
@GuillaumeGomez Thanks and 👍 after regen check fix

Issue #821, opened 2019-06-03
[Bug] - 'libgtk-3 was configured with `--enable-debug=no`


Edu4rdSHL:
I'm facing an issue with my project, it was working fine but stopped running with the following message:

```
thread 'main' panicked at 'libgtk-3 was configured with `--enable-debug=no`. See https://github.com/gtk-rs/gtk/issues/270 for details', /home/sechacklabs/.cargo/registry/src/github.com-1ecc6299db9ec823/gtk-0.6.0/src/rt.rs:141:13
note: Run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```
I was seeing the code of the gtk crate and seems that it looks in the system library for get the debug flag https://github.com/gtk-rs/gtk/blob/master/src/rt.rs#L139, I'm using ArchLinux and doesn't see the  --enable-debug=no flag in the [PKGBUILD for gtk3](https://git.archlinux.org/svntogit/packages.git/tree/repos/testing-x86_64/PKGBUILD?h=packages/gtk3#n42).

What can be the cause?

JFYI: new gtk projects are facing the same issue.

EDIT:

I tried `cargo clean` already, removed the cached crates and recompile again.

Edu4rdSHL:
Downgrading from [testing] fix the issue.

sdroege:
> Downgrading from [testing] fix the issue.

What was the actual problem?

heftig:
GTK3 was built using meson with `--buildtype=release`, which is equivalent to the old `--enable-debug=no`.

sdroege:
That seems like a bug in the meson build, see the old documentation and also https://github.com/gtk-rs/gtk/issues/270

> Thus --enable-debug=no should not be used for stable releases of GLib.

heftig:
That said, buildtypes `plain` and `minsize` (anything but `debug` and `debugoptimized`) are also broken.

You check for `G_ENABLE_DEBUG` being enabled, which it isn't. However, what you actually care about is `G_DISABLE_CHECKS` being disabled, which it is.

sdroege:
Indeed. There's no way to check for that a runtime though :)

Maybe the check should just be removed from `rt.rs`, and people running builds with `G_DISABLE_CHECKS` will just have to live with the consequences.

sdroege:
So, this should be reopened @GuillaumeGomez .

[...]

PR #822, opened 2019-06-06
Remove pre-init assertion to check for non-debug GTK builds

sdroege:
While previously checking for this could be used to determine if GTK was
built with assertions enabled or not, this is not true anymore with
later versions of GTK. There is no way to know if assertions were
disabled or not at runtime.

While disabling assertions easily causes memory unsafety in GTK code,
it is nothing that can be disabled accidentally. Users deploying and
running such code will have to live with the consequences.

Fixes https://github.com/gtk-rs/gtk/issues/821

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -95,7 +95,7 @@ pub fn init() -> Result<(), glib::BoolError> {
         panic!("Attempted to initialize GTK from two different threads.");
     }
     unsafe {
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
:+1:

sdroege:
@GuillaumeGomez All green except for the well-known Windows failure :)

GuillaumeGomez:
Waiting for @EPashkin approval before.

EPashkin:
@sdroege Thanks 👍

GuillaumeGomez:
Thanks!

PR #826, opened 2019-06-13
pad_action_entry: Fix dangling pointer

sfanxiang:
cc @GuillaumeGomez @sdroege

### DIFF
--- a/src/pad_action_entry.rs
+++ b/src/pad_action_entry.rs
@@ -4,50 +4,53 @@
 
 use glib::translate::*;
 use gtk_sys;
+use std::ffi::CStr;
 use PadActionType;
 
-#[repr(C)]
-pub struct PadActionEntry(gtk_sys::GtkPadActionEntry);
+#[derive(Debug, Clone)]
+pub struct PadActionEntry {
+    pub(crate) type_: PadActionType,
+    pub(crate) index: i32,
+    pub(crate) mode: i32,
+    pub(crate) label: String,
+    pub(crate) action_name: String,
+}
 
 impl PadActionEntry {
-    pub fn new(type_: PadActionType, index: i32, mode: i32, label: &str,
-               action_name: &str) -> PadActionEntry {
+    pub fn new(
+        type_: PadActionType,
+        index: i32,
+        mode: i32,
+        label: &str,
+        action_name: &str,
+    ) -> PadActionEntry {
         assert_initialized_main_thread!();
-        PadActionEntry(gtk_sys::GtkPadActionEntry {
-            type_: type_.to_glib(),
+        PadActionEntry {
+            type_,
             index,
             mode,
-            label: label.to_glib_none().0,
-            action_name: action_name.to_glib_none().0,
-        })
+            label: label.to_owned(),
+            action_name: action_name.to_owned(),
+        }
     }
 
     pub fn get_type(&self) -> PadActionType {
-        from_glib(self.0.type_)
+        self.type_
     }
 
     pub fn get_index(&self) -> i32 {
-        self.0.index
+        self.index
     }
 
     pub fn get_mode(&self) -> i32 {
-        self.0.mode
+        self.mode
     }
 
-    pub fn get_label(&self) -> Option<String> {
-        unsafe { from_glib_none(self.0.label) }
+    pub fn get_label(&self) -> Option<&str> {
+        Some(&self.label)
     }
 
-    pub fn get_action_name(&self) -> Option<String> {
-        unsafe { from_glib_none(self.0.label) }
-    }
-}
-
-#[doc(hidden)]
-impl ToGlib for PadActionEntry {
-    type GlibType = gtk_sys::GtkPadActionEntry;
-
-    fn to_glib(&self) -> gtk_sys::GtkPadActionEntry {
-        self.0
+    pub fn get_action_name(&self) -> Option<&str> {
+        Some(&self.action_name)
     }
 }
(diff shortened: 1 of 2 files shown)


src/pad_action_entry.rs
```diff
@@ -4,42 +4,48 @@
 
 use glib::translate::*;
 use gtk_sys;
+use std::ffi::CStr;
 use PadActionType;
 
-#[repr(C)]
-pub struct PadActionEntry(gtk_sys::GtkPadActionEntry);
+pub struct PadActionEntry {
```
sdroege: `#[derive(Debug, Clone)]` :)

src/pad_action_entry.rs
```diff
@@ -48,6 +54,16 @@ impl ToGlib for PadActionEntry {
...
+        gtk_sys::GtkPadActionEntry {
+            type_: self.type_.to_glib(),
+            index: self.index,
+            mode: self.mode,
+            label: CStr::from_bytes_with_nul(self.label.as_bytes())
+                .expect("Unexpected nul")
+                .as_ptr() as *mut _,
+            action_name: CStr::from_bytes_with_nul(self.action_name.as_bytes())
+                .expect("Unexpected nul")
+                .as_ptr() as *mut _,
```
sdroege: This is still unsafe: The returned `GtkPadActionEntry` borrows the pointers.

I'd simply remove this `ToGlib` impl and implement the one conversion that we actually need manually so that it's actually safe.

src/pad_controller.rs
```diff
@@ -17,10 +17,11 @@ impl<O: IsA<PadController>> PadControllerExtManual for O {
     #[cfg(any(feature = "v3_22", feature = "dox"))]
     fn set_action_entries(&self, entries: &[PadActionEntry]) {
         let n_entries = entries.len() as i32;
-        let entries = entries.as_ptr();
+        let entries: Vec<gtk_sys::GtkPadActionEntry> =
+            entries.iter().map(|e| e.to_glib()).collect();
```
sdroege: I.e.
```rust
let entry_strings = entries.iter().map(|e| (CString::new(&e.label), CString::new(&e.action_name))).collect::<Vec<_>>();
let entries = entries.iter().zip(entry_strings.iter()).map(|(e, (label, action_name))| gtk_sys::GtkPadActionEntry {
            type_: e.type_.to_glib(),
            index: e.index,
            mode: e.mode,
            label: label.as_ptr() as *mut _,
            action_name: action_name.as_ptr() as *mut _,
}).collection::<Vec<_>>();
```

sdroege: You'll need to mark the fields of the struct as `pub(crate)` for this

GuillaumeGomez:
cc @EPashkin as well. :)

src/pad_action_entry.rs
```diff
@@ -4,42 +4,48 @@
...
     }
 
     pub fn get_index(&self) -> i32 {
-        self.0.index
+        self.index
     }
 
     pub fn get_mode(&self) -> i32 {
-        self.0.mode
+        self.mode
```
GuillaumeGomez: Shouldn't this be wrapped inside an enum or something to prevent users to change this value?

sdroege: How do you mean? The user sets this very value

sdroege:
IMHO we can also just remove this one function and the struct. It's only "convenience" (how?!) around `gtk_pad_controller_set_action()`. I don't see how it makes anything easier :)

src/pad_action_entry.rs
```diff
@@ -4,50 +4,53 @@
...
     }
 
     pub fn get_label(&self) -> Option<String> {
```
EPashkin: IMHO safe return `Option<&str>`

EPashkin:
Also think that `gtk_pad_controller_set_action` is enough in most cases,
but if `set_action_entries` done and safe IMHO better leave it.

src/pad_action_entry.rs
sdroege: Indeed

sdroege:
> Also think that `gtk_pad_controller_set_action` is enough in most cases,
> but if `set_action_entries` done and safe IMHO better leave it.

Fine with me either way :)

src/pad_action_entry.rs
```diff
@@ -4,50 +4,53 @@
...
-}
-
-#[doc(hidden)]
-impl ToGlib for PadActionEntry {
-    type GlibType = gtk_sys::GtkPadActionEntry;
 
-    fn to_glib(&self) -> gtk_sys::GtkPadActionEntry {
-        self.0
+    pub fn get_action_name(&self) -> Option<&str> {
+        Some(&self.action_name)
```
sdroege: Even more so, these two string getters don't really need to return an `Option`. They can't be `None` ever

src/pad_action_entry.rs
```diff
@@ -4,50 +4,53 @@
...
 
-#[repr(C)]
-pub struct PadActionEntry(gtk_sys::GtkPadActionEntry);
+#[derive(Debug, Clone)]
+pub struct PadActionEntry {
+    pub(crate) type_: PadActionType,
+    pub(crate) index: i32,
+    pub(crate) mode: i32,
+    pub(crate) label: String,
+    pub(crate) action_name: String,
```
sdroege: Actually... as we have getters for all these, they can as well stay completely private and you can use the getters in the other file :)

sdroege:
Thanks, looks good to me once that is fixed. @GuillaumeGomez please merge then :)

GuillaumeGomez:
Yes sir!

Thanks @sfanxiang !

sdroege:
@GuillaumeGomez I said "once that is fixed" :P

@sfanxiang Can you create a new PR for the two open discussions above?

PR #832, opened 2019-06-23
Fix docs for manual functions [ci skip]

EPashkin:
Part of https://github.com/gtk-rs/gir/issues/797

cc @GuillaumeGomez


GuillaumeGomez:
Thanks!

We'd need a tool to check if manual traits are forgotten because it'll happen again for sure...

PR #841, opened 2019-07-04
Make PadController::set_action_entries() public so it can actually be…

sdroege:
… used

CC @sfanxiang 

@GuillaumeGomez This one also needs a small bugfix release :)


GuillaumeGomez:
Urg.

sdroege:
@GuillaumeGomez all green

PR #849, opened 2019-07-12
Regen

GuillaumeGomez:
Part of gtk-rs/gir#806.

cc @EPashkin @sdroege


EPashkin:
@GuillaumeGomez Thanks, 👍

PR #852, opened 2019-07-18
Implement Builder::connect_signals_full

### LINKED ISSUE #707: Implement BuilderExt::add_callback_symbols() and BuilderExt::connect_signals()
These functions are needed to support using the callback function specified in the Handler field of a signal in Glade.

They are described here: https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-add-callback-symbols
and here: https://developer.gnome.org/gtk3/stable/GtkBuilder.html#gtk-builder-connect-signals

And they are listed in gtk_sys here: https://github.com/gtk-rs/sys/blob/master/gtk-sys/src/lib.rs#L14874
and here: https://github.com/gtk-rs/sys/blob/master/gtk-sys/src/lib.rs#L14881

Is there something blocking the support of these two functions?

sfanxiang:
This is a safe way to implement builder signal connection, as discussed in #707.

Closes #707.

cc @sdroege

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -641,6 +641,10 @@ manual_traits = ["BuilderExtManual"]
     ignore = true
     doc_trait_name = "BuilderExtManual"
     [[object.function]]
+    name = "connect_signals_full"
+    #manual: connect object and swapped signal not supported
+    ignore = true
+    [[object.function]]
     name = "get_object"
     #downcast
     ignore = true
--- a/src/builder.rs
+++ b/src/builder.rs
@@ -4,7 +4,9 @@
 
 use glib::object::{Cast, IsA};
 use glib::translate::*;
+use glib::GString;
 use glib::Object;
+use glib::ObjectExt;
 use gtk_sys;
 use std::path::Path;
 use Builder;
@@ -24,6 +26,12 @@ impl Builder {
 pub trait BuilderExtManual: 'static {
     fn get_object<T: IsA<Object>>(&self, name: &str) -> Option<T>;
     fn add_from_file<T: AsRef<Path>>(&self, file_path: T) -> Result<(), Error>;
+    fn connect_signals<
+        P: FnMut(&Builder, &str) -> Box<dyn Fn(&[glib::Value]) -> Option<glib::Value> + 'static>,
+    >(
+        &self,
+        func: P,
+    );
 }
 
 impl<O: IsA<Builder>> BuilderExtManual for O {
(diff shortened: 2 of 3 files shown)


src/builder.rs
```diff
@@ -52,4 +55,27 @@ impl<O: IsA<Builder>> BuilderExtManual for O {
...
+
+            let builder = from_glib_borrow(builder);
+            let object: glib::Object = from_glib_borrow(object);
+            let signal_name: GString = from_glib_borrow(signal_name);
+            let handler_name: GString = from_glib_borrow(handler_name);
+            let callback: *mut P = user_data as *const _ as usize as *mut P;
+            let func = (*callback)(&builder, handler_name.as_str());
+            object
+                .connect(signal_name.as_str(), flags & gobject_sys::G_CONNECT_AFTER != 0, move |args| func(args))
+                .expect("Failed to connect to builder signal");
```
GuillaumeGomez: Instead of panicking, shouldn't we return a `Result<(), String>`? I really don't like panics.

Gir.toml
```diff
@@ -641,6 +641,10 @@ manual_traits = ["BuilderExtManual"]
     ignore = true
     doc_trait_name = "BuilderExtManual"
     [[object.function]]
+    name = "connect_signals_full"
+    #manual
```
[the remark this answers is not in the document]
sfanxiang: "connect object and swapped signal not supported"?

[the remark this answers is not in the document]
GuillaumeGomez: Sounds good!

src/builder.rs
sdroege: To where would you return this `Result`? There's nobody there who can handle errors

src/builder.rs
```diff
@@ -52,4 +55,27 @@ impl<O: IsA<Builder>> BuilderExtManual for O {
...
+            assert!(flags & gobject_sys::G_CONNECT_SWAPPED == 0, "Swapped signal handler is not supported");
+
+            let builder = from_glib_borrow(builder);
+            let object: glib::Object = from_glib_borrow(object);
+            let signal_name: GString = from_glib_borrow(signal_name);
+            let handler_name: GString = from_glib_borrow(handler_name);
+            let callback: *mut P = user_data as *const _ as usize as *mut P;
+            let func = (*callback)(&builder, handler_name.as_str());
+            object
+                .connect(signal_name.as_str(), flags & gobject_sys::G_CONNECT_AFTER != 0, move |args| func(args))
```
sdroege: `flags.contains(...)` might be easier to read

src/builder.rs
GuillaumeGomez: Isn't this function called by users directly?

src/builder.rs
```diff
@@ -52,4 +55,27 @@ impl<O: IsA<Builder>> BuilderExtManual for O {
             }
         }
     }
+
+    fn connect_signals_full<P: FnMut(&Builder, &str) -> Box<dyn Fn(&[glib::Value]) -> Option<glib::Value> + Send + Sync + 'static>>(&self, func: P) {
```
sdroege: I already see people complaining about the trait bounds on the closure :) `Send` and `Sync` are not needed here as this is all for GTK objects. Remove those bounds, and if necessary use some `unsafe` code to cast the bounds in there again for `object.connect()` :)

src/builder.rs
sdroege: It's a callback that is called from GTK

GuillaumeGomez: Oh damn you're absolutely right! I didn't pay attention that we were in the callback... Nevermind then!

src/builder.rs
sfanxiang: @sdroege Actually, flags is `u32` here, and I don't think we should introduce `ConnectFlags` which is not used in public API.

sdroege: Indeed, sorry.

sdroege:
Looks all good to me

EPashkin:
@sfanxiang Thanks, please, update generated files

src/builder.rs
```diff
@@ -24,6 +26,12 @@ impl Builder {
 pub trait BuilderExtManual: 'static {
     fn get_object<T: IsA<Object>>(&self, name: &str) -> Option<T>;
     fn add_from_file<T: AsRef<Path>>(&self, file_path: T) -> Result<(), Error>;
+    fn connect_signals_full<
```
sdroege: I wouldn't call it `connect_signals_full` but only `connect_signals`. We can never have the other variant :)

sdroege:
Good to go for me otherwise. I assume you tested that it actually works and can be used? :) Maybe worth adding an example to the examples repo?

sfanxiang:
@sdroege 
> I assume you tested that it actually works and can be used? :) Maybe worth adding an example to the examples repo?

[This example](https://github.com/sfanxiang/gtk-rs-examples4/blob/6ee2e6b3087c01cdc3a766feafb3d5520cd8ca01/src/bin/builder_signal.rs#L25) builds and works. Does it suffice?

sdroege:
On 20 July 2019 10:35:07 EEST, Xiang Fan <notifications@github.com> wrote:
>@sdroege 
>> I assume you tested that it actually works and can be used? :) Maybe
>worth adding an example to the examples repo?
>
>[This
>example](https://github.com/sfanxiang/gtk-rs-examples4/blob/6ee2e6b3087c01cdc3a766feafb3d5520cd8ca01/src/bin/builder_signal.rs#L25)
>builds and works. Does it suffice?

I guess so :) can you send a PR for that against the gtk3 examples too? Shouldn't need any code changes

In any case, this PR is good to go for me.


Unrelated, did you merge back any changes that happened in the gtk3 versions of the crates into the gtk4 versions since you started? Should do that before the first release in any case.


-- 

sdroege:
@GuillaumeGomez let's get this in?

PR #861, opened 2019-08-05
subclass: Get started on subclassing GtkWidget

BrainBlasted:
Adds support for subclassing GtkWidget and overriding
6 of it's many vfuncs.

### DIFF
--- a/src/subclass/widget.rs
+++ b/src/subclass/widget.rs
@@ -157,108 +157,6 @@ impl <T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
 
 unsafe impl<T: ObjectSubclass + WidgetImpl> IsSubclassable<T> for WidgetClass {
     fn override_vfuncs(&mut self) {
-        unsafe extern "C" fn widget_adjust_baseline_allocation<T: ObjectSubclass>(
-            ptr: *mut gtk_sys::GtkWidget,
-            baseptr: *mut i32,
-        ) where
-            T: WidgetImpl,
-        {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Widget = from_glib_borrow(ptr);
-
-            imp.adjust_baseline_allocation(&wrap, &mut *baseptr)
-        }
-
-        unsafe extern "C" fn widget_adjust_baseline_request<T: ObjectSubclass>(
-            ptr: *mut gtk_sys::GtkWidget,
-            minptr: *mut i32,
-            natptr: *mut i32,
-        ) where
-            T: WidgetImpl,
-        {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Widget = from_glib_borrow(ptr);
-
-            imp.adjust_baseline_request(&wrap, &mut *minptr, &mut *natptr)
-        }
-
-        unsafe extern "C" fn widget_adjust_size_allocation<T: ObjectSubclass>(
-            ptr: *mut gtk_sys::GtkWidget,
-            orientation: gtk_sys::GtkOrientation,
-            minptr: *mut i32,
-            natptr: *mut i32,
-            posptr: *mut i32,
-            sizeptr: *mut i32,
-        ) where
-            T: WidgetImpl,
-        {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Widget = from_glib_borrow(ptr);
-            let wrap_orientation: Orientation = from_glib(orientation);
-
-            imp.adjust_size_allocation(&wrap, wrap_orientation, &mut *minptr, &mut *natptr, &mut *posptr, &mut *sizeptr)
-        }
-
-        unsafe extern "C" fn widget_adjust_size_request<T: ObjectSubclass>(
-            ptr: *mut gtk_sys::GtkWidget,
-            orientation: gtk_sys::GtkOrientation,
-            minptr: *mut i32,
-            natptr: *mut i32,
-        ) where
-            T: WidgetImpl,
-        {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Widget = from_glib_borrow(ptr);
-            let wrap_orientation: Orientation = from_glib(orientation);
-
-            imp.adjust_size_request(&wrap, wrap_orientation, &mut *minptr, &mut *natptr)
-        }
-
-        unsafe extern "C" fn widget_button_press_event<T: ObjectSubclass>(
-            ptr: *mut gtk_sys::GtkWidget,
-            btnptr: *mut gdk_sys::GdkEventButton,
-        ) -> glib_sys::gboolean
-            where T: WidgetImpl,
-        {
-            let instance = &*(ptr as *mut T::Instance);
-            let imp = instance.get_impl();
-            let wrap: Widget = from_glib_borrow(ptr);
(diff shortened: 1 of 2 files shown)


src/subclass/widget.rs
```diff
@@ -0,0 +1,249 @@
...
+
+    fn adjust_size_allocation(&self, widget: &Widget, orientation: Orientation, minimum_size: &mut i32, natural_size: &mut i32, allocated_pos: &mut i32, allocated_size: &mut i32) {
+        self.parent_adjust_size_allocation(widget, orientation, minimum_size, natural_size, allocated_pos, allocated_size)
+    }
+
+    fn adjust_size_request(&self, widget: &Widget, orientation: Orientation, minimum_size: &mut i32, natural_size: &mut i32) {
+        self.parent_adjust_size_request(widget, orientation, minimum_size, natural_size)
+    }
+
+    fn button_press_event(&self, widget: &Widget, event: &mut gdk::EventButton) -> bool {
```
sdroege: Should this be mutable? Is there anything meaningful one can modify in the event and is that allowed?

BrainBlasted: I decided to make it mutable in case anyone wants to use any `&mut self` functions of `gdk::EventButton`.

src/subclass/widget.rs
```diff
@@ -0,0 +1,249 @@
...
+        self.parent_button_press_event(widget, event)
+    }
+
+    fn button_release_event(&self, widget: &Widget, event: &mut gdk::EventButton) -> bool {
```
sdroege: The return value here and above should probably be `gtk::Inhibit`

BrainBlasted: Same for `button_press_event`?

src/subclass/widget.rs
```diff
@@ -0,0 +1,249 @@
...
+                .button_release_event
+                .expect("No parent class impl for \"button_release_event\"");
+            f(widget.to_glib_none().0, event.to_glib_none_mut().0) != 0
+        }
+    }
+}
+
+unsafe impl<T: ObjectSubclass + WidgetImpl> IsSubclassable<T> for WidgetClass {
+    fn override_vfuncs(&mut self) {
+        unsafe extern "C" fn widget_adjust_baseline_allocation<T: ObjectSubclass>(
```
sdroege: I would move these trampolines outside the `override_vfuncs` function. Otherwise you'll end up with 100s lines of trampoline functions declared inside here :) It's going to be hard to read, just move them all to the bottom of the file.

src/subclass/widget.rs
```diff
@@ -0,0 +1,249 @@
...
+            posptr: *mut i32,
+            sizeptr: *mut i32,
+        ) where
+            T: WidgetImpl,
+        {
+            let instance = &*(ptr as *mut T::Instance);
+            let imp = instance.get_impl();
+            let wrap: Widget = from_glib_borrow(ptr);
+
+            let wrap_orientation: Orientation = match orientation {
```
sdroege: Why not `from_glib(orientation)`?

BrainBlasted: `gtk::Orientation` doesn't implement any of the `FromGlib*` traits

src/subclass/widget.rs
```diff
@@ -0,0 +1,249 @@
...
+            btnptr: *mut gdk_sys::GdkEventButton,
+        ) -> glib_sys::gboolean
+            where T: WidgetImpl,
+        {
+            let instance = &*(ptr as *mut T::Instance);
+            let imp = instance.get_impl();
+            let wrap: Widget = from_glib_borrow(ptr);
+            let mut evwrap: gdk::EventButton = from_glib_borrow(btnptr);
+
+            imp.button_press_event(&wrap, &mut evwrap) as glib_sys::gboolean
```
sdroege: This is unsafe: you can't pass a mutable reference to something retrieved via `from_glib_borrow()` as that allows to produce double frees

sdroege: I.e. you must only pass immutable references if you use `from_glib_borrow()`

BrainBlasted: Is there a way that could allow `EventButton` to be mutable? Or will it need to be immutable?

src/subclass/widget.rs
sdroege: Yes

src/subclass/widget.rs
sdroege: The question is if that's valid / allowed by the GTK API. My gut feeling would say no :)

BrainBlasted: 👍🏾

src/subclass/widget.rs
sdroege: It does: https://github.com/gtk-rs/gtk/blob/4e3d79cecc70013f5be1b9ff419b60daaf81ec55/src/auto/enums.rs#L3333-L3356

BrainBlasted: Ah, I missed that when reviewing the docs. Thanks.

src/subclass/widget.rs
sdroege: Hmm, I don't see any clean way right now

sdroege:
Good work :)

src/subclass/widget.rs
sdroege: The event stuff will all be nicer with GTK 4 anyway :)

BrainBlasted:
Pushed support for one more vfunc, will get started on the review

src/subclass/widget.rs
```diff
@@ -34,6 +34,10 @@ pub trait WidgetImpl:
     fn button_release_event(&self, widget: &Widget, event: &mut gdk::EventButton) -> bool {
         self.parent_button_release_event(widget, event)
     }
+
+    fn can_activate_accel(&self, widget: &Widget, signal_id: u32) -> bool {
```
sdroege: That's a curious virtual method. To make proper use of that we need some more bindings for the signal API. To be able to work with signal ids (hint: this shouldn't be a plain `u32`), introspect signals, etc.

BrainBlasted: Should I comment it out for now?

sdroege: Yeah, it needs some more work to be actually usable :)

src/subclass/widget.rs
BrainBlasted: According to `rustc` I think it needs to be mutable, since the vfunc doesn't take an owned `gdk::EventButton` but a mutable pointer:

```
error[E0596]: cannot borrow `*event` as mutable, as it is behind a `&` reference
   --> src/subclass/widget.rs:139:48
    |
131 |     fn parent_button_release_event(&self, widget: &Widget, event: &gdk::EventButton) -> Inhibit {
    |                                                                   ----------------- help: consider changing this to be a mutable reference: `&mut gdk::EventButton`
...
139 |             Inhibit(f(widget.to_glib_none().0, event.to_glib_none_mut().0) != 0)
    |                                                ^^^^^ `event` is a `&` reference, so the data it refers to cannot be borrowed as mutable

```

sdroege: `glib::translate::mut_override(event.to_glib_none().0)` or a simple cast :)

sdroege:
Looks good to me, good to be merged and then we only need an example :)

EPashkin:
@BrainBlasted Thanks

sdroege:
@BrainBlasted Do you want to add `draw` or should we get this in for now?

BrainBlasted:
Can we get this in for now?

GuillaumeGomez:
Sure. Thanks! (CI failures are spurious)

Issue #862, opened 2019-08-06
GtkWidget Subclassing: virtual methods


BrainBlasted:
This issue is to provide a list of vfuncs of GtkWidget, to be crossed of as they are implemented.
The first batch is covered in #861.

* [x] adjust_baseline_allocation
* [x] adjust_baseline_request
* [x] adjust_size_allocation 
* [x] adjust_size_request
* [x] button_press_event
* [x] button_release_event
* [ ] ~~can_activate_accel~~ (Needs work on signals API)
* [ ] child_notify
* [ ] composited_changed
* [ ] compute_expand 
* [ ] configure_event 
* [ ] damage_event
* [ ] delete_event
* [ ] destroy
* [ ] destroy_event 
* [ ] direction_changed
* [ ] dispatch_child_properties_changed
* [ ] drag_begin
* [ ] drag_data_delete 
* [ ] drag_data_get
* [ ] drag_data_received 
* [ ] drag_drop 
* [ ] drag_end
* [ ] drag_failed
* [ ] drag_leave
* [ ] drag_motion
* [ ] draw
* [ ] enter_notify_event
* [ ] event
* [ ] focus
* [ ] focus_in_event
* [ ] focus_out_event 
* [ ] get_accessible 
* [ ] get_preferred_height
* [ ] get_preferred_height_and_baseline_for_width 
* [ ] get_preferred_height_for_width
* [ ] get_preferred_width 
* [ ] get_preferred_width_for_height 
* [ ] get_request_mode 
* [ ] grab_broken_event
* [ ] grab_focus
* [ ] grab_notify 
* [ ] hide 
* [ ] hierarchy_changed
* [ ] key_press_event 
* [ ] key_release_event
* [ ] keynav_failed 
* [ ] leave_notify_event
* [ ] map
* [ ] map_event
* [ ] mnemonic_activate
* [ ] motion_notify_event
* [ ] move_focus 
* [ ] parent_set
* [ ] popup_menu
* [ ] property_notify_event
* [ ] proximity_in_event
* [ ] proximity_out_event 
* [ ] query_tooltip 
* [ ] queue_draw_region 
* [ ] realize 
* [ ] screen_changed
* [ ] scroll_event
* [ ] selection_clear_event
* [ ] selection_get
* [ ] selection_notify_event
* [ ] selection_received
* [ ] selection_request_event
* [ ] show
* [ ] show_all 
* [ ] show_help
* [ ] size_allocate
* [ ] state_changed 
[...]

PR #864, opened 2019-08-10
subclass/widget: Add vfuncs between child_notify and draw

BrainBlasted:
Part of https://github.com/gtk-rs/gtk/issues/862

### DIFF
--- a/src/subclass/widget.rs
+++ b/src/subclass/widget.rs
@@ -341,11 +341,15 @@ impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
             let f = (*parent_class)
                 .compute_expand
                 .expect("No parent class impl for \"compute_expand\"");
+            let mut h: i32 = hexpand_p.to_glib();
+            let mut v: i32 = vexpand_p.to_glib();
             f(
                 widget.to_glib_none().0,
-                *hexpand_p as i32 as *mut i32,
-                *vexpand_p as i32 as *mut i32,
-            )
+                &mut h,
+                &mut v,
+            );
+            *hexpand_p = from_glib(h);
+            *vexpand_p = from_glib(v);
         }
     }
 
@@ -763,10 +767,12 @@ where
     let instance = &*(ptr as *mut T::Instance);
     let imp = instance.get_impl();
     let wrap: Widget = from_glib_borrow(ptr);
-    let mut hexpand_p: bool = *hexpand_ptr != 0;
-    let mut vexpand_p: bool = *vexpand_ptr != 0;
+    let mut hexpand_p: bool = from_glib(*hexpand_ptr);
+    let mut vexpand_p: bool = from_glib(*vexpand_ptr);
 
-    imp.compute_expand(&wrap, &mut hexpand_p, &mut vexpand_p)
+    imp.compute_expand(&wrap, &mut hexpand_p, &mut vexpand_p);
+    *hexpand_ptr = hexpand_p.to_glib();
+    *vexpand_ptr = vexpand_p.to_glib();
 }
 
 unsafe extern "C" fn widget_configure_event<T: ObjectSubclass>(
--- a/src/subclass/widget.rs
+++ b/src/subclass/widget.rs
@@ -86,7 +86,7 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl + 'static {
     }
 
     fn delete_event(&self, widget: &Widget, event: &gdk::Event) -> Inhibit {
-         self.parent_delete_event(widget, event)
+        self.parent_delete_event(widget, event)
     }
 
     fn destroy(&self, widget: &Widget) {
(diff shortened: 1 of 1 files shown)


sdroege:
Can't review it properly now, will do tomorrow evening.

But gdk::EventAny is basically gdk::Event. And glib::ParamSpecs has bindings, but you'll want to hide the n_pspecs integer. That's the size of the Vec already. You'd use from_glib_none_num() here.

So the two commented out ones should be doable.
-- 

BrainBlasted:
How do I translate a `&Vec<glib::ParamSpec>` to `*mut *mut GParamSpec`?

sdroege:
> How do I translate a `&Vec<&glib::ParamSpec>` to `*mut *mut GParamSpec`?

`to_glib_none()` doesn't do that? Alternatively: `vec.iter().map(|p| p.to_glib_none().0).collect::<Vec<_>>()` and then `as_ptr()` on the result.

src/subclass/widget.rs
```diff
@@ -353,6 +716,309 @@ where
...
+    vexpand_ptr: *mut glib_sys::gboolean,
+)
+where
+    T: WidgetImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Widget = from_glib_borrow(ptr);
+    let mut hexpand_p: bool = *hexpand_ptr != 0;
+    let mut vexpand_p: bool = *vexpand_ptr != 0;
```
sdroege: Why not `from_glib()` here?

src/subclass/widget.rs
```diff
@@ -214,6 +311,253 @@ impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
...
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkWidgetClass;
+            let f = (*parent_class)
+                .compute_expand
+                .expect("No parent class impl for \"compute_expand\"");
+            f(
+                widget.to_glib_none().0,
+                *hexpand_p as i32 as *mut i32,
+                *vexpand_p as i32 as *mut i32,
```
sdroege: I think this is wrong. You're casting an integer to a pointer.

I'd do something like
```rust
let mut h: i32 = hexpand_p.to_glib();
f(&mut h);
*hexpand_p = from_glib(h);
```

src/subclass/widget.rs
```diff
@@ -353,6 +716,309 @@ where
...
+where
+    T: WidgetImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Widget = from_glib_borrow(ptr);
+    let mut hexpand_p: bool = *hexpand_ptr != 0;
+    let mut vexpand_p: bool = *vexpand_ptr != 0;
+
+    imp.compute_expand(&wrap, &mut hexpand_p, &mut vexpand_p)
```
sdroege: Afterwards you need to set the new values of `hexpand_p` / `vexpand_p` to the pointers again

```rust
*hexpand_ptr = hexpand_p.to_glib();
```

sdroege:
Looks good to me otherwise

BrainBlasted:
> > How do I translate a `&Vec<&glib::ParamSpec>` to `*mut *mut GParamSpec`?
> 
> `to_glib_none()` doesn't do that? Alternatively: `vec.iter().map(|p| p.to_glib_none().0).collect::<Vec<_>>()` and then `as_ptr()` on the result.

What's the right way to translate this back?

sdroege:
> What's the right way to translate this back?

You don't have to translate it back. It's an immutable array so it can't be changed anywhere along the call chain.

BrainBlasted:
I need to translate it in order to call from the `widget_dispatch_child_properties_changed` trampoline

sdroege:
An `install_properties()` from `glib/src/subclass/object.rs` for the other way around

BrainBlasted:
So ParamSpec should be owned instead of references?

src/subclass/widget.rs
BrainBlasted: `error[E0277]: the trait bound `bool: glib::translate::FromGlib<*mut i32>` is not satisfied`

BrainBlasted:
Cleaned up everything according to suggestions.

BrainBlasted:
Rebased and formatted.

src/subclass/widget.rs
sdroege: `from_glib(*hexpand_ptr)`, you need to dereference

sdroege:
> So ParamSpec should be owned instead of references?

I'd say that it actually should be `&[ParamSpec]` instead of `&Vec<ParamSpec>` as that's more idiomatic Rust.

src/subclass/widget.rs
BrainBlasted: Yup, figured that out and pushed it.

sdroege:
Yeah looks all good apart from that, thanks :)

BrainBlasted:
Updated to use slices.

sdroege:
@GuillaumeGomez Good to go

GuillaumeGomez:
Thanks!

PR #865, opened 2019-08-10
subclass: Add subclassing for GtkContainer

BrainBlasted:
Currently ommitted vfuncs:

* composite_name
* set_child_property
* get_child_property

### DIFF
--- a/src/subclass/container.rs
+++ b/src/subclass/container.rs
@@ -1,0 +1,209 @@
+use gtk_sys;
+
+use glib::translate::*;
+
+use glib::subclass::prelude::*;
+
+use super::widget::WidgetImpl;
+use Container;
+use ContainerClass;
+use Widget;
+use WidgetPath;
+
+pub trait ContainerImpl: ContainerImplExt + WidgetImpl + 'static {
+    fn add(&self, container: &Container, widget: &Widget) {
+        self.parent_add(container, widget)
+    }
+
+    fn remove(&self, container: &Container, widget: &Widget) {
+        self.parent_remove(container, widget)
+    }
+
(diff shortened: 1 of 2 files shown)


sdroege:
Thanks, generally looks good :) will do a proper review tomorrow evening too here
-- 

sdroege:
> composite_name

What's complicated with this one? But it's deprecated since 3.10, so maybe we shouldn't add it at all anyway :)

> * set_child_property
> 
> * set_child_property
> 
> * get_child_property

These are very interesting. I don't know how to properly bind or use them because of the integer property id. Do you happen to know some implementations (in C) of those two virtual methods?

BrainBlasted:
> What's complicated with this one? But it's deprecated since 3.10, so maybe we shouldn't add it at all anyway :)

Nothing particularly complicated - I was just too tired to work on it at the time :)

I can look for implementations.

sdroege:
> Nothing particularly complicated - I was just too tired to work on it at the time :)

That's fine, let's just ignore that one then as it's deprecated anyway :)

For the others, if you find implementations I can probably figure out how to bind them.

sdroege:
Also of course fine to just omit the other two for now too, and simply create an issue to track that. They don't seem super useful.

sdroege:
Looks good to me otherwise, thanks!

BrainBlasted:
Made the issue: https://github.com/gtk-rs/gtk/issues/867

sdroege:
@GuillaumeGomez good to go for me then :)

GuillaumeGomez:
Thanks @BrainBlasted !

PR #875, opened 2019-08-14
Menu: set_property_widget() should take IsA<Widget> instead of Widget

### LINKED ISSUE #846: MenuExt::set_property_attach_widget() should take IsA<Widget> instead of Widget

Cogitri:
fixes #846

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -1334,6 +1334,9 @@ manual_traits = ["GtkMenuExtManual"]
     [[object.function]]
     name = "place_on_monitor"
     version = "3.22"
+    [[object.function]]
+    name = "set_property_attach_widget"
+    ignore = true
 
 [[object]]
 name = "Gtk.MenuItem"
--- a/src/auto/menu.rs
+++ b/src/auto/menu.rs
@@ -663,8 +663,6 @@ pub trait GtkMenuExt: 'static {
     #[cfg(any(feature = "v3_22", feature = "dox"))]
     fn set_property_anchor_hints(&self, anchor_hints: gdk::AnchorHints);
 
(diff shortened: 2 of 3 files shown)


Gir.toml
```diff
@@ -1334,6 +1334,9 @@ manual_traits = ["GtkMenuExtManual"]
     [[object.function]]
     name = "place_on_monitor"
     version = "3.22"
+    [[object.function]]
+    name = "set_property_attach_widget"
+    ignore = true
```
Cogitri: I can't tell `gir` to take a different param, can I? After reading the readme there didn't seem to be a way to do it, but maybe I'm just missing something?

Cogitri: Also, did I typo here? gir still generates the function

sdroege:
I believe this is a bug in gir. It should autogenerate this as `IsA<Widget>` already, like it would do for normal functions. @EPashkin what do you think?

Gir.toml
EPashkin: This is property

EPashkin: so something like this:
```
    [[object.property]]
    #need be IsA<Widget>
    name = "attach-widget"
    generate = ["get", "notify"]
```

sdroege: It should however actually generate this correctly automatically, like it would already for normal functions :)

EPashkin:
@sdroege You right, I added issue

sdroege:
Should we merge this now? @EPashkin @GuillaumeGomez ?

Or wait until it's fixed in gir so that we cover all the possible cases at once?

GuillaumeGomez:
I prefer to wait for gir fix.

Cogitri:
Fine by me, will close this in favour of the GIR issue then

Issue #874, opened 2019-08-13
gtk_style_context_get_property not generated


BrainBlasted:
This might not be generated because it's similar to g_object_get_property? Despite the naming it does not have the same functionality. Rather, it gets the **style property** of the Widget. Without it widgets like [Granite's avatar widget](https://github.com/elementary/granite/blob/6d0dac21b728b1ed28baaf38a004dca378559435/lib/Widgets/Avatar.vala) can't be implemented in Rust.

sdroege:
Unclear why they're not generated, they're explicitly ignored here https://github.com/gtk-rs/gtk/blob/master/Gir.toml#L1768-L1770. Since many years ago in https://github.com/gtk-rs/gtk/commit/4e6d5db1991bdd923069d802d3250ea0ea15f655

Removing that should make them generated correctly, there's nothing complicated about those functions.

sdroege:
Do you want to provide a PR for that? Simply removing those lines from the `Gir.toml` and then regenerating

BrainBlasted:
Sure, I'll get to that tomorrow morning.

On August 13, 2019 10:51:20 PM PDT, "Sebastian Dröge" <notifications@github.com> wrote:
>Do you want to provide a PR for that? Simply removing those lines from
>the `Gir.toml` and then regenerating
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/gtk-rs/gtk/issues/874#issuecomment-521110924

-- 
Sent from my p≡p for Android.

ArekPiekarz:
I am also interested in this functionality instead of having to resort to unsafe workarounds in #949 to get information about fonts used in widgets.

PR #876, opened 2019-08-14
Generate GtkStyleContext get property functions

BrainBlasted:
These were ignored previously.

Fixes https://github.com/gtk-rs/gtk/issues/874

### DIFF
--- a/src/auto/style_context.rs
+++ b/src/auto/style_context.rs
@@ -3,6 +3,7 @@
 // DO NOT EDIT
 
 use gdk;
+use glib;
 use glib::object::Cast;
 use glib::object::IsA;
 use glib::signal::connect_raw;
@@ -170,6 +171,8 @@ pub trait StyleContextExt: 'static {
 
     fn get_path(&self) -> Option<WidgetPath>;
 
+    fn get_property(&self, property: &str, state: StateFlags) -> glib::Value;
+
     fn get_scale(&self) -> i32;
 
     fn get_screen(&self) -> Option<gdk::Screen>;
@@ -179,6 +182,8 @@ pub trait StyleContextExt: 'static {
     fn get_state(&self) -> StateFlags;
 
     //fn get_style(&self, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs);
+
+    fn get_style_property(&self, property_name: &str, value: &mut glib::Value);
 
     //fn get_style_valist(&self, args: /*Unknown conversion*//*Unimplemented*/Unsupported);
 
@@ -365,6 +370,19 @@ impl<O: IsA<StyleContext>> StyleContextExt for O {
         }
     }
 
+    fn get_property(&self, property: &str, state: StateFlags) -> glib::Value {
+        unsafe {
+            let mut value = glib::Value::uninitialized();
+            gtk_sys::gtk_style_context_get_property(
+                self.as_ref().to_glib_none().0,
+                property.to_glib_none().0,
+                state.to_glib(),
+                value.to_glib_none_mut().0,
+            );
+            value
+        }
+    }
+
     fn get_scale(&self) -> i32 {
         unsafe { gtk_sys::gtk_style_context_get_scale(self.as_ref().to_glib_none().0) }
     }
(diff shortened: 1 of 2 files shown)


GuillaumeGomez:
:+1:

src/auto/style_context.rs
```diff
@@ -180,6 +183,8 @@ pub trait StyleContextExt: 'static {
 
     //fn get_style(&self, : /*Unknown conversion*//*Unimplemented*/Fundamental: VarArgs);
 
+    fn get_style_property(&self, property_name: &str, value: &mut glib::Value);
```
sdroege: Ah that's why it was ignored probably... this should *return* an `Option<glib::Value>`

src/auto/style_context.rs
```diff
@@ -170,6 +171,8 @@ pub trait StyleContextExt: 'static {
 
     fn get_path(&self) -> Option<WidgetPath>;
 
+    fn get_property(&self, property: &str, state: StateFlags) -> glib::Value;
```
sdroege: And also this should be `Option`

src/auto/style_context.rs
```diff
@@ -398,6 +416,16 @@ impl<O: IsA<StyleContext>> StyleContextExt for O {
     //    unsafe { TODO: call gtk_sys:gtk_style_context_get_style() }
     //}
 
+    fn get_style_property(&self, property_name: &str, value: &mut glib::Value) {
+        unsafe {
+            gtk_sys::gtk_style_context_get_style_property(
+                self.as_ref().to_glib_none().0,
+                property_name.to_glib_none().0,
+                value.to_glib_none_mut().0,
```
EPashkin: May be this function better be manual as 
`fn get_style_property(&self, property_name: &str) -> glib::Value`

src/auto/style_context.rs
sdroege: This can probably be configured in the toml by marking the return type as nullable

src/auto/style_context.rs
sdroege: Not sure why gir gets the other function right and this one not though...

sdroege: @EPashkin any guess? :)

src/auto/style_context.rs
EPashkin: not sure that it can return `Option`, same as `gtk_style_context_get_property` at minimum by docs

sdroege: What would it return if the property name is invalid though?

EPashkin: Docs don't say that in both cases :(

src/auto/style_context.rs
EPashkin: IMHO because parameter not marked as `out`
```
          <parameter name="value" transfer-ownership="none">
            <doc xml:space="preserve" filename="gtk/gtkstylecontext.c" line="1727">Return location for the property value</doc>
            <type name="GObject.Value" c:type="GValue*"/>
          </parameter>
```

src/auto/style_context.rs
sdroege: Code says `g_critical()`, so we would cause memory unsafety here. We need to check the existence of the property beforehand somehow... like in `Object::get_property()`

src/auto/style_context.rs
sdroege: That could be fixed with some addition to `gir-files/fix.sh` :)

src/auto/style_context.rs
BrainBlasted: This is a style property though and not a regular GObject property. Looking at the StyleContext API I don't see any way to check beforehand.

sdroege: We could have manual code that initializes the `GValue` with `mem::zeroed()` and afterwards checks if the type of it is still 0.

BrainBlasted: Looking into it, this function is deprecated (though not documented as such). Maybe we could ignore this one, and fix `get_property`?

sdroege: Sure, works for me. Since when is the other function deprecated?

BrainBlasted: `get_property` isn't deprecated, and I'm not sure when this one was deprecated. Someone mentioned it in IRC as deprecated, and sure enough it [isn't in the GTK4 docs](https://developer.gnome.org/gtk4/stable/GtkStyleContext.html#gtk-style-context-get-property)

sdroege: Maybe they meant that it's deprecated as in "does not exist in GTK4 anymore".

I think for the "deprecated" one the way forward is to panic if it fails (the `g_critical()` from C is basically a panic), or otherwise make it return a `Result`.

The other one, if we can check beforehand then let's do that or otherwise like above.

sdroege:
How should we move forward here? @BrainBlasted ?

PR #878, opened 2019-08-16
subclass: Add support for subclassing GtkHeaderBar

BrainBlasted:
Like EventBox this is an empty trait and `override_vfuncs`.


sdroege:
@GuillaumeGomez :+1:

GuillaumeGomez:
Thanks!

PR #881, opened 2019-08-20
Update README.md

andreevlex:


README.md
```diff
@@ -26,7 +26,7 @@ Avoid mixing versioned and git crates like this:
 ```toml
 # This will not compile
 [dependencies]
-gdk = "0.2"
+gtk = "0.2"
```
GuillaumeGomez: Seems weird to have gtk twice. What are you trying to do?

andreevlex: This example describes that the versions do not need to be mixed, but two different packages are specified. Do to was so as written above)

EPashkin: @andreevlex By "version" we meant package and git

sdroege: But that's exactly what @andreevlex's change is fixing here? :)

EPashkin: @sdroege IMHO including one crate from both sources is unsound in any project,
and old example show "avoid mixing versioned and git crates like this:"
IMHO "versioned" better changed to "packaged" or "published"

Issue #887, opened 2019-09-06
Remove subclassing feature


GuillaumeGomez:
It shouldn't be a feature. (I opened this issue here but it's the same for all repositories!)

PR #888, opened 2019-09-07
Remove subclass feature

GuillaumeGomez:
Part of gtk-rs/gtk#887

The second commit is for a better dox support.

cc @sdroege @EPashkin


EPashkin:
👍

sdroege:
:+1:

sdroege:
@GuillaumeGomez I think examples are missing and then these can all go in

Issue #892, opened 2019-09-13
gtk::BoxBuilder is missing orientation() 


HossainAdnan:
Hi, if I'm not mistaken, Gtk-rs seems to provide builder patterns for building widgets. Hence I can't help but notice that the `gtk::BoxBuilder` is missing something similar to `orientation(&self, orientation: gtk::Orientation) -> Self` method. Is this intentional?

```rust
let hbox = gtk::BoxBuilder::new().orientation(gtk::Orientation::Horizontal).spacing(5).build();
// should be similar to
let hbox = gtk::Box::new(gtk::Orientation::Horizontal, 5);
```

It's not an issue, it's just a question. 
Also, on a **very** unrelated note, is there any persistent platform (forum, not a chat platform) where one can discuss Gtk-rs related topics?

EPashkin:
@HossainAdnan Thanks for reporting, property "orientation" was in interface Orientable, seems builder generator not scan these.

GuillaumeGomez:
@HossainAdnan: https://gtk-rs.org/docs-src/contact

EPashkin:
All related PRs merged, so this can be closed

PR #894, opened 2019-09-14
Builder use implemented interfaces properties

EPashkin:
Part of https://github.com/gtk-rs/gtk/issues/892

cc @HossainAdnan, @GuillaumeGomez, @sdroege


EPashkin:
Updated with using child configs, also changed many "ignore=true" for properties.

src/auto/about_dialog.rs
```diff
@@ -127,7 +127,6 @@ pub struct AboutDialogBuilder {
     parent: Option<Container>,
     receives_default: Option<bool>,
     sensitive: Option<bool>,
-    //style: /*Unknown type*/,
```
EPashkin: Deprecated property

EPashkin:
Ready for final review

sdroege:
:+1:

EPashkin:
CI passed

EPashkin:
Ping

src/auto/about_dialog.rs
GuillaumeGomez: Shouldn't we handle it like we do with functions? Like adding the deprecated attribute starting from a given version and remove it from the version it got removed?

GuillaumeGomez:
I just have one question but otherwise looks good to me as well!

src/auto/about_dialog.rs
EPashkin: It was deprecated in 3.0: https://developer.gnome.org/gtk3/stable/GtkStyle.html#gtk-style-new
so it totally deprecated

GuillaumeGomez: Perfect then!

GuillaumeGomez:
Thanks!

PR #896, opened 2019-09-27
Fix invalid cargo key for docs.rs

GuillaumeGomez:
Part of gtk-rs/gir#845

cc @sdroege @EPashkin


GuillaumeGomez:
Weird cargo bug...

PR #899, opened 2019-10-04
Get rid of Uninitialized impl for TargetEntry

sdroege:
It used mem::zeroed() and that causes undefined behaviour here. String
and other types can't be simply zeroed here.

----

This also caused a compiler warning with 1.38.


GuillaumeGomez:
👍

sdroege:
@GuillaumeGomez All green now :)

GuillaumeGomez:
Thanks!

EPashkin:
@sdroege Thanks

PR #901, opened 2019-10-18
Warn on macos when initializing from non_main_thread

nipunn1313:
Saw that we could do this with pthread, but opted to check the threadname as it was easier to do.
https://users.rust-lang.org/t/guis-and-the-main-thread/2863/22

I also wasn't sure if it belonged in `fn init` or `fn set_initialized` - tossed it in init as initial approach.

### DIFF
--- a/src/rt.rs
+++ b/src/rt.rs
@@ -10,7 +10,10 @@ use libc::c_uint;
 use std::cell::Cell;
 use std::ptr;
 use std::sync::atomic::{AtomicBool, Ordering, ATOMIC_BOOL_INIT};
-use std::thread;
+
+extern "C" {
+    fn pthread_main_np() -> i32;
+}
 
 thread_local! {
     static IS_MAIN_THREAD: Cell<bool> = Cell::new(false)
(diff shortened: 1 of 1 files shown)


nipunn1313:
issue from http://gtk-rs.org/docs/gtk/#threads

tested on our own app - saw the panic on mac!

src/rt.rs
```diff
@@ -89,6 +90,9 @@ pub fn init() -> Result<(), glib::BoolError> {
         return Ok(());
     } else if is_initialized() {
         panic!("Attempted to initialize GTK from two different threads.");
+    } else if cfg!(target_os = "macos") && thread::current().name() != Some("main") {
```
sdroege: There's nothing preventing you from starting a new thread with the name "main" :)

Ideally you'd use the `pthread_main_np()` API here, that's more reliable. You don't have to depend on some external crate for that, just do

```rust
extern "C" {
    unsafe fn pthread_main_np() -> i32;
}
```

And then call that in here

nipunn1313: Sounds good! Thanks

src/rt.rs
```diff
@@ -11,6 +11,10 @@ use std::cell::Cell;
 use std::ptr;
 use std::sync::atomic::{AtomicBool, Ordering, ATOMIC_BOOL_INIT};
 
+extern "C" {
```
sdroege:
```suggestion
#[cfg(target_os = "macos")]
extern "C" {
```

nipunn1313: I don't think this will compile as is - because `cfg!` cannot determine at compile time. However, I might be able to use a compile time inline `#[cfg]`. If we need an if/else chain, we can use something like https://github.com/alexcrichton/cfg-if - but I think in this case we don't need that and it'll be simpler to put some curly braces around the block! Updating momentarily!

sdroege:
:+1: @GuillaumeGomez

GuillaumeGomez:
Thanks!

PR #910, opened 2019-11-11
Regen

GuillaumeGomez:
Part of gtk-rs/sys#152.


sdroege:
```
error[E0432]: unresolved import `Value`
  --> src/fixed.rs:10:5
   |
10 | use Value;
   |     ^^^^^ no `Value` in the root

```

sdroege:
@GuillaumeGomez should be good to go

sdroege:
:+1:

Issue #913, opened 2019-11-17
Better AccelGroup support?


upsuper:
I'm trying to add some shortcuts to a window without widgets, but found it's quite challenging.

I found that `AccelGroup` in GTK+ seems to be for this usecase, but it is not very intuitive how to use it with gtk-rs, and I'm suspecting that I would need some `unsafe` for using it at the moment. Specifically I feel that [`AccelGroupExt::connect`](https://gtk-rs.org/docs/gtk/trait.AccelGroupExt.html#tymethod.connect) may need a better wrapping method.

It would be greatly appreciated if you can make the binding to `AccelGroup` easier to use with gtk-rs.

sdroege:
Why do you think you'll need `unsafe` code for using it? The way it works currently is that you'd pass a `glib::Closure` in there, which you can create from fully safe code.

However I agree that this is far from convenient because of the `glib::Value` handling required :) Do you have any ideas how the API could look like to be more convenient to use?

upsuper:
Thanks for your reply. I'm glad to hear that this can be done with safe code. Is there any example about how to create a `Closure` from safe code?

sdroege:
```rust
let c = glib::Closure::new(|args| {
    let arg0 = args[0].get::<WhateverTypeYouExpectHere>().expect("wrong type");
    [...]

    // Return some value if the closure is required to return a value, otherwise `None`
    Some(true.to_value())
});
```

Note that this will panic if you get the types of the `glib::Value`s wrong.

upsuper:
It seems this is not so easy. I tried to add a shortcut which changes state of some widget, and the compiler complains about the closure is not `Send` because `GObject`'s pointer is not `Send`.

sdroege:
Use something like `fragile::Fragile` for this use-case.

We should probably add a `new_local()` variant to `glib::Closure` that simply panics if ever called from a different thread then it was created at. That way we could also allow non-`Send`/non-`Sync` closures here. I'll try to get that added before the next release.

upsuper:
> Do you have any ideas how the API could look like to be more convenient to use?

There are several things which can be done.

Firstly `connect` function should be renamed because it conflicts with `ObjectExt::connect` which means people would need to write qualified name (`AccelGroupExt::connect(&accel_group, ...)`) to use it.

The second thing would be that it should take a `Fn` similar to all the other `connect_*` methods (I think they all take `Fn(&Self) + 'static`?) that would make it much easier to use I suppose.

[...]

PR #915, opened 2019-11-19
Make AccelGroup::connect() and ::connect_by_path() more usable

sdroege:
By renaming it to connect_accel_group() for preventing conflicts with
Object::connect() and by taking a normal Rust closure with fixed types
as arguments instead of a generic glib::Closure.

Return the generated glib::Closure as it is required for usage with a
few of the other AccelGroup and AccelLabel functions.

Fixes https://github.com/gtk-rs/gtk/issues/913

----

CC @upsuper does this work better for you?

### DIFF
--- a/src/accel_group.rs
+++ b/src/accel_group.rs
@@ -1,0 +1,119 @@
+// Copyright 2019, The Gtk-rs Project Developers.
+// See the COPYRIGHT file at the top-level directory of this distribution.
+// Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
+
+use gdk;
+use glib;
+use glib::object::{Cast, IsA};
+use glib::translate::*;
+use glib::ToValue;
+use gtk_sys;
+use AccelFlags;
+use AccelGroup;
+
+pub trait AccelGroupExtManual: 'static {
+    fn connect_accel_group<F>(
+        &self,
+        accel_key: u32,
+        accel_mods: gdk::ModifierType,
+        accel_flags: AccelFlags,
+        func: F,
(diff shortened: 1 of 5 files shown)


src/accel_group.rs
```diff
@@ -0,0 +1,119 @@
...
+use glib;
+use glib::object::{Cast, IsA};
+use glib::translate::*;
+use glib::ToValue;
+use gtk_sys;
+use AccelFlags;
+use AccelGroup;
+
+pub trait AccelGroupExtManual: 'static {
+    fn connect_accel_group<F>(
```
EPashkin: maybe it need `#[must_use]`

sdroege: No, you don't *have* to use it. Only if you want to pass the `glib::Closure` to any of the other functions, like `disconnect()` :)

EPashkin: Thanks for explanation

EPashkin:
👍

sdroege:
@GuillaumeGomez CI is green :)

src/accel_group.rs
```diff
@@ -0,0 +1,119 @@
...
+        &self,
+        accel_key: u32,
```
GuillaumeGomez: I'm wondering: shouldn't there be an enum for this one somewhere?

sdroege: Not that I could find, it's using `u32` everywhere.

GuillaumeGomez: Weird that there doesn't seem to be an enum anywhere... That's quite ugly. Well, whatever...

GuillaumeGomez:
Thanks!

PR #918, opened 2019-11-20
Regenerate with latest gir

sdroege:


sdroege:
@GuillaumeGomez This is good to go, CI will fail until the examples are fixed. Please merge :)

PR #921, opened 2019-12-04
subclass/widget: Implement default handling for parent events

### LINKED ISSUE #916: Application subclasss needs to handle parent delete_event
> Unrelated. The problem is that `delete_event` is `NULL` here and we currently don't handle that in the subclass bindings. We would have to check for which virtual methods it is valid to have `NULL` there and then what the default behaviour should be (another thing for which we don't have g-i annotations...). That requires diving into the C code. I've done that for the things in gio and gstreamer, but nowhere else.

> In this specific case, instead of panicking at that line it should return `false` here if it's `NULL`.

_Originally posted by @sdroege in https://github.com/gtk-rs/examples/pull/256#issuecomment-554977469_

alatiera:
Close #916

### DIFF
--- a/src/subclass/container.rs
+++ b/src/subclass/container.rs
@@ -51,10 +51,9 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .add
-                .expect("No parent class impl for \"add\"");
-            f(container.to_glib_none().0, widget.to_glib_none().0)
+            if let Some(f) = (*parent_class).add {
+                f(container.to_glib_none().0, widget.to_glib_none().0)
+            }
         }
     }
 
@@ -62,10 +61,9 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .remove
-                .expect("No parent class impl for \"remove\"");
-            f(container.to_glib_none().0, widget.to_glib_none().0)
+            if let Some(f) = (*parent_class).remove {
+                f(container.to_glib_none().0, widget.to_glib_none().0)
+            }
         }
     }
 
@@ -73,10 +71,9 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .check_resize
-                .expect("No parent class impl for \"check_resize\"");
-            f(container.to_glib_none().0)
+            if let Some(f) = (*parent_class).check_resize {
+                f(container.to_glib_none().0)
+            }
         }
     }
 
@@ -84,10 +81,9 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .set_focus_child
-                .expect("No parent class impl for \"set_focus_child\"");
-            f(container.to_glib_none().0, widget.to_glib_none().0)
+            if let Some(f) = (*parent_class).set_focus_child {
+                f(container.to_glib_none().0, widget.to_glib_none().0)
+            }
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/subclass/widget.rs
```diff
@@ -350,23 +350,25 @@ impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkWidgetClass;
-            let f = (*parent_class)
-                .button_press_event
-                .expect("No parent class impl for \"button_press_event\"");
-            let ev_glib = glib::translate::mut_override(event.to_glib_none().0);
-            Inhibit(from_glib(f(widget.to_glib_none().0, ev_glib)))
+            if let Some(f) = (*parent_class).button_press_event {
```
EPashkin: Can this part be macro?
`call_parent!(button_press_event, f(widget.to_glib_none().0, ev_glib))`

sdroege: No because it looks different depending on what the parent class expects by default.

sdroege:
- `parent_child_notify()` should do nothing instead of panicking
- `parent_compute_expand()` is difficult. We need to replicate the behaviour the base class would do if it's not set here... basically we would need to set the two bools to the return values of `gtk_widget_get_hexpand()` and `gtk_widget_get_vexpand()`.
- Application, container, dialog, window need the same treatment (I think all of them are cases of "do-nothing-if-not-set")

src/subclass/widget.rs
EPashkin: Ok.

sdroege:
Looks good, I think. Only container and widget's compute_expand() missing now :)

alatiera:
Added compute expand now, though given how simple it looks I am not sure if that's what you had in mind?

sdroege:
> Added compute expand now, though given how simple it looks I am not sure if that's what you had in mind?

Yes exactly that. Then only the GtkContainer virtual methods are missing and then we're good to go here :)

alatiera:
oh whoops, forgot aobut container

sdroege:
:+1:

src/subclass/container.rs
```diff
@@ -51,65 +51,59 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
...
 
     fn parent_get_path_for_child(&self, container: &Container, widget: &Widget) -> WidgetPath {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .get_path_for_child
-                .expect("No parent class impl for \"get_path_for_child\"");
-            from_glib_borrow(f(container.to_glib_none().0, widget.to_glib_none().0))
+            if let Some(f) = (*parent_class).get_path_for_child {
```
alatiera: This doesn't make sense, unless it returns a None existing path, which also doesn't really make sense

src/subclass/container.rs
```diff
@@ -51,65 +51,59 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
...
     fn parent_child_type(&self, container: &Container) -> glib::Type {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
-            let f = (*parent_class)
-                .child_type
-                .expect("No parent class impl for \"child_type\"");
-            from_glib(f(container.to_glib_none().0))
+            if let Some(f) = (*parent_class).child_type {
+                from_glib(f(container.to_glib_none().0))
```
alatiera: This ain't compiling, go me. Also I don't think it makes sense for it to fallback to G_TYPE_NONE instead of panicking, this should really not be missing.

sdroege: The docs say:
> Note that this may return G_TYPE_NONE to indicate that no more children can be added

So yes, return `G_TYPE_NONE` here as a default :)

src/subclass/container.rs
sdroege: Alternatively you could return a `gtk_widget_path_new()`, which is the empty path. Should be fine to panic here though, it seems to make no sense to have child elements but no way of returning a path.

src/subclass/container.rs
sdroege: See also https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkcontainer.c#L1036-1051

src/subclass/container.rs
sdroege: This should panic indeed, see https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkcontainer.c#L3874-3895. The C version would simply crash.

alatiera:
Should be good to go!

sdroege:
> Should be good to go!

I agree :) Also your examples PR should be good to merge then?

alatiera:
Okay, that's the last of the panics I am hitting it seems \o/

sdroege:
> Okay, that's the last of the panics I am hitting it seems \o/

:raised_hands: Let's get this in then once the CI is green

GuillaumeGomez:
Thanks!

PR #927, opened 2019-12-20
Delete LGPL

GuillaumeGomez:
Just realised we had this file around whereas it wasn't used...

cc @sdroege @EPashkin


sdroege:
:+1:

PR #937, opened 2020-01-10
Add subclass for CellRenderer and CellRendererPixbuf (Closes #936)

### LINKED ISSUE #936: Can't subclass `gtk::CellRendererPixbuf`
It doesn't have `CellRendererPixbufImpl`

hfiguiere:
So I can subclass.

### DIFF
--- a/src/subclass/cell_renderer.rs
+++ b/src/subclass/cell_renderer.rs
@@ -1,0 +1,699 @@
+use cairo_sys;
+use gdk_sys;
+use glib_sys;
+use gtk_sys;
+
+use libc::{c_char, c_int};
+use std::mem;
+
+use cairo;
+use glib::object::IsA;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::GString;
+use glib::ObjectClass;
+
+use CellEditable;
+use CellRenderer;
+use CellRendererClass;
+use CellRendererState;
+use SizeRequestMode;
+use Widget;
+
+pub trait CellRendererImpl: CellRendererImplExt + ObjectImpl + 'static {
+    fn get_request_mode(&self, renderer: &CellRenderer) -> SizeRequestMode {
+        self.parent_get_request_mode(renderer)
+    }
+
+    fn get_preferred_width<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32) {
+        self.parent_get_preferred_width(renderer, widget)
+    }
+
+    fn get_preferred_width_for_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        height: i32,
+    ) -> (i32, i32) {
+        self.parent_get_preferred_width_for_height(renderer, widget, height)
+    }
+
+    fn get_preferred_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32) {
+        self.parent_get_preferred_height(renderer, widget)
+    }
+
+    fn get_preferred_height_for_width<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        width: i32,
+    ) -> (i32, i32) {
+        self.parent_get_preferred_height_for_width(renderer, widget, width)
+    }
+
+    fn get_aligned_area<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        flags: CellRendererState,
+        cell_area: &gdk::Rectangle,
+    ) -> gdk::Rectangle {
+        self.parent_get_aligned_area(renderer, widget, flags, cell_area)
+    }
+
+    fn render<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        cr: &cairo::Context,
+        widget: &P,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) {
+        self.parent_render(renderer, cr, widget, background_area, cell_area, flags);
+    }
+
+    fn activate<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        event: &gdk::Event,
+        widget: &P,
+        path: &str,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) -> bool {
+        self.parent_activate(
+            renderer,
+            event,
+            widget,
+            path,
+            background_area,
+            cell_area,
+            flags,
+        )
+    }
+
+    fn start_editing<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        event: Option<&gdk::Event>,
+        widget: &P,
+        path: &str,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) -> Option<CellEditable> {
+        self.parent_start_editing(
+            renderer,
+            event,
+            widget,
+            path,
+            background_area,
+            cell_area,
+            flags,
+        )
+    }
+
+    fn editing_canceled(&self, renderer: &CellRenderer) {
+        self.parent_editing_canceled(renderer)
+    }
+
+    fn editing_started(&self, renderer: &CellRenderer, editable: &CellEditable, path: &str) {
+        self.parent_editing_started(renderer, editable, path)
+    }
+}
+
+pub trait CellRendererImplExt {
+    fn parent_get_request_mode(&self, renderer: &CellRenderer) -> SizeRequestMode;
+    fn parent_get_preferred_width<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32);
+    fn parent_get_preferred_width_for_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        height: i32,
+    ) -> (i32, i32);
+    fn parent_get_preferred_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32);
+    fn parent_get_preferred_height_for_width<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        width: i32,
+    ) -> (i32, i32);
+    fn parent_get_aligned_area<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        flags: CellRendererState,
+        cell_area: &gdk::Rectangle,
+    ) -> gdk::Rectangle;
+    fn parent_render<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        cr: &cairo::Context,
+        widget: &P,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    );
+    fn parent_activate<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        event: &gdk::Event,
+        widget: &P,
+        path: &str,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) -> bool;
+    fn parent_start_editing<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        event: Option<&gdk::Event>,
+        widget: &P,
+        path: &str,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) -> Option<CellEditable>;
+    fn parent_editing_canceled(&self, renderer: &CellRenderer);
+    fn parent_editing_started(&self, renderer: &CellRenderer, editable: &CellEditable, path: &str);
+}
+
+impl<T: CellRendererImpl + ObjectImpl> CellRendererImplExt for T {
+    fn parent_get_request_mode(&self, renderer: &CellRenderer) -> SizeRequestMode {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            let f = (*parent_class).get_request_mode.unwrap();
+            from_glib(f(renderer.to_glib_none().0))
+        }
+    }
+
+    fn parent_get_preferred_width<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            let f = (*parent_class).get_preferred_width.unwrap();
+
+            let mut minimum_size = mem::MaybeUninit::uninit();
+            let mut natural_size = mem::MaybeUninit::uninit();
+            f(
+                renderer.to_glib_none().0,
+                widget.as_ref().to_glib_none().0,
+                minimum_size.as_mut_ptr(),
+                natural_size.as_mut_ptr(),
+            );
+            (minimum_size.assume_init(), natural_size.assume_init())
+        }
+    }
+
+    fn parent_get_preferred_width_for_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+        height: i32,
+    ) -> (i32, i32) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            let f = (*parent_class).get_preferred_width_for_height.unwrap();
+
+            let mut minimum_size = mem::MaybeUninit::uninit();
+            let mut natural_size = mem::MaybeUninit::uninit();
+            f(
+                renderer.to_glib_none().0,
+                widget.as_ref().to_glib_none().0,
+                height,
+                minimum_size.as_mut_ptr(),
+                natural_size.as_mut_ptr(),
+            );
+            (minimum_size.assume_init(), natural_size.assume_init())
+        }
+    }
+    fn parent_get_preferred_height<P: IsA<Widget>>(
+        &self,
+        renderer: &CellRenderer,
+        widget: &P,
+    ) -> (i32, i32) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            let f = (*parent_class).get_preferred_height.unwrap();
+            let mut minimum_size = mem::MaybeUninit::uninit();
+            let mut natural_size = mem::MaybeUninit::uninit();
+            f(
+                renderer.to_glib_none().0,
+                widget.as_ref().to_glib_none().0,
+                minimum_size.as_mut_ptr(),
+                natural_size.as_mut_ptr(),
+            );
+            (minimum_size.assume_init(), natural_size.assume_init())
+        }
+    }
+    fn parent_get_preferred_height_for_width<P: IsA<Widget>>(
+        &self,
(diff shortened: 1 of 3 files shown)


src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,12 @@
+use glib::subclass::prelude::*;
+use glib::ObjectClass;
+
+use CellRendererClass;
+
+pub trait CellRendererImpl: ObjectImpl + 'static {}
+
+unsafe impl<T: ObjectSubclass + CellRendererImpl> IsSubclassable<T> for CellRendererClass {
```
sdroege: Please run `cargo fmt` once :)

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,12 @@
...
+    fn override_vfuncs(&mut self) {
+        <ObjectClass as IsSubclassable<T>>::override_vfuncs(self);
+    }
+}
```
sdroege: The `CellRenderer` class has various vfuncs. You don't need any of those for what you're doing?

src/subclass/cell_renderer_pixbuf.rs
```diff
@@ -0,0 +1,13 @@
+use glib::subclass::prelude::*;
+
+use super::cell_renderer::CellRendererImpl;
+use CellRendererClass;
+use CellRendererPixbufClass;
+
+pub trait CellRendererPixbufImpl: CellRendererImpl + 'static {}
+
+unsafe impl<T: ObjectSubclass + CellRendererPixbufImpl> IsSubclassable<T> for CellRendererPixbufClass {
```
sdroege: Please run `cargo fmt` once :)

sdroege:
Looks good to me otherwise

src/subclass/cell_renderer.rs
hfiguiere: it's more like "cut and paste" I'm doing here. Likely I will need them.

sdroege: Ok, I guess you'll send a new PR at that point then :)

src/subclass/cell_renderer_pixbuf.rs
hfiguiere: done now

sdroege: Thanks!

src/subclass/cell_renderer.rs
hfiguiere: or maybe do it in that one.... just need to figure out what to do.

sdroege: Ask on IRC if you need help :)

hfiguiere: Ok. It feels very complicated, but I think I got it, at least a draft....

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+        self.parent_start_editing(
+            renderer,
+            event,
+            widget,
+            path,
+            background_area,
+            cell_area,
+            flags,
+        )
+    }
```
sdroege: There's also `editing_chancelled` and `editing_started`

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+    fn parent_get_request_mode(&self, renderer: &CellRenderer) -> SizeRequestMode {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            if let Some(f) = (*parent_class).get_request_mode {
+                from_glib(f(renderer.to_glib_none().0))
+            } else {
+                SizeRequestMode::__Unknown(0)
+            }
```
sdroege: As the C code unconditionally uses the vfunc in `gtk_cell_renderer_get_request_mode()` we should panic if it does not exist

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    renderer.to_glib_none().0,
+                    widget.as_ref().to_glib_none().0,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                );
+                let minimum_size = minimum_size.assume_init();
+                let natural_size = natural_size.assume_init();
+                (minimum_size, natural_size)
+            } else {
+                (-1, -1)
```
sdroege: Also here panic

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    widget.as_ref().to_glib_none().0,
+                    height,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                );
+                let minimum_size = minimum_size.assume_init();
+                let natural_size = natural_size.assume_init();
+                (minimum_size, natural_size)
+            } else {
+                (-1, -1)
```
sdroege: Also here panic

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    renderer.to_glib_none().0,
+                    widget.as_ref().to_glib_none().0,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                );
+                let minimum_size = minimum_size.assume_init();
+                let natural_size = natural_size.assume_init();
+                (minimum_size, natural_size)
+            } else {
+                (-1, -1)
```
sdroege: Also here panic

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    widget.as_ref().to_glib_none().0,
+                    width,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                );
+                let minimum_size = minimum_size.assume_init();
+                let natural_size = natural_size.assume_init();
+                (minimum_size, natural_size)
+            } else {
+                (-1, -1)
```
sdroege: Also here panic

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                f(
+                    renderer.to_glib_none().0,
+                    widget.as_ref().to_glib_none().0,
+                    flags.to_glib(),
+                    cell_area.to_glib_none().0,
+                    aligned_area.to_glib_none_mut().0,
+                );
+            }
+            aligned_area
+        }
```
sdroege: Panic if it does not exist because `gtk_cell_renderer_get_aligned_area()` also uses it unconditionally

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+        widget: &P,
+        background_area: &gdk::Rectangle,
+        cell_area: &gdk::Rectangle,
+        flags: CellRendererState,
+    ) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            if let Some(f) = (*parent_class).render {
```
sdroege: `gtk_cell_renderer_render()` does a critical warning but does not crash if it's `NULL`. So doing nothing as well as panicking should both be fine here. Your choice :)

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    renderer.to_glib_none().0,
+                    mut_override(event.to_glib_none().0),
+                    widget.as_ref().to_glib_none().0,
+                    path.to_glib_none().0,
+                    background_area.to_glib_none().0,
+                    cell_area.to_glib_none().0,
+                    flags.to_glib(),
+                ))
+            } else {
+                false
```
sdroege: `gtk_cell_renderer_activate()` returns `FALSE` if it's `NULL`, so this is indeed the right behaviour :)

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                    renderer.to_glib_none().0,
+                    mut_override(event.to_glib_none().0),
+                    widget.as_ref().to_glib_none().0,
+                    path.to_glib_none().0,
+                    background_area.to_glib_none().0,
+                    cell_area.to_glib_none().0,
+                    flags.to_glib(),
+                ))
+            } else {
+                None
```
sdroege: :+1:

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+        renderer: &CellRenderer,
+        widget: &P,
+        flags: CellRendererState,
+        cell_area: &gdk::Rectangle,
+    ) -> gdk::Rectangle {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class =
+                data.as_ref().get_parent_class() as *mut gtk_sys::GtkCellRendererClass;
+            let mut aligned_area = gdk::Rectangle::uninitialized();
```
sdroege: The correct pattern for this nowadays (we should deprecated `::uninitialize()`...) is

```rust
let mut aligned_area = mem::MaybeUninitialized::uninit();
...
    aligned_area.as_mut_ptr(),
...

from_glib_none(aligned_area)
```

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+                f(
+                    renderer.to_glib_none().0,
+                    widget.as_ref().to_glib_none().0,
+                    width,
+                    minimum_size.as_mut_ptr(),
+                    natural_size.as_mut_ptr(),
+                );
+                let minimum_size = minimum_size.assume_init();
+                let natural_size = natural_size.assume_init();
+                (minimum_size, natural_size)
```
sdroege: If you want you can move the `assume_init()` calls directly into the return expression

sdroege: Not just here but everywhere :)

src/subclass/cell_renderer.rs
hfiguiere: ah right. I forgot why I left out the signal, but it was part of me not knowing what I was doing.

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+    let wrap: CellRenderer = from_glib_borrow(ptr);
+    let widget: Widget = from_glib_borrow(wdgtptr);
+
+    let rectangle = imp.get_aligned_area(
+        &wrap,
+        &widget,
+        from_glib(flags),
+        &from_glib_borrow(cellarea),
+    );
+    if !alignedptr.is_null() {
```
sdroege: Not nullable according to the GTK code (various parts would dereference NULL without checking)

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+    flags: gtk_sys::GtkCellRendererState,
+    cellarea: *const gdk_sys::GdkRectangle,
+    alignedptr: *mut gdk_sys::GdkRectangle,
+) where
+    T: CellRendererImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: CellRenderer = from_glib_borrow(ptr);
+    let widget: Widget = from_glib_borrow(wdgtptr);
```
sdroege: All these type annotations should not be needed but feel free to keep them

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+    flags: gtk_sys::GtkCellRendererState,
+) -> glib_sys::gboolean
+where
+    T: CellRendererImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: CellRenderer = from_glib_borrow(ptr);
+    let widget: Widget = from_glib_borrow(wdgtptr);
+    let evt: Option<gdk::Event> = from_glib_none(evtptr);
```
sdroege:
```suggestion
    let evt: Option<gdk::Event> = from_glib_borrow(evtptr);
```

And then below `evt.as_ref()`

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+    flags: gtk_sys::GtkCellRendererState,
+) -> *mut gtk_sys::GtkCellEditable
+where
+    T: CellRendererImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: CellRenderer = from_glib_borrow(ptr);
+    let widget: Widget = from_glib_borrow(wdgtptr);
+    let evt: Option<gdk::Event> = from_glib_none(evtptr);
```
sdroege: See https://github.com/gtk-rs/gtk/pull/937/files#r365506112

src/subclass/cell_renderer.rs
sdroege: > Ok. It feels very complicated, but I think I got it, at least a draft....

Looks good to me mostly :) To give some background.

It's indeed a bit complicated and we can't unfortunately autogenerate this even in your case here because of the default handling when chaining up to the parent class' vfunc. That part we must get right as we always override all vfuncs, so even if you subclass does nothing then the chaining-up code is called and has to behave correctly.

Bindings for languages with a virtual machine/interpreter don't have that problem because they only override the vfuncs at runtime for the ones that are actually used by the subclass (e.g. Python). We can't do that in Rust because that information does not exist at runtime or even compile-time.

Other bindings just always do nothing and if there's a return value return "some default" and pray that all goes well (C++ for example), which we can't do because soundness is important for Rust developers. In C++ bindings can get away with blaming the developer if stuff crashes or misbehaves (you should've overridden this virtual method!) but in Rust that's not a choice for safe API if someone is supposed to take the bindings seriously.

I'm not sure how this can be improved. For the common cases we could probably get annotations (default return value / do nothing) but in other cases the default behaviour is complicated and nothing you really want to put into a textual annotation.

src/subclass/cell_renderer.rs
```diff
@@ -0,0 +1,658 @@
...
+
+use libc::{c_char, c_int};
+use std::mem;
+
+use cairo;
+use glib::object::IsA;
+use glib::subclass::prelude::*;
+use glib::translate::*;
+use glib::ObjectClass;
+use glib::GString;
```
sdroege: Misses a run of `cargo fmt`

src/subclass/cell_renderer.rs
sdroege: No worries :) These are virtual functions for the default handlers of these signals. With regard to subclassing you handle them exactly like others, doesn't matter that the signal machinery calls into them instead of other code.

src/subclass/cell_renderer.rs
hfiguiere: I wasn't sure here so I took the safe route. Changing it.

src/subclass/cell_renderer.rs
hfiguiere: I'd rather not crash.

src/subclass/cell_renderer.rs
hfiguiere: ok

src/subclass/cell_renderer.rs
hfiguiere: done

src/subclass/cell_renderer.rs
hfiguiere: ok

src/subclass/cell_renderer.rs
hfiguiere: yeah I realized afterwards I forgot one last round.

src/subclass/cell_renderer.rs
sdroege: `g_return_val_if_fail()` is basically like a panic in Rust, so that would be the consistent way with the C code. But just doing nothing here seems equally fine IMHO

src/subclass/cell_renderer.rs
hfiguiere: It actually doesn't compile as is:
````
   Compiling gtk v0.8.0 (/home/hub/source/gtk-rs)
error[E0277]: the trait bound `gdk_sys::GdkRectangle: glib::translate::Ptr` is not satisfied
    --> src/subclass/cell_renderer.rs:312:28
     |
312  |             from_glib_none(aligned_area.assume_init())
     |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `glib::translate::Ptr` is not implemented for `gdk_sys::GdkRectangle`
     | 
    ::: /home/hub/.cargo/git/checkouts/glib-928cf7b282977403/597cba3/src/translate.rs:1131:33
     |
1131 | pub unsafe fn from_glib_none<P: Ptr, T: FromGlibPtrNone<P>>(ptr: P) -> T {
     |                                 --- required by this bound in `glib::translate::from_glib_none`

error[E0277]: the trait bound `gdk::Rectangle: glib::translate::FromGlibPtrNone<gdk_sys::GdkRectangle>` is not satisfied
    --> src/subclass/cell_renderer.rs:312:13
     |
312  |             from_glib_none(aligned_area.assume_init())
     |             ^^^^^^^^^^^^^^ the trait `glib::translate::FromGlibPtrNone<gdk_sys::GdkRectangle>` is not implemented for `gdk::Rectangle`
     | 
    ::: /home/hub/.cargo/git/checkouts/glib-928cf7b282977403/597cba3/src/translate.rs:1131:41
     |
1131 | pub unsafe fn from_glib_none<P: Ptr, T: FromGlibPtrNone<P>>(ptr: P) -> T {
     |                                         ------------------ required by this bound in `glib::translate::from_glib_none`
     |
     = help: the following implementations were found:
               <gdk::Rectangle as glib::translate::FromGlibPtrNone<*const gdk_sys::GdkRectangle>>
               <gdk::Rectangle as glib::translate::FromGlibPtrNone<*mut gdk_sys::GdkRectangle>>

error: aborting due to 2 previous errors
````

src/subclass/cell_renderer.rs
hfiguiere: ok

src/subclass/cell_renderer.rs
sdroege: `from_glib_none(&aligned_area.assume_init())` should work. It needs a pointer here.

src/subclass/cell_renderer.rs
hfiguiere: Actually `let evt: gdk::Event = from_glib_borrow(evtptr);` works here, and then pass `&evt`

src/subclass/cell_renderer.rs
hfiguiere: Changed to `from_glib_borrow`

src/subclass/cell_renderer.rs
hfiguiere: it doesn't.

src/subclass/cell_renderer.rs
sdroege: Is it guaranteed to be never NULL? Then this is fine. I thought you might have had a reason for making it an `Option` :)

src/subclass/cell_renderer.rs
sdroege: Ok, I'll take a look in a bit once you updated your branch for the other things :)

src/subclass/cell_renderer.rs
hfiguiere: the generated activate expect a reference to the GdkEvent so I did the same.

sdroege: :+1:

src/subclass/cell_renderer.rs
hfiguiere: This is done now.

hfiguiere:
I'm thinking squashing these commits once it is ready to merge.

sdroege:
> I'm thinking squashing these commits once it is ready to merge.

That's a good idea :)

src/subclass/cell_renderer.rs
sdroege: `from_glib_none(&aligned_area.assume_init() as *const _)` would be the solution. But as the autogenerated code still uses `uninitialized()`, let's keep it as is for now for consistency.

We'll change this later once we got rid of `uninitialized()` in general and also changed the code generator accordingly.

sdroege:
Looks all good to me, squash it into one commit and then let's merge it once the CI is green :) Thanks!

hfiguiere:
I have squashed the commits into one

src/subclass/cell_renderer.rs
hfiguiere: I wonder if we can't autogenerate this too...

sdroege:
Perfect, thanks :)

src/subclass/cell_renderer.rs
sdroege: See https://github.com/gtk-rs/gtk/pull/937#discussion_r365506403 . A subset of these can potentially be autogenerated (there's an old PR against gir for this), yes.

sdroege:
@GuillaumeGomez CI is green, please merge :)

PR #931, opened 2020-01-02
Add subclass for IconView

GuillaumeGomez:
Fixes #928.

r? @sdroege @EPashkin 

Also, should we add an option in gir (or extend `not_bound`) to list not yet implemented subclasses?

PR #940, opened 2020-01-12
Added vfunc to IconView

hfiguiere:
This is a follow up on #931 - vfuncs were missing.

### DIFF
--- a/src/subclass/icon_view.rs
+++ b/src/subclass/icon_view.rs
@@ -1,13 +1,255 @@
+use glib_sys;
+use gtk_sys;
+
+use libc::c_int;
+
 use glib::subclass::prelude::*;
+use glib::translate::*;
 
 use super::container::ContainerImpl;
 use ContainerClass;
+use IconView;
 use IconViewClass;
+use MovementStep;
+use TreePath;
 
-pub trait IconViewImpl: ContainerImpl + 'static {}
+pub trait IconViewImpl: IconViewImplExt + ContainerImpl + 'static {
+    fn item_activated(&self, icon_view: &IconView, path: &TreePath) {
+        self.parent_item_activated(icon_view, path)
+    }
+    fn selection_changed(&self, icon_view: &IconView) {
+        self.parent_selection_changed(icon_view)
+    }
+    fn select_all(&self, icon_view: &IconView) {
+        self.parent_select_all(icon_view)
+    }
+    fn unselect_all(&self, icon_view: &IconView) {
+        self.parent_unselect_all(icon_view)
+    }
+    fn select_cursor_item(&self, icon_view: &IconView) {
+        self.parent_select_cursor_item(icon_view)
+    }
+    fn toggle_cursor_item(&self, icon_view: &IconView) {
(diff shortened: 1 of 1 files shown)


src/subclass/icon_view.rs
```diff
@@ -1,13 +1,252 @@
...
 
 use super::container::ContainerImpl;
 use ContainerClass;
+use IconView;
 use IconViewClass;
+use MovementStep;
+use TreePath;
+
+pub trait IconViewImpl: IconViewImplExt + ContainerImpl + 'static {
+    fn item_activated(&self, icon_view: &IconView, path: &mut TreePath) {
```
sdroege: This should probably not be a mutable reference, or is this supposed to be changeable by the subclass? It maps to a signal and to a normal method, the former doesn't mention anything about it being modifiable, the latter takes an immutable reference.

sdroege:
Looks good to me otherwise

src/subclass/icon_view.rs
hfiguiere: I had no choice. The underlying vfunc in the class expect a `*mut` and it seems it's the only way.

`    pub item_activated: Option<unsafe extern "C" fn(*mut GtkIconView, *mut GtkTreePath)>,`

EPashkin: Note that it `&IconView` even it `*mut GtkIconView` 😉 
So better check docs and possible just remove `mut`

sdroege: `glib::translate::mut_override()` to go from `*const T` to `*mut T`. That's generally caused by C API not using `const` correctly, so all pointers are `*mut` on the Rust side.

EPashkin: @sdroege Strange, but seems this function is only one that not initialized by default even if gtk4
https://github.com/GNOME/gtk/blob/mainline/gtk/gtkiconview.c#L369
https://github.com/GNOME/gtk/blob/mainline/gtk/gtkiconviewprivate.h#L55

sdroege: Yeah, just doing nothing if not implemented by a subclass.

hfiguiere: ok I use `mut_override()` here. And removed mut

src/subclass/icon_view.rs
```diff
@@ -1,13 +1,255 @@
...
+unsafe extern "C" fn icon_view_item_activated<T: ObjectSubclass>(
+    ptr: *mut gtk_sys::GtkIconView,
+    path: *mut gtk_sys::GtkTreePath,
+) where
+    T: IconViewImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap = from_glib_borrow(ptr);
+    let path = from_glib_full(path);
```
sdroege:
```suggestion
    let path = from_glib_borrow(path);
```

hfiguiere: done

sdroege:
@GuillaumeGomez Please merge :)

GuillaumeGomez:
Thanks!

PR #948, opened 2020-02-01
InfoBar::get_content_area return Box

### LINKED ISSUE #947: InfoBar get_content_area() should be a Gtk::Box
Currently `InfoBarEx::get_content_area()` returns an `Option<Widget>`, but I believe it should return a `Box`. If we look at the gtk apidoc:

>  To add widgets to the main content area of a InfoBar, use InfoBarExt::get_content_area and add your widgets to the container.

and:

```c
content_area = gtk_info_bar_get_content_area (bar);
gtk_container_add (GTK_CONTAINER (content_area),
                   message_label);
```

also, in the gtk source:
https://github.com/GNOME/gtk/blob/master/gtk/gtkinfobar.c#L490

```c
priv->content_area = gtk_box_new (GTK_ORIENTATION_HORIZONTAL, 0);
```

So I think gtk-rs should update the type in the bindings from `Option<gtk::Widget>` to probably `gtk::Box`.

Assuming I'm correct, I have another question.. Is there a way for me to upcast the `gtk::Widget` that I get from gtk-rs to a `Box`, even through a hack, as a temporary measure until gtk-rs releases a new version with the change?

EPashkin:
Fix #947 

cc @sdroege, @GuillaumeGomez , @emmanueltouzery

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -1222,6 +1222,9 @@ status = "generate"
         [object.function.return]
         nullable = false
         type = "Gtk.Box"
+    [[object.function]]
+    name = "get_action_area"
+    ignore = true
 
 [[object]]
 name = "Gtk.InputHints"
--- a/Gir.toml
+++ b/Gir.toml
@@ -1216,6 +1216,12 @@ status = "generate"
     [[object.property]]
     name = "revealed"
     version = "3.22.29"
+    [[object.function]]
+    name = "get_content_area"
+    ignore = true
+        [object.function.return]
+        nullable = false
+        type = "Gtk.Box"
 
 [[object]]
 name = "Gtk.InputHints"
--- a/src/auto/info_bar.rs
+++ b/src/auto/info_bar.rs
@@ -473,8 +473,6 @@ pub trait InfoBarExt: 'static {
 
     fn get_action_area(&self) -> Option<Widget>;
 
-    fn get_content_area(&self) -> Option<Widget>;
-
     fn get_message_type(&self) -> MessageType;
 
     #[cfg(any(feature = "v3_22_29", feature = "dox"))]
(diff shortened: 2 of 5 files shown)


EPashkin:
We don't have auto-downcast to type changed, it need be done manually :(

EPashkin:
Implemented it manually

EPashkin:
CI passed except for OSX

EPashkin:
Restarted osx jobs

EPashkin:
Osx on beta rust works more that 50min for second time and was terminated again.
Splitting build in 2 part (https://github.com/gtk-rs/examples/pull/287) don't helps too.

sdroege:
Yeah something's wrong on macos lately. Not sure what to do about that. @GuillaumeGomez can you reproduce that on your machine?

Gir.toml
```diff
@@ -1216,6 +1216,12 @@ status = "generate"
     [[object.property]]
     name = "revealed"
     version = "3.22.29"
+    [[object.function]]
+    name = "get_content_area"
```
sdroege: Also `get_action_area` is the same story :)

EPashkin:
Added `get_action_area`, by way it was GtkButtonBox in gtk3

EPashkin:
@sdroege IMHO nothing wrong with macos itself, just rustc with "link-dead-code" do more work and it hit timeout on beta (also rustc seems has some regression)

GuillaumeGomez:
What dead links btw? And let's open an issue for the timeout problem so we can work on it. I'll try to reproduce it locally.

EPashkin:
`"link-dead-code"` links unused functions to and allow detect missing functions,
I added it in https://github.com/gtk-rs/examples/pull/283.

Issue #949, opened 2020-02-01
How to get the size of currently used font?


ArekPiekarz:
I've been trying to find a way to get the size of the currently used font in gtk::TextView. My use case is to dynamically increase or decrease the font size in reaction to a key press event.

It looks like pango::FontDescription::get_size() contains that information. It could be taken from gtk_sys::GtkStyle, but there is no function that returns it in Rust version of gtk. I found that an almost working workaround would be to use gtk::Widget::get_property("style"), but it returns a wrapped glib::Value and conversion from it is not implemented for GtkStyle.

It seems this issue - https://github.com/gtk-rs/gtk/issues/874 - may be a blocker for my case. Do you see any other way to solve it?

EPashkin:
`GtkStyle` not included in current gtk crate at all.
Theoretically you can use https://gtk-rs.org/docs/gtk/trait.WidgetExt.html#tymethod.get_style_context to get current style context, but it don't have https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-font as it deprecated and https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get as it use varargs,
so seems only solution is to call `gtk_style_context_get` manually to get font.

ArekPiekarz:
Thanks, I was able to get the font size with a following workaround:
```rust
let styleContext = widget.get_style_context();
let fontDescriptionPtr = unsafe {gtk_sys::gtk_style_context_get_font(
    styleContext.as_ptr(), gtk_sys::GTK_STATE_FLAG_NORMAL)};
let fontPangoSize = unsafe {pango_sys::pango_font_description_get_size(
    fontDescriptionPtr)}; // gives 9216 Pango units
let fontPointSize = fontPangoSize / pango_sys::PANGO_SCALE; // gives 9 points
```

EPashkin:
@ArekPiekarz It will be good if you can get working code with `gtk-style-context-get` instead of `gtk_style_context_get_font`, wrap it to `FontDescription` with `from_glib_none` and make PR with new manual function in StyleContext.

ArekPiekarz:
After your comment I noticed `gtk_style_context_get_font` was deprecated, so here's the newer version using the variadic `gtk_style_context_get`:
```rust
let fontDesciptionPtr: *mut pango_sys::PangoFontDescription = std::ptr::null_mut();
let fontProperty = std::ffi::CString::new("font").unwrap();
[...]

PR #953, opened 2020-02-07
Add StyleContext::get_font, deignore Value getters

EPashkin:
Fix of #949

cc @GuillaumeGomez , @sdroege


src/style_context.rs
```diff
@@ -0,0 +1,31 @@
...
+use StyleContext;
+
+pub trait StyleContextExtManual: 'static {
+    fn get_font(&self) -> Option<FontDescription>;
+}
+
+impl<O: IsA<StyleContext>> StyleContextExtManual for O {
+    fn get_font(&self) -> Option<FontDescription> {
+        unsafe {
+            let font_desciption_ptr: *mut PangoFontDescription = std::ptr::null_mut();
```
GuillaumeGomez: typo: "font_desciption_ptr" -> "font_description_ptr"

EPashkin: Thanks, fixed.

GuillaumeGomez:
Looks good to me, thanks!

EPashkin:
CI "passed"

src/style_context.rs
```diff
@@ -0,0 +1,31 @@
...
+    fn get_font(&self) -> Option<FontDescription>;
+}
+
+impl<O: IsA<StyleContext>> StyleContextExtManual for O {
+    fn get_font(&self) -> Option<FontDescription> {
+        unsafe {
+            let font_description_ptr: *mut PangoFontDescription = std::ptr::null_mut();
+            gtk_sys::gtk_style_context_get(
+                self.as_ref().to_glib_none().0,
+                gtk_sys::GTK_STATE_FLAG_NORMAL,
```
sdroege: Shouldn't the state flags also be passed into the function by the caller?

EPashkin: You right :(

sdroege: The getter for border, color, etc have that at least.

src/style_context.rs
```diff
@@ -0,0 +1,31 @@
...
+use glib::object::IsA;
+use glib::translate::*;
+use gtk_sys;
+use libc::c_void;
+use pango::FontDescription;
+use pango_sys::PangoFontDescription;
+use StyleContext;
+
+pub trait StyleContextExtManual: 'static {
+    fn get_font(&self) -> Option<FontDescription>;
```
sdroege: Why don't we bind `gtk_style_context_get_property()` btw? That would be the generic getter that allows to get any kind of property by string name

EPashkin: It was ignored with `get_style_property` in 2016 (4e6d5db1991bdd923069d802d3250ea0ea15f655), now it works and we can try to de-ignore it and others

sdroege: You'll add it as part of this PR?

EPashkin: I tried and found that most of ignored functions generated like this.
@sdroege What you think, we can leave as is or implement it manually as -> glib::Value ?
```
    fn style_get_property(&self, property_name: &str, value: &mut glib::Value) {
        unsafe {
            gtk_sys::gtk_widget_style_get_property(
                self.as_ref().to_glib_none().0,
                property_name.to_glib_none().0,
                value.to_glib_none_mut().0,
            );
        }
    }
```

src/style_context.rs
EPashkin: Fixed

src/style_context.rs
EPashkin: Added temporary commit with generation results

sdroege: As `-> glib::Value` or `-> Option<glib::Value>` (or `Result`) :(

Going with an out parameter is problematic here safety-wise, and also a bit annoying usage-wise.

EPashkin: Ok I will implement it manually.

EPashkin: For fix gir to generate it right

EPashkin: Seems gir-file also somewhat wrong:
gtk_style_context_get_property generate return,
```
          <parameter name="value" direction="out" caller-allocates="1" transfer-ownership="full">
            <doc xml:space="preserve" filename="gtk/gtkstylecontext.c" line="797">return location for the style property value</doc>
            <type name="GObject.Value" c:type="GValue*"/>
          </parameter>
```
gtk_style_context_get_style_property: generate &mut Value
```
          <parameter name="value" transfer-ownership="none">
            <doc xml:space="preserve" filename="gtk/gtkstylecontext.c" line="1727">Return location for the property value</doc>
            <type name="GObject.Value" c:type="GValue*"/>
          </parameter>
```

sdroege: Indeed. Can you create issues for that against gtk, or maybe even a MR fixing the annotations? :)

sdroege: And maybe fix the .gir file for now so that we can autogenerate these already

EPashkin: @sdroege Seems no such problem in gtk4, most functions was just removed (Ex. 
https://github.com/GNOME/gtk/commit/632fab0ca702b600fde492319ab0aa2d9d3bbd08)

You don't know, is https://gitlab.gnome.org/GNOME/gtk/tree/gtk-3-24 good for adding PR against it or there other branch for last gtk3

EPashkin: Upstream issue https://gitlab.gnome.org/GNOME/gtk/issues/2418 with link to unchecked fix

EPashkin:
@sdroege , @GuillaumeGomez added generation functions with `glib::ParamSpec`

src/style_context.rs
sdroege: Try your look with a MR against the GTK 3 branch. Maybe there will be another bugfix release with that :)

src/auto/settings.rs
```diff
@@ -43,13 +44,16 @@ impl Settings {
...
 
-    //#[cfg_attr(feature = "v3_16", deprecated)]
-    //pub fn install_property(pspec: /*Ignored*/&glib::ParamSpec) {
-    //    unsafe { TODO: call gtk_sys:gtk_settings_install_property() }
-    //}
+    #[cfg_attr(feature = "v3_16", deprecated)]
+    pub fn install_property(pspec: &glib::ParamSpec) {
+        assert_initialized_main_thread!();
+        unsafe {
+            gtk_sys::gtk_settings_install_property(pspec.to_glib_none().0);
```
sdroege: > gtk_settings_install_property has been deprecated since version 3.16 and should not be used in newly-written code.
This function is not useful outside GTK+.

sdroege: So let's not add it :)

src/style_context.rs
```diff
@@ -0,0 +1,32 @@
...
+
+pub trait StyleContextExtManual: 'static {
+    fn get_font(&self, state: StateFlags) -> Option<FontDescription>;
+}
+
+impl<O: IsA<StyleContext>> StyleContextExtManual for O {
+    fn get_font(&self, state: StateFlags) -> Option<FontDescription> {
+        unsafe {
+            let font_description_ptr: *mut PangoFontDescription = std::ptr::null_mut();
+            gtk_sys::gtk_style_context_get(
```
sdroege: You can do this completely without unsafe code now by calling `get_property()` :)

sdroege:
Looks good to me but we should merge the `gir-files` PR first.

Also I wonder if any of the functions returning a `GValue` can return a value with `G_TYPE_INVALID` or so and for that reason should return an `Option` or `Result` depending on the context. But I guess for now this is OK and we'll find out sooner or later.

src/style_context.rs
EPashkin: I don't know what branch is "dev" for gkt3, so just selected last version

src/auto/settings.rs
EPashkin: removed it

EPashkin:
Use get_property  in get_font.

src/style_context.rs
EPashkin: You right now I can 😉

src/style_context.rs
```diff
@@ -0,0 +1,21 @@
...
+use StateFlags;
+use StyleContext;
+use StyleContextExt;
+
+pub trait StyleContextExtManual: 'static {
+    fn get_font(&self, state: StateFlags) -> Option<FontDescription>;
+}
+
+impl<O: IsA<StyleContext>> StyleContextExtManual for O {
+    fn get_font(&self, state: StateFlags) -> Option<FontDescription> {
```
sdroege: Can this ever be `None`?

EPashkin: In normal case is not, want me add one more `except`?

sdroege: If it's a bug if there is no font description then yes

EPashkin: Old https://developer.gnome.org/gtk3/stable/GtkStyleContext.html#gtk-style-context-get-font don't says that it can be null

sdroege:
`gir-files` PR is merged, so this can be updated and then also be merged :) @EPashkin

EPashkin:
Done

GuillaumeGomez:
Once CI pass, I merge then!

sdroege:
@GuillaumeGomez green :)

GuillaumeGomez:
Thanks @EPashkin !

PR #959, opened 2020-02-21
Issue #862 - Add size_allocate vfunc to gtk::Widget

hfiguiere:
(patch is against 0.8 but should go into master too)

### DIFF
--- a/src/subclass/widget.rs
+++ b/src/subclass/widget.rs
@@ -15,6 +15,7 @@ use crate::SelectionData;
 use crate::TextDirection;
 use cairo;
 use cairo_sys;
+use Allocation;
 use SizeRequestMode;
 use Widget;
 use WidgetClass;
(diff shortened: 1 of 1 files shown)


hfiguiere:
Uh sorry I forgot a change by mistake. It's now in.

EPashkin:
Is not better add it against master branch and then mass cherry-pick it on next minor release ?

sdroege:
Yes, please update this to the master branch. Everything should first go to master :)

src/subclass/widget.rs
```diff
@@ -209,6 +210,10 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl + 'static {
     fn get_preferred_height_for_width(&self, widget: &Widget, width: i32) -> (i32, i32) {
         self.parent_get_preferred_height_for_width(widget, width)
     }
+
+    fn size_allocate(&self, widget: &Widget, allocation: &mut Allocation) {
```
sdroege: Should this be a mutable reference? In GTK it's all const. AFAIU this gives the allocation to the subclass, not returns the allocation from the subclass.

Also the `gtk_widget_size_allocate()` function should probably be exposed at the same time.

hfiguiere: it is exposed as "gtk::WidgetExt::size_allocate". or am I missing something?

hfiguiere:
Replaced by PR #961

src/subclass/widget.rs
hfiguiere: > Should this be a mutable reference? In GTK it's all const. AFAIU this gives the allocation to the subclass, not returns the allocation from the subclass.

See PR #961

sdroege: Indeed

PR #961, opened 2020-02-21
Issue #862 - Add size_allocate vfunc to gtk::Widget

hfiguiere:
Replaces PR #959 (on master this time)

### DIFF
--- a/src/subclass/widget.rs
+++ b/src/subclass/widget.rs
@@ -15,6 +15,7 @@ use crate::SelectionData;
 use crate::TextDirection;
 use cairo;
 use cairo_sys;
+use Allocation;
 use SizeRequestMode;
 use Widget;
 use WidgetClass;
@@ -209,6 +210,10 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl + 'static {
     fn get_preferred_height_for_width(&self, widget: &Widget, width: i32) -> (i32, i32) {
         self.parent_get_preferred_height_for_width(widget, width)
     }
+
+    fn size_allocate(&self, widget: &Widget, allocation: &Allocation) {
+        self.parent_size_allocate(widget, allocation)
+    }
 }
 
 pub trait WidgetImplExt {
@@ -298,6 +303,7 @@ pub trait WidgetImplExt {
     fn parent_get_preferred_width_for_height(&self, widget: &Widget, height: i32) -> (i32, i32);
     fn parent_get_preferred_height(&self, widget: &Widget) -> (i32, i32);
     fn parent_get_preferred_height_for_width(&self, widget: &Widget, width: i32) -> (i32, i32);
+    fn parent_size_allocate(&self, widget: &Widget, allocation: &Allocation);
 }
 
 impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
@@ -794,6 +800,20 @@ impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
             (minimum_size.assume_init(), natural_size.assume_init())
         }
     }
+
+    fn parent_size_allocate(&self, widget: &Widget, allocation: &Allocation) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkWidgetClass;
+            let f = (*parent_class)
+                .size_allocate
+                .expect("No parent class impl for \"size_allocate\"");
+            f(
+                widget.to_glib_none().0,
+                mut_override(allocation.to_glib_none().0),
+            );
+        }
+    }
 }
 
 unsafe impl<T: ObjectSubclass + WidgetImpl> IsSubclassable<T> for WidgetClass {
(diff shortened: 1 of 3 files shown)


src/subclass/widget.rs
```diff
@@ -209,6 +210,10 @@ pub trait WidgetImpl: WidgetImplExt + ObjectImpl + 'static {
     fn get_preferred_height_for_width(&self, widget: &Widget, width: i32) -> (i32, i32) {
         self.parent_get_preferred_height_for_width(widget, width)
     }
+
+    fn size_allocate(&self, widget: &Widget, allocation: &Allocation) {
```
sdroege: The function `gtk_widget_size_allocate()` says in the docs:

> In this function, the allocation may be adjusted. It will be forced to a 1x1 minimum size, and the adjust_size_allocation virtual method on the child will be used to adjust the allocation. Standard adjustments include removing the widget’s margins, and applying the widget’s Widget:halign and Widget:valign properties.

So maybe this needs to be mutable after all and subclasses need to be able to modify it?

sdroege: Note that I have no idea how this function is supposed to work, I'm just guessing from the docs :)

src/subclass/widget.rs
```diff
@@ -1349,3 +1368,17 @@ unsafe extern "C" fn widget_get_preferred_height_for_width<T: ObjectSubclass>(
...
+unsafe extern "C" fn widget_size_allocate<T: ObjectSubclass>(
+    ptr: *mut gtk_sys::GtkWidget,
+    allocation: *mut gtk_sys::GtkAllocation,
+) where
+    T: WidgetImpl,
+{
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Widget = from_glib_borrow(ptr);
+    let allocate: Allocation = from_glib_none(allocation);
```
sdroege: If the allocation should be possible to modify then you need to write it back afterwards.

src/subclass/widget.rs
hfiguiere: That was my initial guess (see previous PR).

Looking at https://gitlab.gnome.org/GNOME/gtk/blob/gtk-3-24/gtk/gtkwidget.c#L6443
The GtkWidget vfunc doesn't modify it.

sdroege: Does any of the subclasses? Maybe worth asking on #gtk :)

hfiguiere: From what I can see they don't. At best they call gtk_widget_set_allocation or the parent class size allocate vfunc.

sdroege: Would be useful to get a definite answer from a GTK developer :)

sdroege: @ebassi maybe? :)

src/subclass/widget.rs
ebassi: The allocation is not an in-out parameter; you're supposed to call `gtk_widge_set_allocation()` or chain up to the parent class's allocation, in which case you can tweak the argument for those.

sdroege: Thanks, so the allocation should actually be an immutable reference. And new allocations could be passed to childs, etc.

sdroege:
So as part of this `WidgetExt::size_allocate()` should also be changed to not take a mutable reference.

EPashkin:
@ebassi Thanks for the clarification.
@hfiguiere Thanks for PR
I restarted failed job in CI
👍

hfiguiere:
> So as part of this `WidgetExt::size_allocate()` should also be changed to not take a mutable reference.

This is auto generated.

sdroege:
That's true, but it can be fixed like here https://github.com/gtk-rs/gtk/blob/2fcc48623a37a1f6899d56a1ce46d18d39e43e8e/Gir.toml#L680-L684 :)

hfiguiere:
OK. Added it as a separate commit as this one we might want to skip cherry-picking to to 0.8

sdroege:
> OK. Added it as a separate commit as this one we might want to skip cherry-picking to to 0.8

Both can be backported :) Changing from `&mut` to `&` is not a breaking change.

src/subclass/widget.rs
```diff
@@ -794,6 +800,18 @@ impl<T: WidgetImpl + ObjectImpl> WidgetImplExt for T {
             (minimum_size.assume_init(), natural_size.assume_init())
         }
     }
+
+    fn parent_size_allocate(&self, widget: &Widget, allocation: &Allocation) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkWidgetClass;
+            let f = (*parent_class).size_allocate.unwrap();
```
sdroege:
```suggestion
            let f = (*parent_class).size_allocate.expect("No parent class impl for \"size_allocate\"");
```

sdroege:
Otherwise looks good to me, thanks!

sdroege:
@GuillaumeGomez please merge :)

GuillaumeGomez:
Thanks!

PR #963, opened 2020-02-24
Use cache to speed up builds

GuillaumeGomez:
The first commit will likely be slower but let's see after this one. :3

cc @EPashkin @sdroege

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,21 +9,37 @@ rust:
   - stable
   - 1.39.0
 env:
-  - GTK=3.14 FEATURES=
-  - GTK=3.24 FEATURES=v3_16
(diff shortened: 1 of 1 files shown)


EPashkin:
May be restart one of finished job to check how cache works?

EPashkin:
Hm,  restarted job stalled on initializing cache.

sdroege:
See what I wrote on IRC:

```
<slomo_> imperio: it can be easily slower in the future if you don't clean the cache often enough
<slomo_> imperio: the cache will become too huge :)
```

GuillaumeGomez:
Might not be worth it then... Closing!

PR #975, opened 2020-03-13
add basic subclassing functionality for treeview

vhdirk:
This adds the possibility to subclass TreeView. It does not, however, enable overriding subclasses since:
1. I can't figure out which ones I would need to implement. The generated bindings seem very illogical in that respect.
2. Most of what these methods do can still be done by connecting to signals.

Would it make sense to provide 'stub' subclassing Impls for all other widgets as well?

### DIFF
--- a/src/subclass/mod.rs
+++ b/src/subclass/mod.rs
@@ -23,6 +23,7 @@ pub mod fixed;
 pub mod header_bar;
 pub mod icon_view;
 pub mod stack;
+pub mod tree_view;
 pub mod widget;
 pub mod window;
 
@@ -48,6 +49,7 @@ pub mod prelude {
     pub use super::header_bar::HeaderBarImpl;
     pub use super::icon_view::{IconViewImpl, IconViewImplExt};
     pub use super::stack::StackImpl;
+    pub use super::tree_view::TreeViewImpl;
     pub use super::widget::{WidgetImpl, WidgetImplExt};
     pub use super::window::{WindowImpl, WindowImplExt};
     pub use gio::subclass::prelude::*;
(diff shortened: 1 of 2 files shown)


sdroege:
>     1. I can't figure out which ones I would need to implement. The generated bindings seem very illogical in that respect.

What do you mean with the generated bindings being illogical? 

> Would it make sense to provide 'stub' subclassing Impls for all other widgets as well?

Yes, over time we should start adding the missing vfuncs though.

@GuillaumeGomez Good to be merged once the CI is happy :)

GuillaumeGomez:
```
error[E0599]: no method named `parent_startup` found for reference `&SimpleApplicationPrivate` in the current scope
   --> src\bin\basic_subclass.rs:187:14
    |
187 |         self.parent_startup(app);
    |              ^^^^^^^^^^^^^^ method not found in `&SimpleApplicationPrivate`
    |
    = help: items from traits can only be used if the trait is in scope
help: the following trait is implemented but not in scope; perhaps add a `use` for it:
    |
15  | use gio::subclass::application::ApplicationImplExt;
    |
```

EPashkin:
It must be fixed by gtk-rs/gio#284, I restarted CI

sdroege:
@GuillaumeGomez good to be merged, only the macos timeout failure as usual.

GuillaumeGomez:
I'll need to investigate this very annoying failure at some point...

EPashkin:
@GuillaumeGomez It just too long, IMHO we need to minimize number of examples for CI, at minimum for macos.

GuillaumeGomez:
Maybe, to be investigated...

PR #978, opened 2020-03-24
Add scroll_event vfunc to WidgetImpl

tsahyt:
Another vfunc. It seems to work, although my test application had an issue where the event only gets triggered when a mouse button is held down at the same time. I suspect this is for a different reason though, e.g. some faulty window setup. Implementation is much the same as for `motion_notify_event`.


sdroege:
Failed CI are unrelated to this.

GuillaumeGomez:
Seems like CI is on fire lately.

Thanks @tsahyt !

PR #980, opened 2020-04-05
Update for new from_glib_borrow signature

sdroege:
See gtk-rs/glib#605


src/text_buffer.rs
```diff
@@ -15,18 +15,21 @@ use TextBuffer;
 use TextIter;
 
 pub trait TextBufferExtManual: 'static {
-    fn connect_insert_text<F: Fn(&Self, &mut TextIter, &str) + 'static>(
+    fn connect_insert_text<F: Fn(&Self, TextIter, &str) -> TextIter + 'static>(
```
sdroege, line 18: This one here was completely broken before. We should also update the other `TextIter` related signal handlers and actually have examples for them. But that's for a separate PR, I'll create an issue once this is merged.

Issue #960, opened 2020-02-21
set_sort_func for TreeModelSort panics


cdepillabout:
I'm trying to use [`gtk::TreeSortableExtManual::set_sort_func`](https://gtk-rs.org/docs/gtk/prelude/trait.TreeSortableExtManual.html#tymethod.set_sort_func) on a `TreeModelSort`, but the callback appears to always panic at runtime.

I don't have a minimally reproducible error, but I do have a small change to my own code base that triggers this error:

https://github.com/cdepillabout/nix-query-tree-viewer/commit/fa5b6ed7557bd9b4c20dd9fafac8207a0f6b678b#diff-781f3944d3debcae5a68e2d54ca2f098R192

In my actual code base, I've worked around this by just coding up my own wrapper to `gtk_sys::gtk_tree_sortable_set_sort_func` based on the code for [`gtk::ListBoxExt::set_sort_func`](https://gtk-rs.org/docs/gtk/trait.ListBoxExt.html#tymethod.set_sort_func) (which I had assumed was working correctly, because it is automatically generated):

https://github.com/cdepillabout/nix-query-tree-viewer/blob/d93a85929704b03deebb96c25dd3926aa50c0ccd/src/ui/stack/tree.rs#L40

Here is where I call it from:

https://github.com/cdepillabout/nix-query-tree-viewer/blob/d93a85929704b03deebb96c25dd3926aa50c0ccd/src/ui/stack/tree.rs#L188

--------------------------------------------

Here's the backtrace when this error happens:

```
$ RUST_BACKTRACE=1 cargo run -- /nix/store/jymg0kanmlgbcv35wxd8d660rw0fawhv-hello-2.10.drv
...
thread 'main' panicked at 'assertion failed: self.is::<T>()', /home/illabout/.cargo/registry/src/github.com-1ecc6299db9ec823/glib-0.9.1/src/object.rs:358:9
stack backtrace:
   0: backtrace::backtrace::libunwind::trace
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/libunwind.rs:88
   1: backtrace::backtrace::trace_unsynchronized
             at /cargo/registry/src/github.com-1ecc6299db9ec823/backtrace-0.3.40/src/backtrace/mod.rs:66
   2: std::sys_common::backtrace::_print_fmt
             at src/libstd/sys_common/backtrace.rs:77
[...]

PR #982, opened 2020-04-08
The TreeSortable functions are called on the underlying model

sdroege:
As such, they are most likely not of type TreeSortable. The only thing
we know about them is that they are some kind of TreeModel, so use that
type instead of panicking.

Also use from_glib_borrow() in the trampolines.

Fixes https://github.com/gtk-rs/gtk/issues/960

----

CC @BenjaminRi @EPashkin @GuillaumeGomez


sdroege:
@GuillaumeGomez Green CI (even macos passed this time, suspicious!)

GuillaumeGomez:
Well, dark magic isn't always bad. :3

PR #987, opened 2020-04-13
Fix transmute mess

GuillaumeGomez:
Part of gtk-rs/gir#902

cc @EPashkin @sdroege


src/signal.rs
```diff
@@ -94,7 +95,7 @@ mod editable {
                 connect_raw(
                     self.to_glib_none().0 as *mut _,
                     b"changed\0".as_ptr() as *mut _,
-                    Some(*(&trampoline::<Self, F> as *const _ as *const _)),
+                    Some(transmute(trampoline::<Self, F> as usize)),
```
sdroege, line 98: Please put the correct types in all the manual code :) `transmute::<_, unsafe extern "C" fn ...>(trampoline as *const ())` etc

GuillaumeGomez: Lazyness overload. :3

sdroege: This can be fixed with sed btw, no need to do that manually

sdroege:
@GuillaumeGomez All green except for the usual macos

PR #992, opened 2020-04-26
Focal

EPashkin:
Part of https://github.com/gtk-rs/gir-files/pull/52

cc @GuillaumeGomez , @sdroege


EPashkin:
Updated to https://github.com/gtk-rs/gir-files/pull/55
Contains two commits, changes before focal from git update and focal changes

Gir.toml
```diff
@@ -2064,6 +2060,9 @@ status = "generate"
     name = "get_model"
         [object.function.return]
         nullable = false
+    [[object.function]]
+    name = "new_with_model"
+    rename = "new"
```
sdroege, line 2065: Why? There's also a normal `new()` that's useful by itself and works differently

sdroege: Oh there is no `new()` and because that's how we had it in manual code before. Why?

EPashkin: Before this version `gtk_tree_model_sort_new_with_model` was function in gtk_tree, so we added call for it in `TreeModelSort` for better usability, now it moved to `TreeModelSort` in upstream, so we just rename it in Rust part

sdroege: Ok :man_shrugging:

EPashkin:
Updated

sdroege:
Looks good to me, let's merge after sys is done :)

PR #993, opened 2020-04-30
Clippy fixes

kinnison:
Fix a variety of clippy issues, some by allowing the lints, some by fixing the issues.

I hope this is of use.


sdroege:
Let's also make clippy warnings fatal in the CI then (and in glib/gio at least, the others need fixing for various clippy warnings first)? :) @GuillaumeGomez ?

src/subclass/widget.rs
```diff
@@ -1490,6 +1490,12 @@ unsafe extern "C" fn widget_size_allocate<T: ObjectSubclass>(
     imp.size_allocate(&wrap, allocate);
 }
 
+/// Realize the given widget
+///
+/// # Safety
+///
+/// If the widget is not yet in a toplevel window then according to GTK,
+/// "bad things will happen".
 pub unsafe extern "C" fn widget_realize<T: ObjectSubclass>(ptr: *mut gtk_sys::GtkWidget)
```
sdroege, line 1499: All these functions shouldn't be `pub` at all. Thanks for noticing :)

kinnison: Aah, should I remove the safety notes and unpub them?

src/page_range.rs
```diff
@@ -52,7 +52,7 @@ impl FromGlibContainerAsVec<gtk_sys::GtkPageRange, *mut gtk_sys::GtkPageRange> f
 
         let mut res = Vec::with_capacity(num);
         for i in 0..num {
-            res.push(PageRange(ptr::read(ptr.offset(i as isize))));
+            res.push(PageRange(ptr::read(ptr.add(i))));
```
GuillaumeGomez, line 55: I didn't know this method. :o

src/rt.rs
```diff
@@ -65,6 +65,12 @@ pub fn is_initialized_main_thread() -> bool {
 }
 
 /// Informs this crate that GTK has been initialized and the current thread is the main one.
+///
+/// # Safety
+///
+/// You must only call this if you have initialised GTK yourself, on the thread
+/// with which you are calling this function, and that this thread is the main thread
```
GuillaumeGomez, line 72: The reading is a bit complicated here...

kinnison: I wasn't sure if this even ought to be public -- I can expand this into bullets instead of a compound sentence if it does need to remain public, otherwise we could unpub it.

src/page_range.rs
kinnison: The violated clippy lint even points you at it :D

src/subclass/widget.rs
sdroege: Yes please

src/rt.rs
sdroege: Should be public, yes

kinnison: I've pushed replacement docs for this.

sdroege:
lgtm. @kinnison can you add the magic to make clippy warnings errors to the `.travis.yml` where it already calls clippy currently?

kinnison:
@sdroege Yes, but only when travis comes back to tell me I've not fluffed anything else up :D

GuillaumeGomez:
Let's merge once the clippy warnings are denied then. :)

kinnison:
I got bored waiting for travis so I pushed what ought to do the trick.

src/lib.rs
```diff
@@ -148,11 +148,14 @@
 //! them once. **Omitting them in the following cargo invocations will not undo
 //! their effects!**
 
-#![cfg_attr(feature = "cargo-clippy", allow(let_unit_value))]
-#![cfg_attr(feature = "cargo-clippy", allow(new_without_default))]
-#![cfg_attr(feature = "cargo-clippy", allow(type_complexity))]
-#![cfg_attr(feature = "cargo-clippy", allow(trivially_copy_pass_by_ref))]
-#![cfg_attr(feature = "cargo-clippy", allow(derive_hash_xor_eq))]
+#![deny(clippy::all)]
```
EPashkin, line 151: IMHO better do this on travis level
```
cargo clippy -- -D warnings
```

GuillaumeGomez: :+1:

kinnison: My thought here was that people would probably miss it if it were CI only, but if you want I can shift it.  @GuillaumeGomez @sdroege ?

GuillaumeGomez: If it's CI only, it's fine so it means it won't break in case a new lint is added. I personally think this is the best solution. And we won't miss it since releases have to go through CI.

kinnison: OK, shifted.

GuillaumeGomez:
The new failure is because there is already documentation in our repository. I think the best way for the `TreeRowReference` method is to add it into the repository (sorry about that...).

kinnison:
@GuillaumeGomez if we move the safety information to the lgpl-docs repo then we also need to ensure we run cargo clippy with the embed-lgpl-docs feature set, I'll try that locally and see if it works

GuillaumeGomez:
Thanks!

PR #994, opened 2020-05-01
Add gtk::Socket and gtk::Plug subclassing support

kinnison:
The Socket class only has two additional signals over a container,
so it's a pretty trivial subclass.

Signed-off-by: Daniel Silverstone <dsilvers@digital-scurf.org>

### DIFF
--- a/src/subclass/mod.rs
+++ b/src/subclass/mod.rs
@@ -22,6 +22,10 @@ pub mod event_box;
 pub mod fixed;
 pub mod header_bar;
 pub mod icon_view;
+#[cfg(any(gdk_backend = "x11", feature = "dox"))]
+pub mod plug;
+#[cfg(any(gdk_backend = "x11", feature = "dox"))]
+pub mod socket;
 pub mod stack;
 pub mod tree_view;
 pub mod widget;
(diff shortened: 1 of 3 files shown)


kinnison:
I *believe* there's no reason this shouldn't be safe, but it's my first ever subclass and all I did was make sure things looked sane and didn't have compile errors.

kinnison:
Aah, I need to detect if X11 is present in gtk for this to be cross-platform safe.  Hmm.  Any suggestions?

sdroege:
> Aah, I need to detect if X11 is present in gtk for this to be cross-platform safe. Hmm. Any suggestions?

See https://github.com/gtk-rs/gtk/blob/master/src/auto/mod.rs#L1334 . Which is of course also very wrong: macOS and Windows also have X11 implementations and not every Unix has one. I think this needs a custom X11 feature flag.

kinnison:
I suppose nominally I ought to write the GtkPlug subclass support too, shall I do that, similarly guarded, and then we could merge those while we think about a better way to guard the X11 stuff?

kinnison:
Depending on the opinion of people on https://github.com/gtk-rs/gtk/pull/995 we may want to hold off merging this until a better config label can be used.

sdroege:
@GuillaumeGomez CI is green

PR #995, opened 2020-05-01
gdk_backend config flags

kinnison:
This is a proposed approach for better handling of targets.  The pkg-config file defines a targets variable which is whitespace separated.  This PR consumes that in `build.rs` and sets `target="x11"` and friends as config values.

To demonstrate its use, this then switches `gtk::Plug` and `gtk::Socket` to be predicated on `target="x11"` rather than `all(not(windows), not(target_os = "macos"))`

@sdroege What do you think of this as an approach?  If considered good, it should probably be spread to `gtk-sys` and possibly other crates.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -1535,7 +1535,7 @@ status = "generate"
 [[object]]
 name = "Gtk.Plug"
 status = "generate"
-cfg_condition = "all(not(windows),not(target_os = \"macos\"))"
+cfg_condition = "gdk_backend = \"x11\""
 
 [[object]]
 name = "Gtk.Popover"
@@ -1748,7 +1748,7 @@ version = "3.20"
 [[object]]
 name = "Gtk.Socket"
 status = "generate"
-cfg_condition = "all(not(windows),not(target_os = \"macos\"))"
+cfg_condition = "gdk_backend = \"x11\""
 trait_name = "GtkSocketExt"
 
 [[object]]
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -41,6 +41,9 @@ dox = ["gdk/dox", "gtk-sys/dox"]
 [package.metadata.docs.rs]
 features = ["dox", "embed-lgpl-docs"]
 
+[build-dependencies]
+pkg-config = "0.3.7"
+
 [target.'cfg(target_os = "macos")'.build-dependencies]
 cc = "^1.0"
 
--- a/build.rs
+++ b/build.rs
@@ -5,6 +5,7 @@ fn main() {
     manage_docs();
     #[cfg(target_os = "macos")]
     build_foreground();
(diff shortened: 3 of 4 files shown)


Gir.toml
```diff
@@ -1535,7 +1535,7 @@ status = "generate"
 [[object]]
 name = "Gtk.Plug"
 status = "generate"
-cfg_condition = "all(not(windows),not(target_os = \"macos\"))"
+cfg_condition = "target = \"x11\""
```
sdroege, line 1538: I didn't know you could define your own targets :open_mouth: 

We should also do the same with `gio-unix` and things like that I guess.

sdroege:
Looks like a good solution to me, only makes it more confusing for users if they want to use the API but it does not exist because of their GTK version. With a feature flag we could print some more useful error if the user tries to enable it but it does not exist.

kinnison:
I've changed the config to `gdk_backend="x11"` and friends, since that seems more descriptive.  I've also added explanatory notes in build.rs

Gir.toml
kinnison: Not sure what you mean re `gio-unix` since this is mostly a gdk/gtk thing.

sdroege: There's a similar problem with GIO (and GLib) about Windows/POSIX/Linux specific APIs.

kinnison: Aah, I don't think gio exports such info in their pkg-config but perhaps there's a different way to discover it for that.

sdroege: Separate `pkg-config` files

EPashkin:
Is there no way to add it to gdk-sys?

kinnison:
@EPashkin Yes, the same work ought to work for gdk-sys if you want to use the capability there for any reason.

EPashkin:
I want detect flags in gdk-sys and use it here

kinnison:
@EPashkin Aaah, right, to my knowledge it's not possible for crates to acquire configuration flags from their dependencies.  If you know of a way to do this, I'm happy to rework it.

EPashkin:
I thought about something like "DEP_GDK_SYS_BACKEND" from https://doc.rust-lang.org/cargo/reference/build-scripts.html#the-links-manifest-key
It should works if gtk depends of gdk-sys directly and need "build.rs" in "gtk" too.

kinnison:
Oh that's interesting, I shall look at doing that because yes, that makes a lot of sense.  I think gdk-sys should compute them gtk-sys should propagate them, and gtk can then consume them.  Hmm...

kinnison:
I've pushed the first part of that new variant here - https://github.com/gtk-rs/sys/pull/167
Once that merges, I'll pop a PR up for replacing this calculating with inheriting it from gtk-sys

PR #996, opened 2020-05-02
Application initialisation

### LINKED ISSUE #984: gtk::Application  created by gtk::ApplicationBuilder doesn't initialize GTK
Here is a repository which reproduction of an error. 
https://github.com/andy128k/reproduce_error_gtk_rs_application_builder

kinnison:
With builder postprocessing support in https://github.com/gtk-rs/gir/pull/912 merged, this fixes #984


src/auto/about_dialog.rs
```diff
@@ -383,10 +383,11 @@ impl AboutDialogBuilder {
...
             properties.push(("width-request", width_request));
         }
-        glib::Object::new(AboutDialog::static_type(), &properties)
+        let ret = glib::Object::new(AboutDialog::static_type(), &properties)
             .expect("object new")
-            .downcast()
-            .expect("downcast")
+            .downcast::<AboutDialog>()
+            .expect("downcast");
+        ret
```
sdroege, line 390: Poor clippy :)

sdroege:
Looks good to me otherwise

src/auto/about_dialog.rs
kinnison: Yeah clippy will get sad, but also not, because we already override that one I think.

kinnison: I'll rebase onto current master to make that clearer though.

kinnison:
I've pushed a rebase onto current master to make it clearer that clippy should be perfectly happy with those let/return pairs :disappointed:

sdroege:
The gir checker thingy is failing

kinnison:
@sdroege I've added a `skip_assert_initialized!();` even though I think the checker ought to ignore internal (`pub(crate)`) functions but I couldn't work out the perl inside of 30s so I decided to do the cheap/lazy fix.

PR #1005, opened 2020-05-08
WIP: test metadeps sys branch

gdesmott:
Not meant to be merged, I just want to have the CI running.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -74,3 +74,12 @@ pango = { git = "https://github.com/gtk-rs/pango" }
 
 [dev-dependencies]
 gir-format-check = "^0.1"
+
+[patch.'https://github.com/gtk-rs/sys']
+gdk-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+gdk-pixbuf-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+gio-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+glib-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+gobject-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+gtk-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }
+pango-sys = { git = "https://github.com/gdesmott/sys.git", branch = "wip-metadeps" }


gdesmott:
@sdroege @GuillaumeGomez : this was a build test using [my metadeps sys branch](https://github.com/gdesmott/sys/commit/d223e92396fa86f60ea48050fc675012694a41ad).

To my surprise all tests passed so I'm wondering what's the actual use case for all the current `build.rs` hacks: `GTK_INCLUDE_DIR`, `GTK_LIB_DIR`, windows specific paths, etc?

sdroege:
Maybe they're not even required anymore and are just cargo culted for future generations. @EPashkin would know.

But as I said on IRC, we should really get rid of those hacks and instead set up the Windows CI environment correctly. Which might've happened by now (maybe msys2 fixed their packages...).

EPashkin:
I tested on Windows with last msys2x64 and seems `hardcode_shared_libs` branches in good pkgconfig case now not needed (I not remember why it was added).
But IMHO we still better have GTK_LIB_DIR and fallback.
Not sure about `GTK_INCLUDE_DIR`

sdroege:
> But IMHO we still better have GTK_LIB_DIR and fallback.
> Not sure about `GTK_INCLUDE_DIR`

The fallback is provided by metadeps in this branch, which allows to override things in a generic way for each dependency so that people can continue to work around their broken environments.

EPashkin:
> The fallback is provided by metadeps in this branch, which allows to override things in a generic way for each dependency so that people can continue to work around their broken environments.

I don't see fallback not in sys branch not in https://github.com/gdesmott/metadeps/blob/master/src/lib.rs

gdesmott:
It's all in my [wip branch](https://github.com/gdesmott/metadeps/blob/wip/src/lib.rs) for now.

EPashkin:
I missed that it have branch, but IMHO it still don't have "similar" failure case  https://github.com/gdesmott/metadeps/blob/wip/src/lib.rs#L310

gdesmott:
What do you mean? This feature is meant for crates building internally the underlying C library to statically link to it. That's not something we currently support in `gtk-rs`.

EPashkin:
Yes, that why I say that you sys'es needed special fallback for this errors.

gdesmott:
How so? metadeps will [raise an error](https://github.com/gdesmott/metadeps/blob/wip/src/lib.rs#L343) if user tries to use this feature as our sys don't implement a build closure.

EPashkin:
> 
> 
> How so? metadeps will [raise an error](https://github.com/gdesmott/metadeps/blob/wip/src/lib.rs#L343) if user tries to use this feature as our sys don't implement a build closure.

And pkgconfig error is lost and we can't find it by metadeps error.

gdesmott:
But that only happens if user explicitly defined `METADEPS_GTK_BUILD_INTERNAL=auto` or `always`. The default value is `never` so `metadeps` will fail on pkg-config and report its error.

EPashkin:
Sorry, I seems totally go to wrong direction.
So we just need don't forget to add it in your sys'es.

sdroege:
>  So we just need don't forget to add it in your sys'es.

Add what?

EPashkin:
https://github.com/gtk-rs/sys/blob/master/gtk-sys/build.rs#L100-L106

sdroege:
That's something for metadeps to take care of. The goal here is to move away from all these custom hacks that work differently for every crate.

Also the above part you linked shouldn't be done unconditionally as that regularly causes problems for users, especially on Windows. It needs to be something to opt-in for the user, which is already possible with metadeps by setting the correct libraries for the corresponding crate.

sdroege:
This was merged

PR #1007, opened 2020-05-14
Don't include LGPL docs in the docs if both embed-lgpl-docs and purge…

sdroege:
…-lgpl-docs features are selected

This makes usage of RLS/rust-analyzer on the repository much faster and
less annoying as the docs don't have to be included and removed on every
change.

----
CC @EPashkin @GuillaumeGomez

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -8,7 +8,10 @@ fn main() {
     check_features();
 }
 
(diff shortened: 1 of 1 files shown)


GuillaumeGomez:
Thanks!

PR #1009, opened 2020-05-25
Let Widget::hide_on_delete return Inhibit

### LINKED ISSUE #1008: WidgetExt::hide_on_delete useless in current form
```rust
fn hide_on_delete(&self) -> bool
```

should be connectable to

```rust
fn connect_delete_event<F: Fn(&Self, &Event) -> Inhibit + 'static>(
    &self,
    f: F
) -> SignalHandlerId
```

which doesn't work out that well.

- https://gtk-rs.org/docs/gtk/trait.WidgetExt.html#tymethod.hide_on_delete
- https://gtk-rs.org/docs/gtk/trait.WidgetExt.html#tymethod.connect_delete_event

sophie-h:
Currently it's not possible to pass this directly to `connect_delete_event()`, because `hide_on_delete()` only accepts one argument. But that also seems to be the case for the C API. That's why I left it like that.

https://developer.gnome.org/gtk3/stable/GtkWidget.html#GtkWidget-delete-event

Closes #1008

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -2349,6 +2349,10 @@ manual_traits = ["WidgetExtManual"]
         [[object.function.parameter]]
         name = "allocation"
         const = true
+    [[object.function]]
+    name = "hide_on_delete"
+    ignore = true
+    doc_trait_name = "WidgetExtManual"
     [[object.signal]]
     name = "button-press-event"
     inhibit = true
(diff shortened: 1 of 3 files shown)


Gir.toml
```diff
@@ -2349,6 +2349,10 @@ manual_traits = ["WidgetExtManual"]
         [[object.function.parameter]]
         name = "allocation"
         const = true
+    [[object.function]]
+    name = "hide_on_delete"
+    ignore = true
```
GuillaumeGomez, line 2354: Please add a comment saying why the function is ignored (we don't do it enough...).

sdroege:
@GuillaumeGomez All green

PR #1018, opened 2020-06-15
Update to new gir

sdroege:


sdroege:
@GuillaumeGomez This is as green as it gets until we merged all the PRs.

PR #1019, opened 2020-06-16
generate missing builders

bilelmoussaoui:
Generates the following builders:
- Gtk.AdjustmentBuilder
- Gtk.CellRendererBuilder
- Gtk.IconViewBuilder
- Gtk.ShortcutLabelBuilder
- Gtk.ShortcutsWindowBuilder

There are still some builders to be generated, but I think they don't make sense
- `Gtk.WidgetBuilder` base class
- `Gtk.ArrowBuilder` deprecated 
- `Gtk.MiscBuilder` base class
- `Gtk.ContainerBuilder` base class
- `Gtk.BuilderBuilder` I'm not sure about this one?
- `Gtk.CellAreaBuilder` base class
- `Gtk.CellAreaContextBuilder` base class
- `Gtk.EventControllerBuilder` base class
- `Gtk.IMContextBuilder` no idea
- `Gtk.NativeDialogBuilder` I don't think anyone creates a NativeDialog other than a Gtk.NativeFileChooser
- `Gtk.GestureBuilder` base class
- `Gtk.GestureSingleBuilder` base class
- `Gtk.MenuShellBuilder` base class
- `Gtk.SettingsBuilder` I'm not sure about this one, most of the time you use the default settings, but it could be useful if you're writing lower level stuff with Rust & GTK
- `Gtk.RangeBuilder` base class
- `Gtk.TreeModelFilterBuilder` no idea
- `Gtk.TreeModelSortBuilder` no idea
- `Gtk.TreeSelectionBuilder` no idea

PR #1021, opened 2020-06-16
Ignore some builders

GuillaumeGomez:
It follows https://github.com/gtk-rs/gtk/pull/1019. The point being they stop showing up when running `gir -m not_bound`.

Although, there is a little issue on gir side which completely ignores the ignored builders. I'll send a fix there.


GuillaumeGomez:
I should sort the ignores though...

GuillaumeGomez:
The ignore will be taken into account when https://github.com/gtk-rs/gir/pull/933 will be merged. I'll update this PR tomorrow.

Gir.toml
```diff
@@ -356,6 +356,28 @@ builders = [
 
 ignore = [
     "Gtk.Activatable",
+    # base class
+    "Gtk.WidgetBuilder",
+    # deprecated
+    "Gtk.ArrowBuilder",
+    # base class
+    "Gtk.MiscBuilder",
+    # base class
```
sdroege, line 365: We can easily detect this case

```xml
    <class name="Misc"
           c:symbol-prefix="misc"
           c:type="GtkMisc"
           parent="Widget"
           abstract="1"
           glib:type-name="GtkMisc"
           glib:get-type="gtk_misc_get_type"
           glib:type-struct="MiscClass">
```

Note the `abstract=1` here. That could be parsed by gir

Gir.toml
```diff
@@ -356,6 +356,28 @@ builders = [
 
 ignore = [
     "Gtk.Activatable",
+    # base class
+    "Gtk.WidgetBuilder",
+    # deprecated
```
sdroege, line 361: It's also an abstract base class

Gir.toml
GuillaumeGomez: Oh interesting! I'll add it in gir then.

GuillaumeGomez:
The removed builders seem to be a soundness bug we had before. They're not supposed to exist.

PR #1023, opened 2020-06-16
ignore all the deprecated objects

bilelmoussaoui:
I have been running `gir -c ./Gir.toml -d gir-files/ -m not_bound | grep "\[NOT GENERATED\]"` a lot today, tracking down missing stuff. This merge requests ignores every deprecated object before 3.14.

### DIFF
--- a/Gir.toml
+++ b/Gir.toml
@@ -547,6 +547,11 @@ manual_traits = ["AccelGroupExtManual"]
     ignore = true
 
 [[object]]
+name = "Gtk.Action"
+#deprecated since 3.10
+status = "ignore"
+
+[[object]]
 name = "Gtk.Actionable"
 status = "generate"
     [[object.property]]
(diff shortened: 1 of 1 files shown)


sdroege:
Shouldn't we simply ignore them in `gir` directly if their deprecated version is before the minimum version and it is configured to remove deprecated things?

GuillaumeGomez:
I agree with @sdroege's suggestion.

bilelmoussaoui:
> Shouldn't we simply ignore them in `gir` directly if their deprecated version is before the minimum version and it is configured to remove deprecated things?

Well, sometimes the object (for example Gtk.VScrollbar) is only mentioned as deprecated on the object doc

bilelmoussaoui:
Makes sense, I will keep this open until there's such an option on gir and update this one accordingly

PR #1024, opened 2020-06-16
Generate GtkGestureStylus

bilelmoussaoui:
In the continuation of my previous merge requests. 

I was wondering about two enums which are deprecated in 3.20, should be generated? 
- `Gtk.ArrowPlacement`
- `Gtk.ToolbarSpaceStyle`


sdroege:
> I was wondering about two enums which are deprecated in 3.20, should be generated?
> 
>     * `Gtk.ArrowPlacement`
> 
>     * `Gtk.ToolbarSpaceStyle`

I think they're only used in GObject properties. Would we get those if we generate them, or were those properties deprecated before 3.14? I'd add them for now if that causes new API to be generated.

src/auto/accel_map.rs
```diff
@@ -0,0 +1,166 @@
...
+    //}
+
+    pub fn lock_path(accel_path: &str) {
+        assert_initialized_main_thread!();
+        unsafe {
+            gtk_sys::gtk_accel_map_lock_path(accel_path.to_glib_none().0);
+        }
+    }
+
+    //pub fn lookup_entry(accel_path: &str, key: /*Ignored*/AccelKey) -> bool {
```
sdroege, line 102: Is this class useful without this function?

src/auto/enums.rs
```diff
@@ -1729,6 +1729,86 @@ impl SetValue for EventSequenceState {
...
+pub enum ExpanderStyle {
+    Collapsed,
+    SemiCollapsed,
+    SemiExpanded,
+    Expanded,
+    #[doc(hidden)]
+    __Unknown(i32),
+}
+
+impl fmt::Display for ExpanderStyle {
```
sdroege, line 1743: This does not seem to be used in any API we have, so just having the enum generated wouldn't provide anything useful.

src/auto/enums.rs
```diff
@@ -6512,6 +6592,84 @@ impl SetValue for StateType {
     }
 }
 
+#[derive(Debug, Eq, PartialEq, Ord, PartialOrd, Hash, Clone, Copy)]
+#[non_exhaustive]
+pub enum TextBufferTargetInfo {
```
sdroege, line 6597: This does not seem to be used in any API we have, so just having the enum generated wouldn't provide anything useful.

src/auto/flags.rs
```diff
@@ -111,6 +111,55 @@ impl SetValue for ApplicationInhibitFlags {
     }
 }
 
+bitflags! {
```
sdroege, line 114: Also this doesn't seem used anywhere

src/auto/flags.rs
```diff
@@ -215,6 +264,74 @@ impl SetValue for CellRendererState {
...
+    pub struct DebugFlag: u32 {
```
sdroege, line 268: And not this

src/auto/gradient.rs
```diff
@@ -0,0 +1,31 @@
...
+
+    match fn {
+        copy => |ptr| gobject_sys::g_boxed_copy(gtk_sys::gtk_gradient_get_type(), ptr as *mut _) as *mut gtk_sys::GtkGradient,
+        free => |ptr| gobject_sys::g_boxed_free(gtk_sys::gtk_gradient_get_type(), ptr as *mut _),
+        get_type => || gtk_sys::gtk_gradient_get_type(),
+    }
+}
+
+impl Gradient {
+    pub fn resolve_for_context<P: IsA<StyleContext>>(&self, context: &P) -> Option<cairo::Pattern> {
```
sdroege, line 23: The whole type is deprecated since 3.8, just in this function it was forgotten to specify the version

bilelmoussaoui:
I should have waited till morning before submitting this, removed the unused enums.

I have only kept Gtk.GestureStylus & Gtk.DebugFlag

bilelmoussaoui:
> > I was wondering about two enums which are deprecated in 3.20, should be generated?
> > ```
> > * `Gtk.ArrowPlacement`
> > 
> > * `Gtk.ToolbarSpaceStyle`
> > ```
> 
> I think they're only used in GObject properties. Would we get those if we generate them, or were those properties deprecated before 3.14? I'd add them for now if that causes new API to be generated.

they don't generate any new API

sdroege:
> I have only kept [...] Gtk.DebugFlag

That one also doesn't seem to be used anywhere.

src/auto/gesture_stylus.rs
```diff
@@ -0,0 +1,184 @@
...
+        skip_assert_initialized!();
+        unsafe {
+            Gesture::from_glib_full(gtk_sys::gtk_gesture_stylus_new(
+                widget.as_ref().to_glib_none().0,
+            ))
+            .unsafe_cast()
+        }
+    }
+
+    //#[cfg(any(feature = "v3_24", feature = "dox"))]
```
bilelmoussaoui, line 49: @sdroege any idea how can I generate this one? :)

sdroege: You have to do it manually, see the docs of this function in C: `axes: array of requested axes, terminated with GDK_AXIS_IGNORE`

bilelmoussaoui:
Updated with `get_axes` manual binding as well

bilelmoussaoui:
Somehow the CI failed to report it's state, but everything is green :)

PR #1035, opened 2020-07-13
Subclassing: Start implementing missing forall func for Gtk.Container

haecker-felix:
With this MR I try to implement the missing `forall` func in `ContainerImpl`. 

Disclaimer: I have no idea what I'm doing right now :). Never worked with gtk-rs binding stuff before.
To get this fully working, I need support from someone who knows how this stuff works in detail. 

Help is appreciated :D 

(Related issue: https://github.com/gtk-rs/gtk/issues/1036)

### DIFF
--- a/src/subclass/container.rs
+++ b/src/subclass/container.rs
@@ -1,6 +1,7 @@
 use gtk_sys;
 
 use glib::translate::*;
+use glib_sys::gboolean;
 
 use glib::subclass::prelude::*;
 
@@ -36,8 +37,13 @@ pub trait ContainerImpl: ContainerImplExt + WidgetImpl {
         self.parent_get_path_for_child(container, widget)
     }
 
-    fn forall<P: FnMut(&Widget)>(&self, container: &Container, callback: P) {
-        self.parent_forall(container, callback);
+    fn forall<P: FnMut(&Widget)>(
+        &self,
+        container: &Container,
+        include_internals: gboolean,
+        callback: P,
+    ) {
+        self.parent_forall(container, include_internals, callback);
     }
 }
 
@@ -48,7 +54,12 @@ pub trait ContainerImplExt {
     fn parent_set_focus_child(&self, container: &Container, widget: Option<&Widget>);
     fn parent_child_type(&self, container: &Container) -> glib::Type;
     fn parent_get_path_for_child(&self, container: &Container, widget: &Widget) -> WidgetPath;
-    fn parent_forall<P: FnMut(&Widget)>(&self, container: &Container, callback: P);
+    fn parent_forall<P: FnMut(&Widget)>(
+        &self,
+        container: &Container,
+        include_internals: gboolean,
+        callback: P,
+    );
 }
 
 impl<T: ContainerImpl> ContainerImplExt for T {
@@ -115,12 +126,21 @@ impl<T: ContainerImpl> ContainerImplExt for T {
         }
     }
 
-    fn parent_forall<P: FnMut(&Widget)>(&self, container: &Container, callback: P) {
+    fn parent_forall<P: FnMut(&Widget)>(
+        &self,
+        container: &Container,
+        include_internals: gboolean,
+        callback: P,
+    ) {
         unsafe {
             let data = self.get_type_data();
             let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
             if let Some(f) = (*parent_class).forall {
-                f(container.to_glib_none().0, false, callback.to_glib_none().0)
+                f(
+                    container.to_glib_none().0,
+                    include_internals,
+                    callback.to_glib_none().0,
+                )
             }
         }
     }
@@ -209,14 +229,14 @@ unsafe extern "C" fn container_get_path_for_child<T: ContainerImpl>(
 
 unsafe extern "C" fn container_forall<T: ObjectSubclass, P: FnMut(&Widget)>(
     ptr: *mut gtk_sys::GtkContainer,
-    callback: P
-)
-where
+    include_internals: gboolean,
+    callback: P,
+) where
     T: ContainerImpl,
 {
     let instance = &*(ptr as *mut T::Instance);
     let imp = instance.get_impl();
     let wrap: Borrowed<Container> = from_glib_borrow(ptr);
 
-    imp.forall(&wrap, callback)
+    imp.forall(&wrap, include_internals, callback)
 }
--- a/src/subclass/container.rs
+++ b/src/subclass/container.rs
@@ -1,8 +1,6 @@
 use gtk_sys;
 
 use glib::translate::*;
-use glib_sys::gboolean;
-use glib_sys::gpointer;
 
 use glib::subclass::prelude::*;
 
@@ -38,12 +36,7 @@ pub trait ContainerImpl: ContainerImplExt + WidgetImpl {
         self.parent_get_path_for_child(container, widget)
     }
 
-    fn forall<P: FnMut(&Widget)>(
-        &self,
-        container: &Container,
-        include_internals: gboolean,
-        callback: P,
-    ) {
+    fn forall(&self, container: &Container, include_internals: bool, callback: &Callback) {
         self.parent_forall(container, include_internals, callback);
     }
 }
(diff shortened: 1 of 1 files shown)


src/subclass/container.rs
```diff
@@ -109,6 +114,16 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
             from_glib_none(f(container.to_glib_none().0, widget.to_glib_none().0))
         }
     }
+
+    fn parent_forall<P: FnMut(&Widget)>(&self, container: &Container, callback: P) {
+        unsafe {
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
+            if let Some(f) = (*parent_class).forall {
+                f(container.to_glib_none().0, false, callback.to_glib_none().0)
```
haecker-felix, line 123: I'm currently struggling with this line. 

`pub forall: Option<unsafe extern "C" fn(*mut GtkContainer, gboolean, GtkCallback, gpointer)>,`

1) What does the gboolean here?!
2) How I'm supposed to handle the `gpointer`?

src/subclass/container.rs
```diff
@@ -35,6 +35,10 @@ pub trait ContainerImpl: ContainerImplExt + WidgetImpl + 'static {
     fn get_path_for_child(&self, container: &Container, widget: &Widget) -> WidgetPath {
         self.parent_get_path_for_child(container, widget)
     }
+
+    fn forall<P: FnMut(&Widget)>(&self, container: &Container, callback: P) {
+        self.parent_forall(container, callback);
```
sdroege, line 40: Should also implement `foreach` while we're at it. Exactly the same API

sdroege: Nevermind, both are handled with the same vfunc

src/subclass/container.rs
```diff
@@ -109,6 +114,16 @@ impl<T: ContainerImpl + ObjectImpl> ContainerImplExt for T {
...
+            let data = self.get_type_data();
+            let parent_class = data.as_ref().get_parent_class() as *mut gtk_sys::GtkContainerClass;
+            if let Some(f) = (*parent_class).forall {
+                f(container.to_glib_none().0, false, callback.to_glib_none().0)
+            }
+        }
+    }
 }
 
 unsafe impl<T: ObjectSubclass + ContainerImpl> IsSubclassable<T> for ContainerClass {
```
sdroege, line 129: You also need to add it here

src/subclass/container.rs
```diff
@@ -206,3 +221,17 @@ where
 
     imp.get_path_for_child(&wrap, &widget).to_glib_none().0
 }
+
+unsafe extern "C" fn container_forall<T: ObjectSubclass, P: FnMut(&Widget)>(
+    ptr: *mut gtk_sys::GtkContainer,
+    callback: P
```
sdroege, line 227: ... because then you would see that the signature here is wrong :) It should be `fn(*mut GtkContainer, gboolean, GtkCallback, gpointer)`

src/subclass/container.rs
sdroege: > 1. What does the gboolean here?!

It switches between the `foreach` and `forall` behaviour. It's called `include_internals` and you would simply add that to all your functions here and pass it through. It's something the implementer has to take care of.

> 2. How I'm supposed to handle the `gpointer`?

That's the actually tricky part here. You get the `gpointer` from C (in `container_forall()` below). You would have to wrap that together with the `GtkCallback` in some new type:

```rust
pub struct Callback {
    cb: gtk_sys::GtkCallback,
    user_data: gpointer,
}

impl Callback {
    pub fn call(&self, widget: &Widget) {
        unsafe {
            self.cb(widget.to_glib_none().0, self.user_data);
        }
    }
}
```

And then pass such a `&Callback` through the Rust functions.

sdroege:
Any updates here?

haecker-felix:
> Any updates here?

I was on vacation for 2 weeks, I'll try to finish this next weekend.

src/subclass/container.rs
haecker-felix: Hm - can you clarify this a bit more? I really struggle to understand how this works. I had the hope it's a bit easier to implement :/

```
unsafe extern "C" fn container_forall<T: ObjectSubclass, P: FnMut(&Widget)>(
    ptr: *mut gtk_sys::GtkContainer,
    include_internals: gboolean,
    callback: P,
    user_data: gpointer,
)
```

- What should be the last `gpointer`? `user_data`?
- For `callback` I have currently `P: FnMut(&Widget)`. That should be get replaced with `GtkCallback`?

src/subclass/container.rs
haecker-felix: Done - The `include_internals` param should be added everywhere now.

src/subclass/container.rs
haecker-felix: It complains about `self.cb(...)`

```
error[E0599]: no method named `cb` found for reference `&subclass::container::Callback` in the current scope
   --> src/subclass/container.rs:269:18
    |
269 |             self.cb(widget.to_glib_none().0, self.user_data);
    |                  ^^----------------------------------------- help: remove the arguments
    |                  |
    |                  field, not a method
```

haecker-felix:
I have tried to add some missing pieces to the puzzle - this is were I'm currently:

```
error[E0061]: this function takes 4 arguments but 3 arguments were supplied
   --> src/subclass/container.rs:140:17
    |
140 |                 f(
    |                 ^ expected 4 arguments
141 |                     container.to_glib_none().0,
    |                     --------------------------
142 |                     include_internals,
    |                     -----------------
143 |                     callback.to_glib_none().0,
    |                     ------------------------- supplied 3 arguments

error[E0599]: no method named `to_glib_none` found for type parameter `P` in the current scope
   --> src/subclass/container.rs:143:30
    |
143 |                     callback.to_glib_none().0,
    |                              ^^^^^^^^^^^^ method not found in `P`
    |
    = note: `callback` is a function, perhaps you wish to call it
    = help: items from traits can only be used if the type parameter is bounded by the trait
help: the following trait defines an item `to_glib_none`, perhaps you need to restrict type parameter `P` with it:
    |
130 |     fn parent_forall<P: glib::translate::ToGlibPtr + FnMut(&Widget)>(
    |                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0599]: no method named `cb` found for reference `&subclass::container::Callback` in the current scope
   --> src/subclass/container.rs:268:18
    |
268 |             self.cb(widget.to_glib_none().0, self.user_data);
    |                  ^^----------------------------------------- help: remove the arguments
    |                  |
    |                  field, not a method

error: aborting due to 3 previous errors
```

If anyone wants to finish this for me, please feel free to do so. 
I have the feeling I will have to struggle with this PR for a long time, since this is kinda black magic for me. 🪄

src/subclass/container.rs
sdroege: I also have a hard time remembering how this worked after that long time, will have to read it up again :) It's not a simple case unfortunately.

src/subclass/container.rs
sdroege: `(self.cb)(...)` would be the solution here

sdroege:
> If anyone wants to finish this for me, please feel free to do so.

I'll take a look at finishing this some time later today. But you'll have to tell me if it actually works as expected :)

sdroege:
This should work fine now. @haecker-felix can you confirm?

@GuillaumeGomez @EPashkin can you review?

GuillaumeGomez:
The code looks good to me. However: my knowledge of subclassing is close to 0 so my opinion isn't worth much here. :)

sdroege:
Ok, then let's get it in and have @haecker-felix confirm that it solves whatever he's trying to do later :) Thanks!

haecker-felix:
My use case doesn't exist anymore, I implemented my widget directly in C / libhandy since it wasn't possible in Rust earlier. 

https://gitlab.gnome.org/GNOME/libhandy/-/merge_requests/562

I just glued a small test together, and it seems that it works. Thanks a lot @sdroege for finishing this. I'll try to make a simple Gtk.Container subclassing example later.

PR #1048, opened 2020-08-24
Remove unneeded wrappers

GuillaumeGomez:
cc @sdroege @EPashkin

### DIFF
--- a/src/signal.rs
+++ b/src/signal.rs
@@ -3,62 +3,10 @@
 // Licensed under the MIT license, see the LICENSE file or <http://opensource.org/licenses/MIT>
 
 use gdk::Rectangle;
-use glib;
 pub use glib::signal::Inhibit;
 use glib::signal::SignalHandlerId;
-use glib::SourceId;
 
-use glib::Continue;
 use {ScrollType, Widget};
-
-/// Adds a closure to be called by the default main loop when it's idle.
-///
-/// `func` will be called repeatedly until it returns `Continue(false)`.
-///
-/// Similar to `glib::idle_add` but only callable from the main thread and
-/// doesn't require `Send`. It is the same as `glib::idle_add_local`.
-pub fn idle_add<F>(func: F) -> SourceId
-where
-    F: FnMut() -> Continue + 'static,
-{
-    assert_initialized_main_thread!();
-    glib::idle_add_local(func)
-}
-
-/// Adds a closure to be called by the default main loop at regular intervals
-/// with millisecond granularity.
-///
-/// `func` will be called repeatedly every `interval` milliseconds until it
-/// returns `Continue(false)`. Precise timing is not guaranteed, the timeout may
-/// be delayed by other events. Prefer `timeout_add_seconds` when millisecond
-/// precision is not necessary.
-///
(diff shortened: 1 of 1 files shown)


sdroege:
:+1:

src/signal.rs
```diff
@@ -11,55 +11,6 @@ use glib::SourceId;
 use glib::Continue;
```
sdroege, line 11: This is not used anymore in this file so probably causes a compiler warning

GuillaumeGomez:
Updated!

fengalin:
Shouldn't we deprecate them with a message hinting to use the `glib` variants?

GuillaumeGomez:
And it'll be mentionned in the release notes because we're nice people. :)

fengalin:
I don't mind, but my point was to use a language feature that makes it automatic to point user in the right direction when they upgrade their dependencies, without the need to look in any doc.

We can still do it C-old-school though ;)

GuillaumeGomez:
We can do that but it means that'll have to track them...

sdroege:
> We can do that but it means that'll have to track them...

And we did that in gstreamer-rs once and only today I noticed that the deprecated functions were not removed yet, more than a month after release. It's additional maintenance burden.

> We can still do it C-old-school style though ;)

You can also do that in C. Maybe we should do a bugfix release that deprecates the functions now on crates.io already?

fengalin:
That or just launch a [rip]grep every now and then to remove previously deprecated items.

I find it very nice to get a warning for such a thing when I upgrade my dependencies in mass. I don't necessarily browse all the release notes for all crates.

fengalin:
> And we did that in gstreamer-rs once and only today I noticed that the deprecated functions were not removed yet, more than a month after release. It's additional maintenance burden.

Sorry, that was my idea. I should have checked with you how long we would keep them deprecated before removing them and created an issue to track their removal.

I still believe this is since to users though.

GuillaumeGomez:
How do you differentiate between the not-yet deprecated items (meaning, not released as such yet) and the others?

fengalin:
I agree there's some tracking involved at one point. You can indicate the release from which deprecation occurs. I think deprecated items can stay for a couple of releases. It might help people still get the hint should they skip one or two releases. Then we can follow a principle such as: after n releases after deprecation, the item is removed, which tracking is obvious thank to the depreciation release attribute.

sdroege:
While it seems like the right thing to do and is possible in this case, it's much harder or even impossible to add deprecations in other cases as the old API can't really be kept.

GuillaumeGomez:
Two things: first I opened a PR to [examples](https://github.com/gtk-rs/examples/pull/320) to fix the CI issue. Second, I propose to make a small release with the deprecation notice (which will be in a commit only in the corresponding branch) so that everyone is satisfied: no need for us to keep track of anything and users who are up-to-date will have the deprecation notice.

GuillaumeGomez:
Reminder: still needs to add the deprecation marker on the current release and make a small release with it.

GuillaumeGomez:
Usual mac failures so let's finally move on.

PR #1057, opened 2020-09-13
Readme: Fix doc link

Hofer-Julian:


GuillaumeGomez:
The link is not invalid here. But I wonder if we should point directly to the doc pages. Not sure...

cc @sdroege @EPashkin

Hofer-Julian:
> The link is not invalid here. But I wonder if we should point directly to the doc pages. Not sure...
> 
> cc @sdroege @EPashkin

Really? I get this when I click on it
grafik

GuillaumeGomez:
Oh damn, you're right. The correct URL is: https://gtk-rs.org/docs-src/ Can you use this one instead please?

Hofer-Julian:
Done :)

GuillaumeGomez:
Thanks!

PR #1059, opened 2020-09-21
Subclassing: ListBox and ListBoxRow

kavanmevada:


src/subclass/list_box_row.rs
```diff
@@ -0,0 +1,14 @@
...
+use super::bin::BinImpl;
+
+use BinClass;
+use ListBoxRowClass;
+
+pub trait ListBoxRowImpl: BinImpl {}
+
+unsafe impl<T: ListBoxRowImpl> IsSubclassable<T> for ListBoxRowClass {
+    fn override_vfuncs(&mut self) {
+        <BinClass as IsSubclassable<T>>::override_vfuncs(self);
```
sdroege, line 12: Might make sense to add the `activate` vfunc here. Easy enough and seems useful

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
...
+}
+
+unsafe extern "C" fn list_box_unselect_all<T: ListBoxImpl>(
+    ptr: *mut gtk_sys::GtkListBox) {
+    let instance = &*(ptr as *mut T::Instance);
+    let imp = instance.get_impl();
+    let wrap: Borrowed<ListBox> = from_glib_borrow(ptr);
+
+    imp.unselect_all(&wrap)
+}
```
sdroege, line 208: Missing newline at the end

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
...
+        self.list_box_selected_rows_changed(list_box)
+    }
+
+    fn toggle_cursor_row(&self, list_box: &ListBox) {
+        self.list_box_toggle_cursor_row(list_box)
+    }
+
+    fn unselect_all(&self, list_box: &ListBox) {
+        self.list_box_unselect_all(list_box)
+    }
```
sdroege, line 40: `move_cursor()` is missing here and also seems easy to add, then we have this one complete at least :)

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
...
+use WidgetClass;
+use ListBox;
+use ListBoxRow;
+
+pub trait ListBoxImpl: ListBoxImplExt + ContainerImpl + WidgetImpl {
+    fn row_activated(&self, list_box: &ListBox, row: &ListBoxRow) {
+        self.list_box_row_activated(list_box, row)
+    }
+
+    fn row_selected(&self, list_box: &ListBox, row: &ListBoxRow) {
```
sdroege, line 18:
```suggestion
    fn row_selected(&self, list_box: &ListBox, row: Option<&ListBoxRow>) {
```

It can be `None` according to the docs

sdroege (commented):
Looks generally good to me, thanks! :)

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
+
```
GuillaumeGomez, line 1: Missing header (take a look at the other files and copy/paste it).

src/subclass/list_box_row.rs
```diff
@@ -0,0 +1,14 @@
+use glib::subclass::prelude::*;
```
GuillaumeGomez, line 1: Missing header.

GuillaumeGomez, line 1: Also, please don't use wildcard import in the crate directly.

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
...
+use glib::translate::*;
```
GuillaumeGomez, line 2: No wildcard card import please.

src/subclass/list_box.rs
```diff
@@ -0,0 +1,208 @@
...
+
+use glib::subclass::prelude::*;
```
GuillaumeGomez, line 4: No wildcard card import please.

src/subclass/list_box_row.rs
sdroege: Why not? We do that elsewhere too :)

GuillaumeGomez: We shouldn't. My thinking here is that it makes the code harder (and eventually slower to compile) in the location where it shouldn't. The crate provides the API, so we should make clear what it uses. I can update the other places if needed.

sdroege: We're not in `glib` here but in `gtk`, so it comes from another crate :) And it doesn't change anything with regards to compile times: the compiler knows all the traits anyway, or how would it otherwise tell you "maybe you want to put this trait into scope for being able to call this function"?

By not including the prelude from GLib here we will need more inter-crate changes whenever some trait inside glib changes its name or functions get moved between traits or traits get split or ...

sdroege: So I completely disagree here, changing it to explicit imports only causes more inter-crate dependencies and we really don't want that. Too bad @kavanmevada already changed it now...

src/subclass/list_box_row.rs
```diff
@@ -0,0 +1,42 @@
...
+use BinClass;
+use ListBoxRow;
+use ListBoxRowClass;
+
+use glib::translate::{Borrowed, from_glib_borrow};
+use glib::subclass::types::{IsSubclassable, InstanceStruct};
+
+use super::bin::BinImpl;
+
+
```
sdroege, line 14: Two empty newlines, one is enough :)

sdroege:
Looks good to me otherwise, thanks!

sdroege:
You want to run `cargo fmt` once :)

kavanmevada:
`cargo fmt` also done!, thanks.

sdroege (approved):
CI seems fine apart from travis falling apart generally. @GuillaumeGomez you can merge :)

PR #1061, opened 2020-09-26
Add gtk::Clipboard::request_uris()

YaLTeR:
Did I do this right?

### DIFF
--- a/src/clipboard.rs
+++ b/src/clipboard.rs
@@ -5,7 +5,7 @@
 use glib::translate::*;
 use glib_sys::gpointer;
 use gtk_sys;
-use libc::c_uint;
+use libc::{c_char, c_uint};
 use std::boxed::Box as Box_;
 use Clipboard;
 use SelectionData;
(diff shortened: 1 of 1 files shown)


src/clipboard.rs
```diff
@@ -68,4 +68,27 @@ impl Clipboard {
...
+
+    pub fn request_uris<P: FnOnce(&Clipboard, &[glib::GString]) + 'static>(&self, callback: P) {
+        let callback_data: Box_<P> = Box_::new(callback);
+        unsafe extern "C" fn callback_func<P: FnOnce(&Clipboard, &[glib::GString]) + 'static>(
+            clipboard: *mut gtk_sys::GtkClipboard,
+            uris: *mut *mut c_char,
+            data: glib_sys::gpointer,
+        ) {
+            let clipboard = from_glib_borrow(clipboard);
+            let uris = from_glib_borrow(uris);
```
GuillaumeGomez, line 80: Just want to be sure here: does `glib_from_borrow` handle correctly pointers of pointers and stop at the first `0`/`NULL`?

Maybe you know @sdroege ?

GuillaumeGomez:
Just something I'm wondering about but otherwise, it looks good to me, thanks!

src/clipboard.rs
sdroege: I don't think this compiles

sdroege: See this for an example

```rust
    pub fn get_uris(&self) -> Vec<GString> {
        unsafe {
            FromGlibPtrContainer::from_glib_full(gtk_sys::gtk_selection_data_get_uris(
                self.to_glib_none().0,
            ))
        }
    }
```

YaLTeR: Should this be `from_glib_full`, `container` or `none`? The docs have no transfer annotation.

sdroege: `none`

YaLTeR: Okay, take two.

sdroege (approved):

sdroege:
@GuillaumeGomez Good to be merged apart from travis timing out

GuillaumeGomez (approved):

GuillaumeGomez:
Thanks!

PR #1067, opened 2020-10-09
groovy update: regenerate

bilelmoussaoui:


src/auto/settings.rs
```diff
@@ -77,6 +77,10 @@ pub trait SettingsExt: 'static {
         gtk_application_prefer_dark_theme: bool,
     );
 
+    fn get_property_gtk_cursor_aspect_ratio(&self) -> f32;
```
sdroege, line 80: No `feature = "v3_24_XX"`?

sdroege (commented):
Looks good to me otherwise

src/auto/settings.rs
sdroege: This a) needs annotations in the C code, and b) configured in the `Gir.toml`

bilelmoussaoui: It was added in Gtk 3.24.11 https://gitlab.gnome.org/GNOME/gtk/-/blob/gtk-3-24/NEWS#L90

sdroege:
@bilelmoussaoui Please regenerate with the merged version of `gir-files` and then let's get it in :)

src/auto/settings.rs
sdroege: Can you (generally) also link to the upstream MR (or issues) you created for these things?

sdroege (approved):

GuillaumeGomez:
Thanks!

